PR #25, opened 2015-01-13
Add per exe prefix, useful for gdb, gprof, valgrind, etc.

nicholash:


marshallward:
I'm afraid that if I merge this then I'll forget how to implement gprof and massif!

I think it'd be good to create the "profiler" drivers which set (or append) exec_prefix.

But let's just merge this for now and then add those later?

nicholash:
Haha, :), I know what you mean, but you can still keep these commands in
your config.yaml.

OK, if you're happy with it for now I'll merge.

marshallward:
yeah please do
The profiler code is already getting too sloppy
This will at least prevent me from putting more in

PR #34, opened 2015-03-11
Add a simple MOM6 model based on GOLD

angus-g:
The input scheme for MOM6 is pretty much exactly the same as GOLD, however the `MOM_input` file format is very slightly different to `GOLD_input` (this doesn't matter at the moment, as there's no dynamic configuration done by payu here).


marshallward:
Thanks! I'll roll it into a 0.5 release (once I fix the mppnccombine bug that Aidan found)

PR #38, opened 2015-06-23
Added option to directly specify collate directory

aidanheerdegen:
Hey Marshall,

I had a crack at implementing the option to directly specify a collate directory. 

I did a few quick tests, and it seems to work from payu collate and payu-collate. 

I'm not sure if this is the way you would have done it, or if I might have broken something else. In any case, have a look, and tell me what you think.

It has another use I hadn't thought about: collating restarts for changing processor count/layout.

Cheers

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -101,6 +101,9 @@ all collation jobs are single CPU jobs and are not parallelised.
    Ignore these files during collation. This can either be a single filename or
    a list of filenames. Only FMS models (MOM, GOLD) support this setting.
 
+``collate_flags``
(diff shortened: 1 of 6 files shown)


marshallward:
This is in addition to the "payu collate -i ##" selection right?
Yeah, this is a good idea.
Let me give it a try later today and I'll merge it.

aidanheerdegen:
Yeah. Either/or. I mentioned the idea a while back and thought I'd just get off my keister and do it. I came across the need when I had a directory that was no longer in the payu directory structure and was annoyed I had do the collation "by hand".

PR #39, opened 2015-06-26
Removed dependency on git -C

aidanheerdegen:
The git dependency thing was tripping me up a lot and I saw an easy way around it.

If the cwd flag for subprocess isn't desirable then don't merge.

I also changed the import statement for subprocess to match (most of) the rest of the code

### DIFF
--- a/payu/runlog.py
+++ b/payu/runlog.py
@@ -11,7 +11,7 @@
 import datetime
 import os
 import shlex
-import subprocess
+import subprocess as sp
 
 # Local
 from payu import envmod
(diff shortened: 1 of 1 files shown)


marshallward:
Can you revert the "subprocess as sp" change? I'm trying to phase out the module renames.

marshallward:
Otherwise yea, using `cwd` is a great idea. (If this ever needs to be called in an external `control_path` then we can worry about that issue, or just insert a `chdir`).

aidanheerdegen:
Sorry, I just did what seemed to be most common. 

As it turns out, I am trying to figure out the best way to pass a single flag and get payu to spit commands to stdout rather than execute them. I thought this might constitute a reasonable first step at implementing a testing framework, just compare the list of commands to the "truth".

So I thought it might be better to keep sp, as this could end up being a wrapped module?

marshallward:
so `sp` would be set to either `subprocess` or an inert echo? I think that's a good idea. But I'd prefer it have a new name like `handler` or something.

I just don't like renaming things unless there's a need (e.g. namespace conflict).

Anyway... it is not that important. I will merge it.

aidanheerdegen:
I wasn't too sure the best way to do it TBH, so if you've got an idea how to implement it that'd be great.

PR #40, opened 2015-07-08
Added parallel collation for fms and fixed a small bug in '-d' implementation

aidanheerdegen:
Paul's collation jobs were taking a long time, he was saving at high temporal resolution and doing ensembles, and my high temporal resolution 0.1 deg sims were taking 12+ hours to collate, so this is a fix of sorts.

I think it would be better to submit multiple collate jobs, but that would have been a bigger change, so I settled for using multiprocessing to submit multiple jobs within one collate job.

This adds config option collate_ncpus, to specify the number of cpus to be requested for the collate PBS submission.

The drawback is that multiprocessor jobs can't be submitted to the copy queue on raijin, so normal or express must be specified. Care must be taken to also request enough memory for multiple collate jobs to run simultaneously.

There was a bug in the '-d' directory path implementation, that it didn't work at all when invoked in a directory that wasn't the laboratory, which sort of defeats the purpose. In any case a minimal config.yaml must be supplied so that payu knows the model type. If the job name can't be determined payu-collate falls back to the name the PBS job after the directory path being collated. I had an issue with a trailing slash on the PBS job name interacting poorly with PBS. I sprinkled a few os.path.normpath calls around, but in the end the only one that matters is in collate_cmd.py. I left the others in as I didn't think they could do any harm, and might be beneficial.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -159,10 +159,10 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None):
         payu_env_vars['PAYU_N_RUNS'] = n_runs
 
     if lab_path:
-        payu_env_vars['PAYU_LAB_PATH'] = lab_path
+        payu_env_vars['PAYU_LAB_PATH'] = os.path.normpath(lab_path)
 
     if dir_path:
-        payu_env_vars['PAYU_DIR_PATH'] = dir_path
+        payu_env_vars['PAYU_DIR_PATH'] = os.path.normpath(dir_path)
 
     return payu_env_vars
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -251,7 +251,7 @@ class Experiment(object):
         # Check to see if we've provided a hard coded path -- valid for collate
         dir_path = os.environ.get('PAYU_DIR_PATH')
         if dir_path is not None:
-            self.output_path = dir_path
+            self.output_path = os.path.normpath(dir_path)
         else:
             output_dir = 'output{:03}'.format(self.counter)
             self.output_path = os.path.join(self.archive_path, output_dir)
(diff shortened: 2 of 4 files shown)


marshallward:
Thanks, this looks interesting. Do we need to do a new release ASAP?

aidanheerdegen:
I've installed this as payu/dev version under /projects/v45 so Paul can access it. I'm not sure there is anyone else who needs any of these features just yet. It might be a good idea to test it for a while first in that case.

marshallward:
ok.
I am doing multidim support for f90nml so I'd rather get that working
before an update.

aidanheerdegen:
Yup. Might sneak a few more features in too ... :)

aidanheerdegen:
I stuffed up and introduced a bug, so I've fixed that and added a couple of other things, but I realised that by default, as it is currently written, when running payu-collate it will run in parallel and use as many CPUs as are available.

This could be considered a feature or a bug, depending on who pays your salary. 

Should I change this behaviour?

marshallward:
"as many CPUs as are available" as in as many requested for the PBS job? As many on the node? Or raijin?

I don't see the first as a problem. If you request 16 then you should use 16. The other cases could be a problem though.

aidanheerdegen:
Yeah, I realised that payu-run is still faithful to the ncpus setting in config.yaml, so payu-collate should be too. I'll change that before I commit & pull.

marshallward:
"ncpus" might be the problem.. maybe a collate_ncpus...
(or maybe a new collate subcategory)

aidanheerdegen:
There is already a collate_ncpus config option, I just wasn't doing anything with it when payu-collate was called. I will honour that setting. As I said, makes it consistent, which is a virtue.

aidanheerdegen:
I don't think we can have a new collate subcategory without break previous scripts can we?

marshallward:
ah.. well you just check for both for backwards compatibility.. but its not
really necessary. Dont worry about it

PR #41, opened 2015-07-09
Made payu-collate respect collate_ncpus from config. Fixed bug introduced in last commit: referencing unknown variable

aidanheerdegen:
Previous version didn't work when submitted as a batch job because of a dumb error I introduced. Should have checked that.

The only substantial revision is to make fms.collate respect the collate_ncpus setting from config. This wasn't the case in the previous version. That config was only used in the batch version, which sets up the number of CPUs for the job and fms.collate just grabbed the number of CPUs from the environment.

So now payu-collate does (collate_cmd.py) does a (slightly bodgy) test to see if we're in a PBS environment (not strictly necessary I guess ... ) and then sets an environment variable to pass this information on to fms.collate.

Marshall, maybe you'd prefer not to do that bodgy test and just set the environment variable? I tried looking in the run command to find an analogous example, but couldn't see anything. If there is a standard/usual/better way of doing this I'm happy to change it.

### DIFF
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -15,14 +15,23 @@ import os
 import resource as res
 import shlex
 import subprocess as sp
-import multiprocessing
-from multiprocessing.dummy import Pool
-
-def runcmd(cmd, cwd):
-    return sp.call(shlex.split(cmd), cwd=cwd)
+# Use multiprocessing dummy (threads) as collate jobs run in own process
+import multiprocessing.dummy as multiprocessing
 
 # Local
 from payu.models.model import Model
+
+def runcmd(cmd, cwd):
+    # This is run in a thread, so the GIL of python makes it sensible to
+    # capture the output from each process and print it out at the end so
+    # it doesn't get scrambled when collates are run in parallel
+    result = True
+    try:
+        output = sp.check_output(shlex.split(cmd), cwd=cwd, stderr=sp.STDOUT)
+    except:
+        result = False
+    print output
+    return result
 
 class Fms(Model):
 
@@ -93,9 +102,11 @@ class Fms(Model):
 
             mnc_tiles[t_base].append(t_fname)
 
-        # We use a multiprocessing dummy pool (so just threads) as the collate
-        # jobs are run in their own process
-        pool = Pool(processes=multiprocessing.cpu_count())
+
+        # If this is run interactively NCPUS is set in collate_cmd, otherwise
+        # the cpu_count will return the number of CPUs assigned to the PBS job
+        count = int(os.environ.get('NCPUS',multiprocessing.cpu_count()))
+        pool = multiprocessing.Pool(processes=count)
 
         # Collate each tileset into a single file
         for nc_fname in mnc_tiles:
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -21,7 +21,7 @@ import multiprocessing.dummy as multiprocessing
 # Local
 from payu.models.model import Model
 
-def runcmd(cmd, cwd):
+def cmdthread(cmd, cwd):
     # This is run in a thread, so the GIL of python makes it sensible to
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
Should I have just ncpus in the model config instead of the environ?

marshallward:
Quick comments

- `print output` needs to be `print(output)`, otherwise it will fail in python 3. 
   We do currently force the user onto python 2.7, but I hope that will change someday.
   (Or is this a leftover diagnostic?)

- Not sure why this:

        pbs_config['jobname'] = pbs_config.get('jobname', default_job_name)[:13] + '_c'

  was changed to this:

       pbs_config['jobname'] = pbs_config['jobname'][:13] + '_c'

  I don't think `pbs_jobname` is always defined, is it?

- Maybe just assumbe that PBS is available, and read `PBS_NCPUS`, if it exists.
  
  I don't know why, but using `PAYU_CURRENT_RUN` to distinguish between a shell session or a payu run seems like a bad idea.

I can come by tomorrow and talk about this. So far I haven't even tested it yet!

aidanheerdegen:
Thanks for the heads up on the print.

I print the output to be consistent with how it was done before, where the stdout and stderr just get spewed out when it was check_call. I capture it so that when it is spat out it is the output from one job at a time. Now there is the possibility of multiple collate jobs running the output is a ridiculous mess otherwise.

I changed the jobname bit because that was the bug. I was trying to use default_job_name which sometimes didn't exist in non-PBS environment. I was going to do it one way, decided on another and then forgot to take that out. The new line is in a conditional that checks for the existence of jobname, so it should always be defined. I think this was just catching the case when I was running payu from a directory with a minimal config.yaml, which didn't have a jobname defined. I was (am?) trying to get to a point where some of the payu functionality can be divorced from a particular location, but this might be a stupid idea. It was motivated by wanting to run payu-collate on a directory that was no longer in a payu lab directory hierarchy. I now realise it needs (at the very least) to know what sort of model it is trying to collate, so a minimal config.yaml is required.

Still I'm happy with the functionality of the -d option, as I can easily collate restarts as well, which is fantastic for easy changing between layouts etc.

I do wonder if -d was the best letter, as this is usually for debug, which might be required down the track? I could change it to -p (--path) now before anyone starts to use it. Whaddya reckon?

The PBS check is to see if PBS *isn't* running, to set an NCPUS variable. Don't need to do it for the PBS case, as it just defaults to using all the CPUs.

Could just check for PBS_NCPUS and set it if it hasn't been already? But this is a bit odd, as it is being set to be used specifically because it isn't in a PBS call.

If you want to brave the rain you're more than welcome to come over for a chinwag.

marshallward:
I was actually thinking of chaning `default_jobname` to the directory name.

Would that change things for you?

aidanheerdegen:
Shouldn't be an issue.

marshallward:
hm... `default_job_name` was not even defined in that file...

aidanheerdegen:
Yeah, hence the bug. Just testing a revision, I'll push and then you can make whatever changes you want.

marshallward:
ok sounds good, thanks for the patch

aidanheerdegen:
So that update fixes the print command (is return without parentheses ok?) and changes the test for the PBS status. I didn't change the enviro variable to PBS_NCPUS as I thought this would be confusing. It is specifically for when it isn't running under PBS.

I have tested payu-collate and payu collate and both seem fine.

marshallward:
I just noticed that NCPUS appears to already be set, and seems to be an alias for PBS_NCPUS.

But I'm happy to not get involved in this, and let you make the call.

aidanheerdegen:
Really?! This is just for PBS jobs? 

Sorry. I thought I'd checked through the enviro vars.

I don't think it matters then. The only problem I has was setting PBS_NCPUS for a job that was running interactively.

Do you agree?

Issue #59, opened 2016-08-19
ACCESS and OASIS restarts copied incorrectly if symlinks


aidanheerdegen:
When the output directory is a symlink copying the ice and oasis restart files resulted in broken links (back to work directory).

marshallward:
This has been merged, but leaving this issue open since there may still be a way to do this with symlinks.  (More discussion in #60)

aidanheerdegen:
No update traffic, so will close.

PR #60, opened 2016-08-19
 Issue 59: Changed the shutil.move() to shutil.copy2(), which will follow symlinks.

aidanheerdegen:
Fix for Issue #59

### DIFF
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -149,10 +149,10 @@ class Access(Model):
         for model in self.expt.models:
             if model.model_type == 'cice':
 
-                # Move supplemental restart files to RESTART path
+                # Copy supplemental restart files to RESTART path
                 for f_name in model.access_restarts:
                     f_src = os.path.join(model.work_path, f_name)
                     f_dst = os.path.join(model.restart_path, f_name)
 
                     if os.path.exists(f_src):
-                        shutil.move(f_src, f_dst)
+                        shutil.copy2(f_src, f_dst)
--- a/payu/models/oasis.py
+++ b/payu/models/oasis.py
@@ -118,7 +118,7 @@ class Oasis(Model):
             f_dst = os.path.join(self.restart_path, f)
 
             if os.path.exists(f_src):
-                shutil.move(f_src, f_dst)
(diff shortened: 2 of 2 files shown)


marshallward:
Does this delete the original?  (And does it matter?)

aidanheerdegen:
Uh, nope. I just assumed, which was lazy, that it would get deleted when the work directory was blown away.

marshallward:
If I recall, some work directories get clobbered, while others get turned into the new archive.  I think it's model dependent.

Might be good to delete the original, just in case.

I am reading about `shutil.copy2` and it seems to be more for metadata preservation than following symlinks.  `shutil.copy` also seems to support symlink following.

I'm guessing this works, but maybe there are some Python version differences here?  I ask because I am thinking of rewriting everything to work under Python 2.6 with Python 3.x compatibility, which would eliminate the need for all the `reversion.repython` code.

marshallward:
Actually the symlink path may be the metadata that gets fixed in this case...?

aidanheerdegen:
I chose copy2 to be as close as possible to the move command. copy would be fine as well. I'll addd code to delete the originals and push the changes.

marshallward:
Did you want to actually copy the symlinked contents?  That seems to be what `shutil.copy` is doing here.

Traditionally I've deleted any symlinks within the work directories, with the assumption that the data must already be archived elsewhere.

(BTW this bug is coming back to me now... we decided to let someone else fix it someday... I guess that someone is you!)

aidanheerdegen:
Sorry, just noticed this was still open. Is it still relevant?

marshallward:
The issue was that you were copying rather than linking, is that right?

I don't use payu for access-cm, do you want me to check this out and test it?

aidanheerdegen:
It was fixing this issue https://github.com/marshallward/payu/issues/59

marshallward:
Right, you fixed it by explicitly copying, and (if I am remembering) I was asking if there was a way to do that without copying.

Let me try this out and see if I can clarify further.

aidanheerdegen:
Looks like I never pushed the copy/delete changes I said I would do. Bad me.

PR #61, opened 2017-01-31
Collatefix

aidanheerdegen:
This includes those shutil.copy2 changes. Not sure if you wanted those or not.

### DIFF
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -149,10 +149,10 @@ class Access(Model):
         for model in self.expt.models:
             if model.model_type == 'cice':
 
-                # Move supplemental restart files to RESTART path
+                # Copy supplemental restart files to RESTART path
                 for f_name in model.access_restarts:
                     f_src = os.path.join(model.work_path, f_name)
                     f_dst = os.path.join(model.restart_path, f_name)
 
                     if os.path.exists(f_src):
-                        shutil.move(f_src, f_dst)
+                        shutil.copy2(f_src, f_dst)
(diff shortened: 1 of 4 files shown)


aidanheerdegen:
Sorry, also meant to say I added in a collate_exe option, which I've been meaning to have for a while. The collate_* config namespace is a bit of a dog's breakfast. I did think when I started adding stuff to it about breaking it up into
```yaml
collate:
      exe:
      ncpus:
      flags:
```
but it would have broken the existing boolean config, or was there a clever way to do it?

marshallward:
All good, thanks Aidan.

I will create an issue about the potential storage issues with copying the coupling files.  Copying is better than preserving dead symlinks.

marshallward:
I would prefer a more tree-link configuration file like your example, but happy to do that in the future.  (Unless you want to do it now?)

aidanheerdegen:
But how could we preserve backwards compatibility with the

```yaml
collate: true
```
Current usage?

marshallward:
yeah, that would have to be changed, which is fine with me.

A hacky solution is possible where collate is either a boolean (True/False) or a dict, which is implicitly true and has more info.

marshallward:
But I am happy to defer these sort of changes.

aidanheerdegen:
It will introduce a lot of cruft, where all the collate_* options have to be either booleans or members of a dict. Ugh. I should have done it properly the first time. Sorry.

marshallward:
I don't mind cleaning it up, happy to add another ugly `collate_*` setting for now.  (Unless you want to do it?)  Should I merge this in for now?

aidanheerdegen:
Sure. If you do want to clean it up, and have the time, it is probably best to do so before collate_exe becomes widely used. One less crufty option to support.

At some stage in the future those configs can be deprecated I suppose.

marshallward:
ok, i will do it before a version update.

PR #67, opened 2017-04-11
qgcm support

aidanheerdegen:
Nasty hack to get it to work, need to explicitly specify the number of CPUs to be the same as the number of threads, and turn off binding
```yams
ncpureq: 4
mpthreads: 4

mpirun:
   - --bind-to none
```


marshallward:
Thanks Aidan, we'll get it cleaned up at some point in the future.

PR #74, opened 2017-06-22
Added explicit exception catching to cmdthread in fms.py

aidanheerdegen:
Addressing https://github.com/marshallward/payu/issues/73

I'm not entirely sure why I let this just go through, I guess we're sanguine about a collation failing, and will want to still collate other files if possible. It is up to the user to check and maybe alter resources for the collation step.


marshallward:
Thanks Aidan

PR #88, opened 2017-11-09
87 no prior output for restart

nicjhan:


marshallward:
I think I modified something here, but now it won't let me merge...?

marshallward:
Nevermind, working now!  As for release...

PR #89, opened 2017-11-14
Add option for multiple runs per PBS submit (runspersub)

aidanheerdegen:
Added a config option `nrunspersub` to run multiple model runs for each PBS submission which addresses https://github.com/marshallward/payu/issues/84

Implemented this as a parameter in `config.yaml` as it affects the choice of PBS parameters, i.e. `walltime`.

This option respects the existing `-n` method of specifying multiple model runs, as `n_runs` is decremented as normal.

I did try specifying the number of resubmits as a command line option, but I don't think it works as well.

### DIFF
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -59,8 +59,9 @@ def setup(basepath=DEFAULT_BASEPATH):
     # the problematic variable here.  But a more general solution would be nice
     # someday.
 
-    bash_func_module = os.environ['BASH_FUNC_module()']
-    os.environ['BASH_FUNC_module()'] = bash_func_module.replace('\n', ';')
+    if 'BASH_FUNC_module()' in os.environ:
+        bash_func_module = os.environ['BASH_FUNC_module()']
+        os.environ['BASH_FUNC_module()'] = bash_func_module.replace('\n', ';')
 
 
 def module(command, *args):
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -85,6 +85,12 @@ class Access(Model):
 
                     prior_cpl_fpath = os.path.join(model.prior_restart_path,
                                                    cpl_fname)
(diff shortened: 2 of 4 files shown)


aidanheerdegen:
Happy to change the config parameter name (`runspersub`), couldn't think of anything better.

marshallward:
So this just does output000 output 001 output002 etc as separate jobs instead of serially?  Is that right?

aidanheerdegen:
Instead of serially? They are still serial, just within the "envelope" of a single PBS submission.

Actually, there is a bug when `runspersub=1`, so don't merge yet.

marshallward:
Ahh I see, sorry, ok.  You want me to test it?  Not so easy to do from here.   But I can just merge it in.

marshallward:
evermind deletd comment (and this one)

aidanheerdegen:
Fixed the looping logic bug. Seems to work. I have tested it, but happy for any sanity checks.

aidanheerdegen:
Andy H is running now, so we can wait to make sure it works in production mode before merging.

marshallward:
ok lets do it when i get back late next week if theres no rush

aidanheerdegen:
Ok, sounds good.

aidanheerdegen:
Added in a couple of extra changes, one which broke Paul's tcsh setup, and another which prevented Andy running from an existing access-om2 run, detailed here https://github.com/marshallward/payu/issues/91

aidanheerdegen:
Oh yeah, I should have said @marshallward that the runspersub option has been tested and works with MOM-SIS and access-om2. The error Andy was seeing was due to the bug mentioned above.

marshallward:
OK so I'll merge this and close #89, right?

aidanheerdegen:
Yeah, but I already closed https://github.com/marshallward/payu/issues/91

Issue #99, opened 2018-02-16
cice5 iced restart files grow in number


aidanheerdegen:
The `cice5` driver developed for `access-om2` keeps more and more `iced.*` restart files the longer a run goes on.

There is some logic to remove old restarts, but the break statement here:

https://github.com/marshallward/payu/blob/master/payu/models/cice5.py#L89

Means only the last restart is deleted. Depending on the frequency of output this may lead to the number of restarts growing with time. The restart directory of a 1 deg simulation may be of the order of 20GB by year 400, with all restarts using 1.3TB.

Do we need more than the last restart file, which is named in `ice.restart_file`?

PR #100, opened 2018-02-23
Fix issue #99, number of ice restart files growing with time

aidanheerdegen:
This is a fix for https://github.com/marshallward/payu/issues/99

The number of iced restarts grows with time because they aren't all properly deleted. I think the previous logic assumed only one restart needed to be deleted. In the case where more than one restart was created the number of restarts would grow.

I have tested with a 1 deg ACCESS-OM model and this is the listing of restarts after 4 resubmissions

```bash
ls  -1 archive/restart27[2-7]/ice/iced.*
archive/restart272/ice/iced.0541-04-01-00000.nc
archive/restart273/ice/iced.0541-05-01-00000.nc
archive/restart274/ice/iced.0541-06-01-00000.nc
archive/restart275/ice/iced.0541-07-01-00000.nc
```

So one restart for every restart directory. I had a feeling that `CICE` could use multiple restarts, but I can't recall why I think this.

Any comments @nicjhan?

### DIFF
--- a/payu/models/cice5.py
+++ b/payu/models/cice5.py
@@ -83,10 +83,11 @@ class Cice5(Cice):
 
         # Delete the old restart file (keep the one in ice.restart_file)
         for f in self.get_prior_restart_files():
-            if 'iced.' in f:
-                assert f != res_name
+            if f.startswith('iced.'):
+                if f == res_name:
+                    continue
                 os.remove(os.path.join(self.restart_path, f))
-                break
+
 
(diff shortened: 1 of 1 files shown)


marshallward:
Aidan, happy to trust you on this, feel free to merge whenever you like.

aidanheerdegen:
Is the logic correct here @nicjhan? Only one ice restart file per restart dir?

PR #123, opened 2018-09-25
Add support for fast collation with FMS models (mppnccombine-fast)

aidanheerdegen:
- Add support for mppnccombine-fast
- Move collate options to dict

Addresses https://github.com/marshallward/payu/issues/121 and https://github.com/marshallward/payu/issues/122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-    - "2.6"
     - "2.7"
     - "3.4"
     - "3.5"
@@ -13,10 +12,6 @@ before_script:
     - pip install -r requirements_test.txt
 script:
     - coverage run --source payu test_payu.py
-
-matrix:
-    allow_failures:
-        - python: "2.6"
 after_success:
     - coverage report -m
     - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,10 @@ before_script:
     - pip install -r requirements_test.txt
 script:
     - coverage run --source payu test_payu.py
+
+matrix:
+    allow_failures:
+        - python: "2.6"
 after_success:
     - coverage report -m
     - coveralls
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -638,13 +638,8 @@ class Experiment(object):
                 if os.path.isdir(res_path):
                     shutil.rmtree(res_path)
 
-        do_collate = self.config.get('collate', {})
-
-        # Support legacy config
-        if type(do_collate) is bool:
-            do_collate = {'enable': do_collate}
-
-        if do_collate.get('enable', True):
+        collate_config = self.config.get('collate', True)
+        if collate_config.get('enable', True):
             cmd = 'payu collate -i {}'.format(self.counter)
             sp.check_call(shlex.split(cmd))
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -638,7 +638,7 @@ class Experiment(object):
                 if os.path.isdir(res_path):
                     shutil.rmtree(res_path)
 
-        collate_config = self.config.get('collate', True)
+        collate_config = self.config.get('collate', {})
         if collate_config.get('enable', True):
             cmd = 'payu collate -i {}'.format(self.counter)
             sp.check_call(shlex.split(cmd))
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -638,11 +638,11 @@ class Experiment(object):
                 if os.path.isdir(res_path):
                     shutil.rmtree(res_path)
 
-        do_collate = self.config.get('collate',{})
+        do_collate = self.config.get('collate', {})
 
         # Support legacy config
         if type(do_collate) is bool:
-            do_collate = {'enable' : do_collate}
+            do_collate = {'enable': do_collate}
 
         if do_collate.get('enable', True):
             cmd = 'payu collate -i {}'.format(self.counter)
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -638,7 +638,13 @@ class Experiment(object):
                 if os.path.isdir(res_path):
                     shutil.rmtree(res_path)
 
-        if self.config.get('collate', True):
+        do_collate = self.config.get('collate',{})
+
+        # Support legacy config
+        if type(do_collate) is bool:
+            do_collate = {'enable' : do_collate}
+
+        if do_collate.get('enable', True):
             cmd = 'payu collate -i {}'.format(self.counter)
             sp.check_call(shlex.split(cmd))
 
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -17,6 +17,7 @@ from itertools import count
 from payu.models.model import Model
 from payu import envmod
 
+
 def cmdthread(cmd, cwd):
     # This is run in a thread, so the GIL of python makes it sensible to
     # capture the output from each process and print it out at the end so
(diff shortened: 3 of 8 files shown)


aidanheerdegen:
The checks have failed because the py2.6 test fails:
```
pycparser requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 2.6.9
```

marshallward:
At home with Finn, but will look tonight

.travis.yml
```diff
@@ -13,6 +13,10 @@ before_script:
     - pip install -r requirements_test.txt
 script:
     - coverage run --source payu test_payu.py
+
+matrix:
+    allow_failures:
+        - python: "2.6"
```
marshallward: I think just remove it from the tests, rather than run-and-ignore.  We may be stuck with a 2.6 environment, but it's just becoming impossible to get support from 2.6 modules.

payu/models/fms.py
```diff
@@ -12,9 +12,10 @@
 import shlex
 import subprocess as sp
 import sys
+from itertools import count
```
marshallward: Is this used?  I don't see it

payu/experiment.py
```diff
@@ -638,7 +638,13 @@ def archive(self):
                 if os.path.isdir(res_path):
                     shutil.rmtree(res_path)
 
-        if self.config.get('collate', True):
+        do_collate = self.config.get('collate',{})
```
marshallward: I think the place to do the first part of this (convert bool to dict) might be in the `read_config` function of `payu.fsops.py`.

payu/models/fms.py
```diff
@@ -65,9 +66,20 @@ def collate(self):
...
+
+        if type(collate_conf) is bool:
+            # Cycle through old collate config and convert to newer dict format
+            collate_conf = {'enable' : collate_conf}
+            for key in ['exe','flags','ignore']:
+                oldkey = "collate_{}".format(key)
+                print("Use of this key is deprecated: {}. Use collate dictionary and subkey {}".format(oldkey,key))
+                if oldkey in self.expt.config:
+                    collate_conf[key] = self.expt.config[oldkey]
+
```
marshallward: Similarly, I think it might be possible to convert these keys inside `read_config`?

marshallward: Also may be useful to do a `pop` instead of a read from `config`, then replace `config['collate']` with the updated dict.
Though this could potentially cause problems elsewhere.

payu/models/fms.py
```diff
@@ -106,13 +133,25 @@ def collate(self):
...
+            nprocesses = int(collate_conf.get('threads', 1))
+        else:
+            nprocesses = int(collate_conf.get('threads', cpucount))
+
+        ncpusperprocess = int(cpucount/nprocesses)
+
+        if ncpusperprocess == 1 and mpi:
+            print("Warning: running collate with mpirun with a single processor")
+
+        pool = multiprocessing.Pool(processes=nprocesses)
```
marshallward: Do you want to use pools with the MPI version?  Probably OK but could have some unexpected issues.

payu/subcommands/collate_cmd.py
```diff
@@ -23,10 +23,30 @@ def runcmd(model_type, config_path, init_run, n_runs, lab_path, dir_path):
...
+    collate_conf = pbs_config.get('collate',{})
+
+    # Cycle through old collate config and convert to newer dict format
+    if type(collate_conf) is bool:
+        collate_conf = {'run' : collate_conf}
+        for key in ['queue','ncpus','walltime','mem']:
+            oldkey = "collate_{}".format(key)
+            print("Use of this key is deprecated: {}. Use collate dictionary and subkey {}".format(oldkey,key))
+            if oldkey in pbs_config:
+                collate_conf[key] = pbs_config[oldkey]
```
marshallward: I was going to again say "move this to `read_config`, but I notice that `cli.get_config` does not use `read_config`, even though it's extremely similar.

Might be worth merging these functions somehow.

payu/models/fms.py
marshallward: Also, it's probably OK to just transform everything starting with `collate_`, unless we are just being extra-safe here about our legacy namespace.

payu/models/fms.py
```diff
@@ -95,6 +120,8 @@ def collate(self):
         tile_fnames = [f for f in os.listdir(self.output_path)
                        if f[-4:].isdigit() and f[-8:-4] == '.nc.']
 
+        tile_fnames.sort()
```
marshallward: What is the sorting for?  (Just wondering, not important)

aidanheerdegen: The list of filenames isn't sorted at all, so unlike a glob is in pretty much random order. It makes it easier to visually scan the command that is echoed to stdout with this change, and it is what a user would expect to see IMO.

payu/models/fms.py
aidanheerdegen: I've tested it and it works. It allows for the most flexibility. Can you a priori think of any issues that might arise? Then I could test for it.

payu/models/fms.py
aidanheerdegen: I guess I was only doing the ones that I were explicitly being used in the FMS model. The other ones are to do with PBS submission. I don't mind either, but the way I'm doing it here it is explicit about what is being used. Adding a new flag won't be affected, as it won't have a legacy equivalent.

aidanheerdegen: I agree it would be nice to make all these transformations in a single place, which is what you're suggesting in other comments.

payu/subcommands/collate_cmd.py
aidanheerdegen: The major difference seems to be in their error handling. The `cli` version is willing to have config_path as `None`. Is this important?
```python
except (TypeError, IOError) as exc:
        if config_path is None:
            config = {}
        elif type(exc) == IOError and exc.errno == errno.ENOENT:
            print('payu: error: Configuration file {} not found.'
                  ''.format(config_path))
            sys.exit(errno.ENOENT)
        else:
            raise

    return config```

File: payu/models/fms.py (hunk 5 above)
Comment (marshallward) [reply payu/models/fms.py]: I can't think of any specific problem.  I am just thinking that we are adding a process inbetween the parent process launched by PBS and the pool subprocess which is launching `mpirun`, and wondered if the PBS server and MPI process were fully aware of each other.

It's all probably OK, just wondering how it all plays out, or if another scheduler would object here.

Comment (aidanheerdegen) [reply payu/models/fms.py]: I agree it is a bit of a dog's breakfast, but actually it is sort of easier and cleaner to do it this way for the moment. What is the option? A great big
```python
if mpi:
else:
```
conditional?

payu/subcommands/collate_cmd.py
marshallward: This may not have even been tested... I don't think I've run anything with no `config.yaml`... 

Using `payu.fsops.read_config` seems to "work" on some basic cases, so maybe let's just try replacing `get_config` with `read_config` in `collate_cmd.py` and see how it goes?

aidanheerdegen: I assume I'll replace it wherever it appears:
```
$ grep -ir get_config payu 
payu/subcommands/run_cmd.py:    pbs_config = cli.get_config(config_path)
payu/subcommands/profile_cmd.py:    pbs_config = cli.get_config(config_path)
payu/subcommands/collate_cmd.py:    pbs_config = cli.get_config(config_path)
```

marshallward: Going to open an issue about `read_config` and `get_config`; there's more I can say but it's becoming a distraction.  For now, let's exclusively use `read_config` and phase out `get_config`, starting with `collate_cmd.py`.

aidanheerdegen: Oops. I just replaced them all.

payu/models/fms.py
marshallward: No it's fine, lets leave it for now

PR #146, opened 2018-12-19
Add file tracking to payu

aidanheerdegen:
Add input/restart and executable file tracking to payu using the yamanifest module to create YaML based manifest files which are tracked by git. 

When they exist, manifest files are used to populate the input directories, restarts (when required) and links to the executables. This should allow for easier configuration sharing, as the experiment just needs to be cloned and can be run without further configuration changes.

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -26,6 +26,7 @@ from payu.fsops import mkdir_p, make_symlink, read_config
 from payu.models import index as model_index
 import payu.profilers
 from payu.runlog import Runlog
+from payu import Manifest
 
 # Environment module support on vayu
 # TODO: To be removed
@@ -39,7 +40,7 @@ default_restart_history = 5
 
 class Experiment(object):
 
-    def __init__(self, lab):
+    def __init__(self, lab, reproduce=False):
         self.lab = lab
 
         # TODO: replace with dict, check versions via key-value pairs
@@ -78,6 +79,9 @@ class Experiment(object):
             model.set_input_paths()
 
         self.set_output_paths()
+
+        # Initialize manifest
+        self.manifest = Manifest(self, reproduce=reproduce)
 
         # Miscellaneous configurations
         # TODO: Move this stuff somewhere else
@@ -166,7 +170,7 @@ class Experiment(object):
                                         for d in restart_dirs
                                         if d.startswith('restart')])
             else:
-                # uepeat runs do not generate restart files, so check outputs
+                # repeat runs do not generate restart files, so check outputs
                 try:
                     output_dirs = [d for d in os.listdir(self.archive_path)
                                    if d.startswith('output')]
@@ -177,6 +181,7 @@ class Experiment(object):
                         raise
 
                 # First test for restarts
+                # Now look for output directories
                 if output_dirs:
                     self.counter = 1 + max([int(d.lstrip('output'))
                                             for d in output_dirs
@@ -273,6 +278,7 @@ class Experiment(object):
 
         for model in self.models:
             model.set_model_pathnames()
+            model.set_local_pathnames()
 
         # Stream output filenames
         # TODO: per-model output streams?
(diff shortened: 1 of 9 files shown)


aidanheerdegen:
Outstanding problems:

- [ ] manifest names are dodgy
- [ ] changing the input path in `config.yaml` will not trigger a new input manifest to be populated
- [ ] not tested against any other models other than MOM5 and ACCESS-OM2

aidanheerdegen:
Fixed reproduce checks and added information on non-matching hashes.

aidanheerdegen:
WRT to this item: "reproduce run doesn't pick up run counter correctly"

This code block correctly deduces the run counter:

https://github.com/marshallward/payu/blob/3fe5e1a9bf6fca562db250886eb71b6b6627f5ff/payu/manifest.py#L216

But it is too late, other variables have already been set with the default value of zero.

I'm not sure what would need to be updated to do this correctly, or the original set_counters routine would have to read the restart manifest?

aidanheerdegen:
WRT to this: "changing the input path in config.yaml will not trigger a new input manifest to be populated"

All the user has to do is delete the manifest and it will regenerate it from the new input path, but it may not be obvious, so we might want some way to detect this condition and warn the user.

Could store then input path in the manifest, and check if it has changed, then purge manifest and start again? 

Could check if the input path appears in any of the inputs, and warn user if not? I don't think this should be an error condition, as it might be legit that the manifest has been hand edited with paths that aren't in the input. Not a great idea, but shouldn't be not allowed.

aidanheerdegen:
Re: "existing restart manifests in control directory are ignored when not reproduce. This is arguably a feature."

Could support reading in the existing manifest, but would need to remove all entries that are no longer relevant. Only relevant if `setup` is called twice, the second time the hashes are not required to be calculated, but are anyway. 

If `setup` should write the restart file to the restart directory then the other issue with reading restarts from restart directory could allow that to be re-read. 

A downside: can't always guarantee can write into a restart directory

aidanheerdegen:
WRT this: "changing the input path in config.yaml will not trigger a new input manifest to be populated"

Just added some code which will write the current git hash into the manifest file when it is dumped. This is useful info anyway, but can be used to query state of experiment when manifest last updated.
In this way it could be used to compare config.yaml, see if input directories have been updated and warn user.

This would mean using something like 
```
git cat-file -p 1fcef4b5da961bebb9620520ceb3757546847a80 config.yaml
```
to get an old config, parse it and compare it to the current one. Or maybe there is a better solution @marshallward?

aidanheerdegen:
I force pushed a commit after rebasing to the current master, as I wanted to pick up and test some of Marshall's changes. Hope this doesn't stuff anyone around

aidanheerdegen:
Tested with MOM6 and MITgcm.

aidanheerdegen:
Re: 'changing the input path in config.yaml will not trigger a new input manifest to be populated'

This commit 
https://github.com/marshallward/payu/pull/146/commits/8a454ac0496f76f2d1155ef06164a1b851ea51d5 addresses this to an extent. It emits a warning if the last commit when `config.yaml` was changed is not an ancestor of the git commit id for the input manifest.

A more thorough check would be to look and see if the input paths have been altered in `config.yaml` between those two commits, but this would take a fair bit of work.

aidanheerdegen:
Re: 'a deleted input file produces an error, but is not removed from the manifest'

This is fixed in 
https://github.com/marshallward/payu/pull/146/commits/931eb4dc1940a5031cf4850af16e0fc68ebe3c78

marshallward:
looks like there is a new `yamanifest` dependency, could you add those to `setup.py`?

aidanheerdegen:
Is there? Ok, I'll take a look

marshallward:
Seems that yamanifest is failing now?
```
$ PYTHONPATH=$(pwd) coverage run --source payu bin/payu
  File "bin/payu", line 4, in <module>
    cli.parse()
  File "/home/travis/build/marshallward/payu/payu/cli.py", line 37, in parse
    subcmds = [importlib.import_module(mod) for mod in modnames]
  File "/opt/python/2.7.14/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/travis/build/marshallward/payu/payu/subcommands/archive_cmd.py", line 3, in <module>
  File "/home/travis/build/marshallward/payu/payu/experiment.py", line 29, in <module>
    from payu.manifest import Manifest
  File "/home/travis/build/marshallward/payu/payu/manifest.py", line 19, in <module>
    from yamanifest.manifest import Manifest as YaManifest
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/yamanifest/__init__.py", line 1, in <module>
    from .manifest import HashExists, FilePathNonexistent, HashNonexistent, Manifest
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/yamanifest/manifest.py", line 159
    if 'hashes' not in self.data[filepath]
                                         ^
SyntaxError: invalid syntax
The command "PYTHONPATH=$(pwd) coverage run --source payu bin/payu" exited with 1.
```
Not sure how these are coming up, is this the wrong version of yamanifest?

aidanheerdegen:
Sorry. I'll have to look into on Tuesday, don't waste any more time.

aidanheerdegen:
Ok, fixed `yamanifest` (sorry about that). So tests now passing.

Happy to merge @marshallward?

marshallward:
Yep, lets have some fun.

PR #150, opened 2019-01-17
Dump some job information stats

aidanheerdegen:
Dump some job information stats to a yaml file to obviate the need to scrape output files for the information.


aidanheerdegen:
Whaddya reckon @aekiss? Is this going to tick all the boxes?

It runs `qstat -f` on the `PBS_JOBID`, grabs all the key/value pairs, bungs them in a dict and dumps them to a YAML file. 

It adds these entries:

```yaml
PAYU_FINISH_TIME: '2019-01-17T17:26:47.142074'
PAYU_JOB_STATUS: 0
PAYU_PATH: /home/502/aph502/.local/bin
PAYU_START_TIME: '2019-01-17T17:26:42.135421'
PAYU_WALLTIME: 5.006653 s
```

Does it have all the things you would want? Maybe this is overkill?

The full file:

```yaml
Checkpoint: u
Error_Path: raijin3:/home/502/aph502/payu/mitgcm/test/lala1.e4797005
Hold_Types: n
Job_Name: lala1
Job_Owner: aph502@raijin3
Join_Path: n
Keep_Files: n
LD_LIBRARY_PATH: /g/data3/hh5/public/apps/miniconda3/envs/analysis27-18.10/lib
Mail_Points: a
Output_Path: raijin3:/home/502/aph502/payu/mitgcm/test/lala1.o4797005
PAYU_FINISH_TIME: '2019-01-17T17:26:47.142074'
PAYU_JOB_STATUS: 0
PAYU_PATH: /home/502/aph502/.local/bin
PAYU_START_TIME: '2019-01-17T17:26:42.135421'
PAYU_WALLTIME: 5.006653 s
PBS_NCI_FS_GDATA1: '0'
PBS_NCI_FS_GDATA1A: '0'
PBS_NCI_FS_GDATA1B: '0'
PBS_NCI_FS_GDATA2: '0'
PBS_NCI_FS_GDATA3: '0'
PBS_NCI_HT: '0'
PBS_NCI_IMAGE: ''
PBS_NCI_JOBFS_LOCAL: 104857600b
PBS_NCI_WD: '1'
PBS_NCPUS: '4'
PBS_NGPUS: '0'
PBS_O_HOME: /home/502/aph502
PBS_O_HOST: raijin3
PBS_O_LANG: C
PBS_O_LOGNAME: aph502
PBS_O_MAIL: /var/spool/mail/aph502
PBS_O_PATH: /g/data3/hh5/public/apps/miniconda3/envs/analysis27-18.10/bin:/projects/v45/cmstools/bin:/apps/git/2.9.5/bin:/apps/nco/4.7.7/bin:/apps/python/2.7.11/bin:/opt/bin:/bin:/usr/bin:/opt/pbs/default/bin:.:/home/502/aph502/bin:/home/502/aph502/v45_short/spack/bin:/apps/ncview/2.1.2/bin:/home/502/aph502/bin:/home/502/aph502/v45_short/spack/bin:/g/data3/hh5/public/apps/miniconda3/bin
PBS_O_QUEUE: express
PBS_O_SHELL: /bin/bash
PBS_O_SYSTEM: Linux
PBS_O_WORKDIR: /home/502/aph502/payu/mitgcm/test
PBS_VMEM: '4294967296'
PROJECT: x77
PYTHONPATH: /projects/v45/python
Priority: '0'
Rerunable: 'False'
Resource_List.HT: '0'
Resource_List.jobfs: 104857600b
Resource_List.jobfs_local: 104857600b
Resource_List.jobprio: '100017.5938'
Resource_List.mem: 4294967296b
Resource_List.mpiprocs: '4'
Resource_List.ncpus: '4'
Resource_List.nodect: '1'
Resource_List.place: free
Resource_List.queue_charge_rate: '3'
Resource_List.select: 1:ncpus=4:mpiprocs=4:mem=4294967296:job_tags=express-6h:jobfs_local=104857600
Resource_List.walltime: 00:30:00
Resource_List.wd: '1'
Submit_arguments: -q express -P x77 -l walltime=0:30:00 -l ncpus=4 -l mem=4GB -N lala1
  -l wd -j n -v PYTHONPATH=/projects/v45/python,PAYU_PATH=/home/502/aph502/.local/bin,LD_LIBRARY_PATH=/g/data3/hh5/public/apps/miniconda3/envs/analysis27-18.10/lib
  -- /g/data3/hh5/public/apps/miniconda3/envs/analysis27-18.10/bin/python /home/502/aph502/.local/bin/payu-run
argument_list: <jsdl-hpcpa:Argument>/home/502/aph502/.local/bin/payu-run</jsdl-hpcpa:Argument>
comment: Job was sent for execution at Thu Jan 17 at 17:26 on (r51:ncpus=4:mem=4194304kb:jobfs_local=102400kb)
ctime: Thu Jan 17 17:23:35 2019
eligible_time: 00:00:00
etime: Thu Jan 17 17:23:35 2019
exec_host: r51/4*4
exec_vnode: (r51:ncpus=4:mem=4194304kb:jobfs_local=102400kb)
executable: <jsdl-hpcpa:Executable>/g/data3/hh5/public/apps/miniconda3/envs/analysis27-18.10/bin/python</jsdl-hpcpa:Executable>
group_list: x77
job_state: R
mtime: Thu Jan 17 17:26:37 2019
project: x77
qtime: Thu Jan 17 17:23:35 2019
queue: express-def
run_count: '1'
server: r-man2
substate: '41'
```

aidanheerdegen:
Unfortunately `qtime` and `stime` are not reliable. `qstat` does include some resources used fields, but they get updated only sporadically and can't be relied upon to exist all the time.

aidanheerdegen:
Are we absolutely sure we don't want to commit this to the `git` repo? :)

marshallward:
Looks like a good start, but I'd recommend moving those functions from fsops to another module, maybe something devoted to schedulers.  (`scheduler/pbs.py`?).

Also `dump_yaml` looks a little too trivial to be a function.

aidanheerdegen:
You've hurt `dump_yaml`'s feelings. 

I did think about something PBS specific, but that started down the path of making a class for the job scheduling stuff, and I thought that was another PR. But maybe a start is to dump something in the file you suggest and then bling it up later on.

Do we want to do any pruning of that dump? There is a lot of junk in there, but I don't want to white list stuff, in case some useful info I don't know about right now pops up in later versions of `qstat`. Could blacklist tings we know we definitely don't want. Could most probably filter out anything that is empty, but that is only one entry so far:
```yaml
PBS_NCI_IMAGE: ''
```

aidanheerdegen:
Also, is there anything else that needs to go in there? Like maybe `module` info?

aidanheerdegen:
In all cases the info from `qstat` will be out of date compared to that dumped to the PBS output file. For longer runs maximum memory usage is likely to be correct, or very near, as it is sampled every minute or so. 

The correct values could be obtained by running a job *after* completion with a `qstat -x -f` query, but then you're probably better off just scraping the output file as is already done by @aekiss' code.

marshallward:
I too was thinking that something like `scheduler/pbs.py` could be a holding place for something more sophisticated in the future (a class, etc).  But I am also trying discourage turning `fsops.py` into a dumping ground for every utility function.

As for other info, it would be very good to snapshot `os.environ`, since NCI help often asks for this info.      But I would expect this to grow or change over time anyway.

aidanheerdegen:
I was putting `os.environ` in there as well, but it looked pretty junky.

Should that be dumped into a separate file, say `environ.yaml`? `job info.yaml` is literally about the job itself. Too many files?

Also .. what do you think about an option to add `job info.yaml` to the `git` repo. Just an option ... can even default `False` ... maybe

aidanheerdegen:
Ok, after talking with @aekiss I think I've realised we want to save some information as part of the model input, and commit before running. So definitely a dump of `os.environ` for example. This actually won't change much, so not a big burden as far as size goes, but it will be really useful if something stops working to see if there was an associated change in the environment that might be the proximate cause.

aekiss:
these could also be useful?
LD_LIBRARY_PATH
PBS_O_PATH
PBS_O_QUEUE
PBS_O_WORKDIR 
PROJECT
PYTHONPATH
Resource_List.mem
Resource_List.walltime

maybe these too? or are they per-node?
PBS_NCI_JOBFS_LOCAL
PBS_NCPUS

but is there any reason not to just save everything? That way the script won't break if some of the items cease to exist in the future. It's not a lot of data (particularly if we don't commit to git) and who knows what we might want in the future. 

I think the decision to include/exclude should be made in the scripts that read this YAML dump file to generate run summaries or whatever, so the YAML dump should just have everything.

marshallward:
Separate files seems ok for me (`environ.yaml`, `pbs.yaml`, etc).  If it were me, I'd save as much as possible.  I'm guessing we're talking <1MiB here?

marshallward:
I'd also probably not add any of this to the runlog repo (is that what you mean?)  It's just output, like anything else.

aidanheerdegen:
Size is tiny. A few KB
```
$ env > env.tst
$ ls -lh env.tst 
-rw-r----- 1 aph502 v45 4.1K Jan 18 11:52 env.tst
```
And it wouldn't change much, so wouldn't be a large contributor to repo size.

I get your objection to adding the timing info to the runlog, that is output, but the environment is input. It can't be reproduced from the other inputs, it is potentially unique at the time of a run.

marshallward:
Runlog is information required to run the model.  Environment is just a snapshot of when it was run.

aidanheerdegen:
I think an argument could be made that the environment is an ephemeral input ... and I'm making it

marshallward:
Are you suggesting I run your job with your environment variables?

aidanheerdegen:
No. I'm suggesting we could run the same job, our environments are different, and that might lead to different results. I'd hope not, but it could happen. The environment is an input that we can't specify ... or maybe we could?

My feeling is, if we can't control it, log it.

marshallward:
You are logging the environment variables, it's in the output.  But you are not using this file to set anything, so why put it in the runlog, which is explicitly for tracking and sharing input files?

aidanheerdegen:
I would say it the runlog is for tracking and sharing model state.

The reason for adding it to the repo is that the information is then retained independent of the output directories.

For example, test runs or failed runs, where the output directories are often deleted (space constraints) mean the information is then lost. 

If we're not keeping a run DB independent of the configs, then this is a way to not lose that information.

If it makes you feel any better, no-one else agrees with me. BUT THAT JUST MAKES ME MORE DETERMINED! ;)

marshallward:
A lack of output management is not a reason to flag something as input, when it is not a user-defined input.  (Note: I am overlooking the explicitly set variables here, which perhaps should be saved in some way.)

On top of that, this may even be a slight security risk.  I am not sure that what we really want to be doing is snapshotting our environment variables, putting them in repos, and then sharing them with the entire world.  I know that I personally do not want to do this.

We all agree that the information is important, and should be saved, but you are also trying to redefine the runlog in the process, whose contours are still not yet clear.  This really ought to be a separate discussion.

The easiest path forward is to save the file, and keep it in the output.  If nothing changes, then we have progress forward.  If we change it and start putting extensive environment snapshots into the runlog (which I would not count on btw), then we can make this change at a later time.

aidanheerdegen:
Yeah agreed. I was implementing that already, but thanks for taking the time out to consider it. I really wanted to be convinced, and I think I have been now. Tah.

marshallward:
Sorry just merged my last PR, didn't want to interfere with this

aidanheerdegen:
I should never do this stuff in a rush. So I rebased after your merged, and I may well have stuffed it up.

aidanheerdegen:
Did some quick tests. Seems to be working ok.

aidanheerdegen:
I'm still working on this BTW, so please don't merge without consultation.

aidanheerdegen:
Ok this is ready to review @marshallward when you have a chance.

payu/experiment.py
```diff
@@ -558,13 +576,37 @@ def run(self, *user_flags):
...
+            'PAYU_CURRENT_RUN': self.counter,
+            'PAYU_N_RUNS':  self.n_runs,
+            'PAYU_JOB_STATUS': rc,
+            'PAYU_START_TIME': self.start_time.isoformat(),
+            'PAYU_FINISH_TIME': self.finish_time.isoformat(),
+            'PAYU_WALLTIME': "{0} s".format(
+              (self.finish_time - self.start_time).total_seconds()
+             ),
+          }
+        )
```
marshallward: Can you 4-space this indentation?

payu/experiment.py
```diff
@@ -614,7 +667,10 @@ def run(self, *user_flags):
             self.n_runs -= 1
 
         # Move logs to archive (or delete if empty)
-        for f in (self.stdout_fname, self.stderr_fname):
+        for f in (self.stdout_fname,
+                  self.stderr_fname,
+                  self.job_fname,
+                  self.env_fname):
```
marshallward: Would be good to pre-form this tuple and name it something like `output_fnames`

payu/fsops.py
```diff
@@ -127,3 +127,7 @@ def patch_lustre_path(f_path):
             f_path = './' + f_path
 
     return f_path
+
+def dump_yaml(info, fname):
```
marshallward: I'm still not convinced this function should exist.

payu/experiment.py
```diff
@@ -19,10 +19,13 @@
 import shutil
 import subprocess as sp
 import sysconfig
+import datetime
```
marshallward: Would be good to keep these alphabetized within their group.

payu/experiment.py
```diff
@@ -385,8 +397,7 @@ def setup(self, force_archive=False):
         self.manifest.make_links()
 
         # Copy manifests to work directory so they archived on completion
-        manifest_path = os.path.join(self.work_path, 'manifests')
-        self.manifest.copy_manifests(manifest_path)
+        self.manifest.copy_manifests(os.path.join(self.work_path, 'manifests'))
```
marshallward: Use two lines here?
```
manifest_path = os.path.join(self.work_path), 'manifests')
self.manifest.copy_manifests(manifest_path)
```
Or, maybe just add `self.manifest.dir` and set it during init?

payu/experiment.py
```diff
@@ -596,6 +638,17 @@ def run(self, *user_flags):
...
                 shutil.copyfile(src, dest)
 
+            for fname in (self.job_fname, self.env_fname):
+
+                src = os.path.join(self.control_path, fname)
+
+                # Assumes .yaml extension
+                dest = os.path.join(error_log_dir,
+                                    fname[:-5] + '.' + job_id + fname[-5:])
+                print(src, dest)
```
marshallward: Testing diagnostic?  Delete?

payu/experiment.py
```diff
@@ -596,6 +638,17 @@ def run(self, *user_flags):
 
                 shutil.copyfile(src, dest)
 
+            for fname in (self.job_fname, self.env_fname):
+
+                src = os.path.join(self.control_path, fname)
+
+                # Assumes .yaml extension
+                dest = os.path.join(error_log_dir,
+                                    fname[:-5] + '.' + job_id + fname[-5:])
```
marshallward: Consider using `splitext` here, and maybe bail if it's not a YAML file (unless there's some case where it ought to work?)

Something like this:
```
base, ext = os.path.splitext(fname)
assert(ext == '.yaml')
job_fname = ''.join(base, '.', job_id, ext)
dest = os.path.join(error_log_dir, job_fname)
```

payu/manifest.py
```diff
@@ -9,6 +9,13 @@
 """
 from __future__ import print_function, absolute_import
 
+# Local
+from payu import envmod
+from payu.fsops import make_symlink
+
+# External
+from yamanifest.manifest import Manifest as YaManifest
+import yamanifest as ym
```
marshallward: I've been hoping to eliminate these abbreviated modulenames.  Not too fussed about it though.

payu/runlog.py
```diff
@@ -274,3 +274,29 @@ def authenticate(self):
...
+    cmd = ['git', 'rev-parse', 'HEAD']
+
+    try:
+        with open(os.devnull, 'w') as devnull:
+            revision_hash = subprocess.check_output(
+              cmd,
+              cwd=dir,
+              stderr=devnull
+            )
+        if sys.version_info.major > 2:
```
marshallward: Should not need an explicit version check, should be ok to do `.decode('ascii`)` regardless of Py 2 vs 3.

payu/runlog.py
```diff
@@ -274,3 +274,29 @@ def authenticate(self):
...
+    Return commit hash for HEAD of checked out branch of the
+    specified directory.
+    """
+
+    cmd = ['git', 'rev-parse', 'HEAD']
+
+    try:
+        with open(os.devnull, 'w') as devnull:
+            revision_hash = subprocess.check_output(
+              cmd,
```
marshallward: Indentation is not 4-space

payu/experiment.py
aidanheerdegen: Heh. I only just changed it to that as I thought it's not code indentation, so might be good to differentiate, and PEP8 didn't complain so I figured it was ok.

payu/experiment.py
marshallward: haha, just noticed this was previously two lines :D

payu/fsops.py
aidanheerdegen: It is now used more than once, you'd rather have the code inlined?

marshallward: I think so, yeah.  It's two lines and is basically the mechanism exposed by the module.

Not gonna die on this hill though, if you really like it then it's fine.

payu/experiment.py
aidanheerdegen: So change it back? Or set `self.manifest.dir`?

payu/experiment.py
aidanheerdegen: I just copied your code. You have this for the `.err` and `.out`. I have coalesced them into a single code block now, but happy to delete the `print` statement if you don't think it is still necessary.

payu/experiment.py
aidanheerdegen: Hah! Just did this before I noticed your comment.

payu/manifest.py
aidanheerdegen: Was redundant in any case. Deleted.

payu/experiment.py
marshallward: Yeah lets revert back?

payu/experiment.py
marshallward: Haha I see... I think that print statement is weird, but I think it was to give a bit of feedback on the sweep?

I guess leave for now, if it's weird we can take it out.

payu/runlog.py
aidanheerdegen: Subprocess `check_output` returns strings in python2 and byte objects in python3. If we do `.decode('ascii')` in python2 we end up with a `unicode` type, which can be problematic. We want it to be a string.

Happy to defer, I hate this stuff, but I'm pretty sure I had to do it this way.

marshallward: Oh huh, you're right, I did not know that.  But it will still be a unicode string in 3 anyway, right?

marshallward: I mean, we are talking about hashes anyway, so should be ASCII regardless, right?

aidanheerdegen: For python3 the `.decode('ascii')` will transform the byte object to a str (not unicode, python3 no longer makes that distinction). python2 returns a string, so doesn't need the `decode` step.

aidanheerdegen: Like I said, I hate this stuff. Most people don't bother supporting both, which is where the problem lies.

marshallward: I had thought a Python 3 `str` is actually a unicode string, but I guess it's not that important.

Once upon a time I had come up with some way to deal with this that didn't involve version checks, but it's probably not important, can deal with it another time.

aidanheerdegen: Could check the return type I guess. Agreed this can be tackled at another time. At some point python2 support will be dropped, and this stuff can be removed with a nice fat

```python
assert(sys.version_info.major > 2)
```

somewere, or requiring a `python3` interpreter.

aidanheerdegen:
Oh bugger, just remembered I have to add one more thing to the dump

aidanheerdegen:
No scratch that, I already added the PBS job id. Good to merge if you're happy @marshallward

marshallward:
One more :)
```
    from payu.fsops import dump_yaml
ImportError: cannot import name dump_yaml
```

aidanheerdegen:
Ok, now should be good to go

aidanheerdegen:
Thanks for the review BTW

PR #153, opened 2019-01-22
Parse MITgcm config files to check they are correct namelists

aidanheerdegen:
Parse MITgcm config files to check they are correct namelists adding to list self.config_files.

Addresses https://github.com/marshallward/payu/issues/148.

### DIFF
--- a/payu/models/mitgcm.py
+++ b/payu/models/mitgcm.py
@@ -37,9 +37,23 @@ class Mitgcm(Model):
     def setup(self):
 
         # TODO: Find a better place to generate this list
-        self.config_files = [f for f in os.listdir(self.control_path)
+        files = [f for f in os.listdir(self.control_path)
                              if f.startswith('data')]
-        self.config_files.append('eedata')
+        files.append('eedata')
+
+        # Rudimentary check that matching files are namelists. Can only check if 
+        # namelist is empty. May excluded false positives, but these are # devoid 
+        # of useful information in that case
+        for fname in files:
+            nml_parser = f90nml.Parser()
+            nml_parser.comment_tokens += '#'
+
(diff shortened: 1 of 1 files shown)


payu/models/mitgcm.py
```diff
@@ -37,9 +37,23 @@ def __init__(self, expt, name, config):
...
+        for fname in files:
+            nml_parser = f90nml.Parser()
+            nml_parser.comment_tokens += '#'
+
+            data_nml = nml_parser.read(fname)
+            if len(data_nml) > 0:
+                self.config_files.append(fname)
+            else:
+                print("Excluding {0} from list of configuration files: "\
+                      "assumed to be not a namelist file (or empty)".format(fname))
```
marshallward: can probably ditch the slash, I think string concatenation is automatic inside of parentheses.

marshallward:
Sorry I have zonked out on this issue, does this mean that the list of valid input files can be determined from the namelists?

aidanheerdegen:
Oh, I dunno, was that a possibility? This is just parsing them and making sure they're valid namelist files. As you surmised, it will also discard valid namelist files with no content, but all they need is an empty group to be considered valid.

So this is considered valid:
```fortran
! Packages
 &PACKAGES
 &
```

payu/models/mitgcm.py
aidanheerdegen: Fixed in the PEP8 cleanup

marshallward:
Ah right, I see, I see.  Yeah that seems OK to me.

aidanheerdegen:
I have done some light testing, seems to work ok .. gad I need to get some tests in the repo. Feel free to merge

PR #158, opened 2019-01-23
Add option to collate restart files as well as output files

aidanheerdegen:
This will collate the previous restart, to avoid issues with half
collated restarts if collate doesn't complete before next run starts.

Also avoids any possible speed hit when using collated restarts.

This is basically https://github.com/marshallward/payu/pull/125/commits/1039acc9442ad2cada967cbbfbc07936cc0a46ce but applied to the previous restart.

I don't know why, but that other PR was merged, but then overwritten at some stage.

### DIFF
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -56,9 +56,14 @@ class Fms(Model):
 
     @staticmethod
     def get_uncollated_files(dir):
+
+        if not os.path.isdir(dir):
+            return []
+
         # Generate collated file list and identify the first tile
         tile_fnames = [f for f in os.listdir(dir)
                        if f[-4:].isdigit() and f[-8:-4] == '.nc.']
+
         # print("dir: ",tile_fnames)
         tile_fnames.sort()
         return tile_fnames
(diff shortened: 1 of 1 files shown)


payu/models/fms.py
```diff
@@ -125,28 +134,36 @@ def collate(self):
...
+
+        # print(mnc_tiles)
 
         if mpi and collate_config.get('glob', True):
             for t_base in mnc_tiles:
                 globstr = "{}.*".format(t_base)
                 # Try an equivalent glob and check the same files are returned
-                mnc_glob = fnmatch.filter(os.listdir(self.output_path,
-                                                     globstr))
+                mnc_glob = fnmatch.filter(os.listdir(self.output_path),
```
aidanheerdegen: This was the error I was referring to @marshallward

marshallward: oops sorry about that

aidanheerdegen: No worries. I didn't test it obviously.

aidanheerdegen:
Fixes https://github.com/marshallward/payu/issues/124

PR #160, opened 2019-02-01
Added support for ability of cice5 to define separate input and restart directories.

aidanheerdegen:
Added support for splitting input and restart directories in the `cice5` driver (actually mostly in the `cice` driver but there is no `cice4` version that supports this)

The `cice5` code change is here:

https://github.com/OceansAus/cice5/commit/465494bb551ec15ec3ec82308359e6c7d3ae28a5

This change tries to be backwards compatible. I have tested on an access-om2 1 degree model and it seems to work fine.

To use this you also need to make a couple of changes to the cice_in.nml file:

Must add this to `&setup_nml`
```
input_dir      = 'INPUT/'
```
and change all instances of paths with `RESTART` at the beginning, to `INPUT`
```
pointer_file   = 'INPUT/ice.restart_file'
grid_file    = 'INPUT/grid.nc'
kmt_file     = 'INPUT/kmt.nc'
```

### DIFF
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -62,9 +62,9 @@ class Cice(Model):
         setup_nml = self.ice_in['setup_nml']
 
         res_path = os.path.normpath(setup_nml['restart_dir'])
-        input_dir = setup_nml.get('input_dir',None)
+        input_dir = setup_nml.get('input_dir', None)
 
-        if input_dir is None: 
+        if input_dir is None:
             # Default to reading and writing inputs/restarts in-place
             input_path = res_path
         else:
@@ -85,21 +85,25 @@ class Cice(Model):
         if not path == input_path:
             print('Grid file path in {nmlfile} ({gridpath}) does not match '
                   'input path ({inputpath})'.format(
-                    nmlfile=self.ice_nml_fname, 
-                    gridpath=path, 
-                    inputpath=input_path
-            ))
+                    nmlfile=self.ice_nml_fname,
+                    gridpath=path,
+                    inputpath=input_path))
             sys.exit(1)
 
         # Check paths are consistent in namelist
-        for var, nml in (('kmt_file', 'grid_nml'), ('pointer_file', 'setup_nml')):
+        for var, nml in (('kmt_file', 'grid_nml'),
+                         ('pointer_file', 'setup_nml')):
             tmp_path, _ = os.path.split(self.ice_in[nml].get(var))
             tmp_path = os.path.normpath(tmp_path)
             if not path == tmp_path:
-                print('Paths for {var} ({varpath}) and grid_file ({gridpath}) in {nmlfile} do not '
-                      'match'.format(var=var, varpath=tmp_path, gridpath=path, nmlfile=self.ice_nml_fname))
+                print('Paths for {var} ({varpath}) and grid_file '
+                      '({gridpath}) in {nmlfile} do not '
+                      'match'.format(var=var,
+                                     varpath=tmp_path,
+                                     gridpath=path,
+                                     nmlfile=self.ice_nml_fname))
                 sys.exit(1)
-            
+
         if not os.path.isabs(input_path):
             input_path = os.path.join(self.work_path, input_path)
         self.work_input_path = input_path
@@ -114,7 +118,6 @@ class Cice(Model):
         if not os.path.isabs(work_out_path):
             work_out_path = os.path.join(self.work_path, work_out_path)
         self.work_output_path = work_out_path
-
 
     def set_model_output_paths(self):
         super(Cice, self).set_model_output_paths()
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -141,7 +141,6 @@ class Cice(Model):
         return [sorted(restart_files)[-1]]
 
     def get_ptr_restart_dir(self):
-        # return self.ice_in['setup_nml']['restart_dir']
         return os.path.relpath(self.work_input_path, self.work_path)
 
     def get_access_ptr_restart_dir(self):
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -62,26 +62,41 @@ class Cice(Model):
         setup_nml = self.ice_in['setup_nml']
 
         res_path = os.path.normpath(setup_nml['restart_dir'])
+        input_path = setup_nml.get('input_dir',None)
+
+        if input_path is None: 
+            # Default to reading and writing inputs/restarts in-place
+            input_path = res_path
+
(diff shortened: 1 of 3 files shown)


aidanheerdegen:
In fact that is all that needs to be done to use this feature, is alter the paths in `cice_in.nml` and add an `input_path`.

aidanheerdegen:
Don't merge this quite yet, I'd like to add some checks for the input path to make sure the paths to `grid.nc` and `kmt.nc` are consistent. This is an example where it is easy to have contradictory information.

marshallward:
Sure no problem.

Probably best to keep these on separate branches until they're ready to be merged.

aidanheerdegen:
I have added some tests for path consistency in the `cice` driver. One existed already, but didn't give a useful error message.

These changes are backwards compatible. The split input and restart stuff only kicks in if `input_dir` is specified in `cice_in.nml` as well as `restart_dir`. If there is only `restart_dir` it behaves as it did before.

Because we're now using symlinks in `INPUT`, this solves issues like https://github.com/marshallward/payu/issues/161

payu/models/cice.py
```diff
@@ -62,27 +62,63 @@ def set_model_pathnames(self):
...
+        if input_dir is None:
+            # Default to reading and writing inputs/restarts in-place
+            input_path = res_path
+        else:
+            input_path = os.path.normpath(input_dir)
+
+        # Determine if there is a work input path from grid path
+        grid_nml = self.ice_in['grid_nml']
+        path, _ = os.path.split(grid_nml['grid_file'])
+        if path and not path == '.':
```
marshallward: replace `'.'` with `os.path.curdir` ?

payu/models/cice.py
```diff
@@ -62,27 +62,63 @@ def set_model_pathnames(self):
...
+            assert not os.path.isabs(path)
+            path = os.path.normpath(path)
+            # Get input_dir from grid_file path unless otherwise specified
+            if input_dir is None:
+                input_path = path
+
+        # Check for consistency in input paths due to cice having the same
+        # information in multiple locations
+        if not path == input_path:
+            print('Grid file path in {nmlfile} ({gridpath}) does not match '
```
marshallward: maybe prepend with `payu: error: `

payu/models/cice.py
```diff
@@ -105,7 +141,8 @@ def get_prior_restart_files(self):
         return [sorted(restart_files)[-1]]
 
     def get_ptr_restart_dir(self):
-        return self.ice_in['setup_nml']['restart_dir']
+        # return self.ice_in['setup_nml']['restart_dir']
+        return os.path.relpath(self.work_input_path, self.work_path)
```
marshallward: Should we use absolute path here?  Any advantage/disadvantage?

I've generally tried to use abspaths in the path, though not for any explicit reason.

payu/models/cice5.py
```diff
@@ -61,24 +61,27 @@ def setup(self):
     def set_model_pathnames(self):
         super(Cice5, self).set_model_pathnames()
 
-        # Change the INPUT path, we want everything to go into RESTART
-        self.work_input_path = self.work_restart_path
+        self.split_paths = (not self.work_input_path == self.work_restart_path)
```
marshallward: use `!=` ?  Not sure if (`not A == B`) is correct here?

payu/models/cice.py
```diff
@@ -62,27 +62,63 @@ def set_model_pathnames(self):
         setup_nml = self.ice_in['setup_nml']
 
         res_path = os.path.normpath(setup_nml['restart_dir'])
+        input_dir = setup_nml.get('input_dir', None)
+
+        if input_dir is None:
+            # Default to reading and writing inputs/restarts in-place
+            input_path = res_path
+        else:
+            input_path = os.path.normpath(input_dir)
```
marshallward: Should this be absolute path?  (Genuinely not sure BTW)

payu/models/cice.py
marshallward: `os.path.curdir` is just the platform-agnostic name for `.`, btw.  It doesn't literally mean the current directory.

Yes, I don't think I've used it in the past.

aidanheerdegen: Oh right, didn't know that. Ok, I'll add it in.

payu/models/cice.py
```diff
@@ -62,27 +62,63 @@ def set_model_pathnames(self):
...
+
+        # Determine if there is a work input path from grid path
+        grid_nml = self.ice_in['grid_nml']
+        path, _ = os.path.split(grid_nml['grid_file'])
+        if path and not path == '.':
+            assert not os.path.isabs(path)
+            path = os.path.normpath(path)
+            # Get input_dir from grid_file path unless otherwise specified
+            if input_dir is None:
+                input_path = path
```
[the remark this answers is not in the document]
aidanheerdegen: It is done like that to support the previous behaviour. Would have to check the path for `grid_file` to check it against the others paths regardless of the state of `input_dir`. Does that make sense?

payu/models/cice.py
aidanheerdegen: I have always wondered if we shouldn't be using some sort of logger ... but point taken. Do we have  standard way of doing this that I can refer to?

marshallward: No, there is no standard way, and yes, I wish there were.

The `payu: warning:` and `payu: error:` stuff has been slowly emergent but is by no means consistent.

payu/models/cice.py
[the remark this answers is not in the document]
marshallward: Yeah, makes sense, I've had a few cases like that in other places.

payu/models/cice.py
aidanheerdegen: It is used to print the path to the restart file in the `ice.restart_file` pointer file. I don't really know how CICE groks this, though I guess a full path would work. As it is for a pretty specific purpose I think I'd rather leave that for now.

payu/models/cice.py
aidanheerdegen: It is kept a relative path until the end (line 108) where it is joined to `work_path`. It is done like that so it can be compared with the other paths on the grid, kmt and restart pointer files.

payu/models/cice.py
aidanheerdegen: Okey doke

aidanheerdegen:
Made changes as per review comments.

aidanheerdegen:
I should also say, I have tested these changes with an access-om2 1 degree config. Both with split paths (`input_dir` and `restart_dir`), a single input path (`restart_dir`) and also split paths, but setting both to `RESTART`. All work without error.

PR #165, opened 2019-02-12
Per-file inputs

marshallward:
Minor change to input symlinks for individual files, rather than just
the contents of directories.

Also added the `mask` config variable to the model config fields.  This
is just a nasty hack to add this config to the MOM models.  This really
needs a very serious rethink.

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -122,6 +122,9 @@ class Experiment(object):
         model_fields = ['model', 'exe', 'input', 'ncpus', 'npernode', 'build',
                         'mpthreads', 'exe_prefix']
 
+        # XXX: Temporarily adding this to model config...
+        model_fields += ['mask']
+
         # TODO: Rename this to self.submodels
         self.models = []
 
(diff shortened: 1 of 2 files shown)


marshallward:
@aidanheerdegen can you see if this works for you, esp. with respect to manifests?

aidanheerdegen:
Yeah seems ok. You want me to rebase on to this to check it?

aidanheerdegen:
That code block now looks like this:
```python
            # Add files to manifest
            for input_path in self.input_paths:
                for path, dirs, files in os.walk(input_path):
                    workrelpath = os.path.relpath(path, input_path)
                    subdir = os.path.normpath(
                                os.path.join(self.work_input_path_local,
                                workrelpath))
                    if not os.path.exists(subdir):
                        os.mkdir(subdir)
                    for f_name in files:
                        f_orig = os.path.join(path, f_name)
                        f_link = os.path.join(
                            self.work_input_path_local, workrelpath, f_name)
                        # Do not use input file if it is in RESTART
                        if not os.path.exists(f_link):
                            self.expt.manifest.add_filepath(
                                'input',
                                f_link,
                                f_orig,
                                self.copy_inputs
                            )
```
So naively when it tries to `os.walk` a file, nothing will happen. It will be ignored.

marshallward:
sounds like it will be disruptive to merge this with manifest changes going on, closing.

PR #170, opened 2019-02-25
Introduced bug in check_fast when fixing the previous fullpath bug.

aidanheerdegen:
Used all paths instead of just the paths of the changed files. Fixed


aidanheerdegen:
@marshallward I will merge this, as it is small and a clear bug

PR #172, opened 2019-02-27
Altered UM restart calendar handling

aidanheerdegen:
Added explicit `um_calendar.res.yaml` file from which to restart the UM, rather than reading in prior version of `namelists`. This breaks a dependency on a prior output directory.

Fixes https://github.com/marshallward/payu/issues/168

`namcouple` is not required for model time restarts, so is not copied to restarts, but instead is archived in output.


aidanheerdegen:
I am just checking I haven't stuffed up `accessom2` with these changes.

PR #173, opened 2019-03-01
Fileinputs

marshallward:


marshallward:
Working on a method using `os.walk` over files, at the insistence of Dr Heerdegen

aidanheerdegen:
I should have left a nasty message in the PR totally slamming this code

aidanheerdegen (approved):
Nice.

aidanheerdegen:
Just for future reference (when python2 support is dropped)
```python
def walkf(path):
    """
    Add to os.walk to return a file name if that is passed in
    """
    if os.path.isfile(path):
        path, fname = os.path.split(path)
        yield path, None, fname
    else:
        yield from os.walk(path)
```

PR #174, opened 2019-03-04
Updated MITgcm calendar handling

aidanheerdegen:
Can change timestep within a run and maintain correct model time and step number.

Uses a yaml restart file to save the model end time for the current
run which is read in by the next run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,10 @@ language: python
 python:
     - "2.7"
     - "3.5"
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
Addresses https://github.com/marshallward/payu/issues/169

aidanheerdegen:
I will get Nick and Callum to test it out before we merge it in, to make sure I have the timing logic correct.

PR #177, opened 2019-04-02
Avoid mutable default parameters

kinow:
Hi!

Small change to prevent unwanted behaviour for `add_fast`. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Cheers

### DIFF
--- a/payu/manifest.py
+++ b/payu/manifest.py
@@ -159,12 +159,14 @@ class PayuManifest(YaManifest):
 
         return True
 
-    def add_fast(self, filepath, hashfn=fast_hashes, force=False):
+    def add_fast(self, filepath, hashfn=None, force=False):
         """
         Bespoke function to add filepaths but set shortcircuit to True, which
         means only the first calculatable hash will be stored. In this way only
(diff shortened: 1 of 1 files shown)


marshallward:
Thanks for finding this.  Also it seems that the Travis tests (as minimal as they are) were turned off after moving to payu-org.  Will keep this one open until we get those reactivated.

aidanheerdegen:
Thanks @kinow. I agree your change is a good one, but for anyone following along at home I don't believe the existing code leads to any unintended behaviour, as the way this routine is currently used is quite limited.

Are you evaluating `payu` for running models, or just taking an interest?

kinow:
The latter. Agree it is harmless, and all good if not merged :-) it was for the just in case

marshallward:
I agree it'sa good idea to avoid mutables as inputs, even if there's no real danger in this case.

@aidanheerdegen if you're happy to merge then no objections from me.

aidanheerdegen:
Didn't merge until replies were in, seemed rude otherwise .. :)

PR #178, opened 2019-04-02
Fix string format in model.py

kinow:
Hi!

Running into that `Exception` would raise a `KeyError` 👍


marshallward:
Is this an issue?  AFAIK most (all?) exceptions have `__str__()` defined, including `FileNotFoundError`.

marshallward:
ah sorry nevermind, I see.  Thanks for catching it :)

PR #179, opened 2019-04-02
Fix typos

kinow:
Hi! Few minor typos. I think the only more important is the main comment in `payu/__init__.py` which may introduce Payu to users 👍

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -147,7 +147,7 @@ configuration.
 
 ``exe``
    Binary executable for the model. This can either be a filename in the
-   laboratorys ``bin`` directory, or an absolute filepath. Various model
+   laboratories ``bin`` directory, or an absolute filepath. Various model
    drivers typically define their own default executable names.
 
 ``submodels``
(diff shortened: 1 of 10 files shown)


docs/source/config.rst
```diff
@@ -147,7 +147,7 @@ configuration.
 
 ``exe``
    Binary executable for the model. This can either be a filename in the
-   laboratorys ``bin`` directory, or an absolute filepath. Various model
+   laboratories ``bin`` directory, or an absolute filepath. Various model
```
marshallward: laboratory's

payu/manifest.py
```diff
@@ -162,7 +162,7 @@ def add_filepath(self, filepath, fullpath, copy=False):
     def add_fast(self, filepath, hashfn=fast_hashes, force=False):
         """
         Bespoke function to add filepaths but set shortcircuit to True, which
-        means only the first calculatable hash will be stored. In this way only
+        means only the first calculable hash will be stored. In this way only
```
marshallward: (seems both are valid but smaller is better :D )

marshallward:
Thanks, especially for hunting the repeated typos down :)

payu/manifest.py
kinow: Oh! Today-I-Learned! Thanks @marshallward !

kinow:
Thank you too @marshallward !

PR #180, opened 2019-04-02
Remove duplicate key

kinow:
`man_pages` is being declared twice, so one of the values is not used. I left the last one, as that's probably what's being used now.


marshallward:
Thanks, still a great deal of generic content in that file.

PR #181, opened 2019-04-02
Remove unused imports

kinow:
Hi,

A few unused imports. Hopefully Travis-CI should be able to confirm they are not used.

### DIFF
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -10,9 +10,6 @@
 # Standard library
 import errno
 import os
-import shlex
-import subprocess
-import sys
 
 # Extensions
 import yaml
--- a/payu/manifest.py
+++ b/payu/manifest.py
@@ -9,12 +9,7 @@
 """
 from __future__ import print_function, absolute_import
 
-# Local
-from payu import envmod
-from payu.fsops import make_symlink
-
 # External
-from copy import deepcopy
 import fnmatch
 import os
 import sys
@@ -24,7 +19,6 @@ from distutils.dir_util import mkpath
 
 from yamanifest.manifest import Manifest as YaManifest
 
-from payu import envmod
 from payu.fsops import make_symlink
 
 
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -14,11 +14,8 @@ from __future__ import print_function
 # Standard Library
 import errno
 import os
-import stat
 import sys
-import shlex
 import shutil
-import subprocess as sp
 import datetime
 
 # Extensions
--- a/payu/models/cice5.py
+++ b/payu/models/cice5.py
@@ -11,7 +11,6 @@ http://www.apache.org/licenses/LICENSE-2.0
 from __future__ import print_function
 
 import os
-import shutil
 
 from payu.models.cice import Cice
 from payu.fsops import mkdir_p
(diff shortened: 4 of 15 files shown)


marshallward:
Thanks for finding this. Also it seems that the Travis tests (as minimal as they are) were turned off after moving to payu-org. Will keep this one open until we get those reactivated.

aidanheerdegen:
Cripes that is a lot of cruft. Did you run a linter across the code to pick this up @kinow? If so, we should probably auto-lint with travis.

marshallward:
Just a warning that the current number of pylint warnings is very large :).

kinow:
@aidanheerdegen found some typos, then ran pycharm's inspection. Which uses several linters (some pycharm custom ones too).

Then read the docs, read and executed unit tests, and finally created the class diagram to see what was connected to what.

I liked the design of Payu, and as a thank you for the docs and open sourcing, sent what appeared in the ide.

marshallward:
Definitely appreciate the input, thanks!

From the cylc activity, I'm guessing you are at NIWA or at least connected to similar activities, would be great to catch up if the opportunity arises.

kinow:
>From the cylc activity, I'm guessing you are at NIWA or at least connected to similar activities, would be great to catch up if the opportunity arises.

That's correct, I'm based in Auckland, working on Cylc for NIWA. Found Payu while searching cylc related things in GitHub, and found an issue related to Jinja2 in Payu I think. One thing leads to another, and then learned a bit about Payu 😁

aidanheerdegen:
One of the "must get around to doing this one day" was encapsulating the PBS stuff in a `Scheduler` class so we can support other schedulers, like `slurm`, but also maybe `cylc`. 

Does that sound interesting to you @kinow? 

Is there some prior art we could ~nick~ repurpose for this? I noticed the PBSpro sources have a large python test harness but I took one look and ran away (14K lines)

https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/lib/pbs_testlib.py

kinow:
>One of the "must get around to doing this one day" was encapsulating the PBS stuff in a Scheduler class so we can support other schedulers, like slurm, but also maybe cylc.
>
>Does that sound interesting to you @kinow?

Oh, it does sound interesting!

>Is there some prior art we could nick repurpose for this? I noticed the PBSpro sources have a large python test harness but I took one look and ran away (14K lines)

I think we have something similar implemented in Cylc, but in the next releases it may change to support Python setuptools' entry points, and maybe some object hierarchy. At the moment the logic in Cylc is in [`BatchSysManager`](https://github.com/cylc/cylc/blob/6216eef87d207e9b4a990f1402b7c6a451e86a41/lib/cylc/batch_sys_manager.py) and at the implementations in [`cylc.batch_sys_handlers`](https://github.com/cylc/cylc/tree/master/lib/cylc/batch_sys_handlers).

We also have a ticket for [DRMAA](https://github.com/cylc/cylc/issues/2205). Perhaps if a `Scheduler` class was generic enough, we could use that in Payu and also in Cylc, as Cylc also interfaces with these schedulers.

WDYIT @aidanheerdegen ?

aidanheerdegen:
Sorry .. the day caught up with me. Looks very interesting, and as always the devil is in the detail. I will try and remember to make an issue to continue the discussion.

marshallward:
@aidanheerdegen Just noticed this is still here, looks pretty innocent.  Easy enough to merge?

Issue #188, opened 2019-06-04
Old restarts retained in restart manifest when using runspersub


aidanheerdegen:
@navidcy had an issue with previous restarts remaining in his restart manifest when using a large number of runspersub (150). 

In this case it is a `mom6` model with
```fortran
RESTART_CONTROL = 3  
```
set in `MOM_input` and the restart interval set so that every run produced a `MOM.res.nc` file as well as one that looks like `MOM.res_Y0002_D066_S00000.nc`. In this case they have identical information.

This bug is not noticed when the restart files are identical as the entries just get overwritten. But with the above settings a unique restart file was created every time, and the restart manifest was not properly reset between subruns.

PR #189, opened 2019-06-04
Previous restarts polluting restart manifests with subruns

aidanheerdegen:
Initialise manifest in run_cmd for subruns, otherwise previous restart can remain in the manifest.

Addresses https://github.com/payu-org/payu/issues/188


aidanheerdegen:
I've tested this and it fixes the bug, so I'll merge it

PR #195, opened 2019-09-18
Make get_qstat_info more robust

aidanheerdegen:
Added retry decorator (from tenacity) to get_qstat_info as calls to qstat were timing out and bringing down simulations with multiple runs per submit.

### DIFF
--- a/payu/scheduler/pbs.py
+++ b/payu/scheduler/pbs.py
@@ -14,6 +14,7 @@ import sys
 import shlex
 import subprocess
 
+import tenacity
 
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
@marshallward I did wonder if it was a good idea to add yet another dependency, but decided that maybe this is something that could be useful more than just in this one place, so decided to go with the tried and tested solution.

aidanheerdegen:
Checked and seems to be working on raijin. Will merge as I need a fix for a user.

PR #199, opened 2019-09-18
Added payu list call to conda test to ensure code imports cleanly

aidanheerdegen:
Will also redo incorrect tagging if this passes.


aidanheerdegen:
This seems to work ok, so will merge

Issue #203, opened 2019-10-08
Fine grained control over reproduce


aidanheerdegen:
It would be good to have finer-grained control of what must be reproduced. The current `--reproduce` flag insists `input` and `restart` are reproduced, and optionally `exe` reproduce can be turned off, for the use case where the executable has changed for a good reason, or just due to time and code updates.

There could be sub-options to `manifest` to enforce reproduction (checksums matching) for each manifest separately.

The use case is for the COSIMA `access-om2` model configs, where a configuration is updated and published, and then users clone and use that configuration. It would be advantageous to specify something like:

```
manifest:
    exe: True
   input: True
```

to make sure the executable and input files are checked and match those from the config. In this case it is not necessarily desirable to enforce restart reproduction.

PR #204, opened 2019-10-11
Fine-grained reproduce flags and testing

aidanheerdegen:
Added test model for ... testing. Has minimal requirements and means
tests can be independent of model driver implementation. A platonic
ideal if you will.
    
Added config options for fine-grained reproduce options. Input, exe
and restart reproduce can be turned on independently of each other.
    
Added extensive testing of init/setup and reproduce.

Fixes https://github.com/payu-org/payu/issues/203

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,9 +13,9 @@ jobs:
             - run: |
                 wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O conda.sh
                 bash conda.sh -b -p ~/conda
+                ~/conda/bin/conda update conda
                 ~/conda/bin/conda config --system --add channels conda-forge
                 ~/conda/bin/conda config --system --add channels coecms
-                ~/conda/bin/conda update conda
                 ~/conda/bin/conda install --yes conda-build conda-verify
 
             - run: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
             - run: |
                 wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O conda.sh
                 bash conda.sh -b -p ~/conda
-                ~/conda/bin/conda update -all
+                ~/conda/bin/conda update --all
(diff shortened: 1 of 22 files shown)


aidanheerdegen:
@marshallward I've moved the pbs environment init function to `cli.py`so that it is only called if a job is submitted. In this way PBS isn't required on travis unless a PBS job is submitted.

(The PBS install stuff seems to be broken BTW. It couldn't find `/etc/pbs.conf` as the install script errors)

marshallward:
I think a lot of that was cut-and-paste from the bundled Python bootstrap script.

Is there not some way to retain 2.7 and 3.5 testing?  I don't know if removing versions is the right path foward here.

marshallward:
Are you dropping 2.7 to use `exist_ok` in `os.makedirs`?  Can not some of this be resolved using `mkdir_p` in the `fs` module?

aidanheerdegen:
Sorry, I thought we'd decided to drop 2.7 support, as people could always use an older version?

All the tests I just wrote are not python2 compatible as they use `pathlib` to make writing the tests less painful. I also dropped `3.5` as I believe there are a few `pathlib` compatibilities that only work from `3.6`, but it isn't a deal breaker. I did mean to add `3.7` so it didn't look quite so bad!

marshallward:
Ah yeah, ok, it's just the test itself.  `pathlib` does seem pretty fantastic, as does the `exist_ok` argument in `os.makedirs`.

All good, just wondering how things might work if I ever try porting payu to Gaea.

aidanheerdegen:
Could have version dependent tests where the 2.7 test just does `payu list` like the `conda` test

https://github.com/payu-org/payu/blob/master/conda/meta.yaml#L34

to make sure there are no breaking changes to `payu` itself.

Do you only have access to `python2.7` on gaea? Eeeuuuwww.

aidanheerdegen:
So now the tests will do `payu list` and `pylint -E` for all python versions. The latter should pick up anything not python27 compliant.

Additionally for python3.x interpreters `pytest test/test_manifest.py` is run. This is just testing `init`, `setup` and `sweep` for a test model.

Issue #209, opened 2019-11-18
Work with Environment Modules 4 on Gadi


penguian:
Today I tried running on Gadi. Building succeeded. Unfortunately, I was unable to run the 1deg_jra55_ryf  experiment. The error in [payu/envmod.py](https://github.com/payu-org/payu/blob/master/payu/envmod.py) line 40 is
```
FileNotFoundError: [Errno 2] No such file or directory: '/opt/Modules/v4.3.0/init/.modulespath’
```
This error is essentially because payu is written for Environment Modules version 3.2.6 on Rajin, and the version of Environment Modules on Gadi is 4.3.0. This version of Environment Modules is backwards incompatible with version 3.2.6. See https://modules.readthedocs.io/en/latest/diff_v3_v4.html
Therefore payu needs to be changed to be compatible with Environment Modules version 4, in particular the current configuration on Gadi. It may also be possible that the configuration of Environment Modules on Gadi could change. 

See also pull request #128 and issue #200.

aidanheerdegen:
The proximate cause of this issue is that the module environment variables are not consistent with how they're configured on `raijin`.

On raijin the following environment variables are set:
```
MODULE_VERSION=3.2.6
MODULESHOME=/opt/Modules/3.2.6
```
on `gadi` the only equivalent is
```
MODULES_CMD=/opt/Modules/v4.3.0/libexec/modulecmd.tcl
```

Without a code change the above issue is solved by setting `MODULE_VERSION`:
```
export set MODULE_VERSION=v4.3.0
```
I think this should be the first fix, but a code fix to extract `MODULE_VERSION` from `MODULES_CMD`, or just use `MODULES_CMD` directly.

aidanheerdegen:
Once the above issue was solved another error occurred:
```
Traceback (most recent call last):
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/bin/payu-run", line 10, in <module>
    sys.exit(runscript())
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/payu/subcommands/run_cmd.py", line 128, in runscript
    expt.run()
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/payu/experiment.py", line 505, in run
    'libmpi.so'
[...]

PR #210, opened 2019-11-19
Check for MODULESHOME environment variable for gadi support

aidanheerdegen:
Fixes https://github.com/payu-org/payu/issues/209

On `gadi` `MODULE_VERSION` environment variable is no longer set by default, but `MODULESHOME` is. This change checks for `MODULESHOME` before trying to generate it from `MODULE_VERSION`.

### DIFF
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -22,7 +22,11 @@ def setup(basepath=DEFAULT_BASEPATH):
     """Set the environment modules used by the Environment Module system."""
 
     module_version = os.environ.get('MODULE_VERSION', DEFAULT_VERSION)
-    moduleshome = os.path.join(basepath, module_version)
+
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
As`MODULE_VERSION` is set on `raijin` it is safe to change the default to match `gadi` in any case.

I have tested this works on `raijin`

aidanheerdegen:
If there are no objections I'll merge, tag a new version and push it to `conda/analysis3-unstable`

aidanheerdegen:
As`MODULE_VERSION` is set on `raijin` it is safe to change the default to match `gadi` in any case.

I have tested this works on `raijin`

aidanheerdegen:
I'll just merge I think, as there could be others testing `gadi` who will be tripped up by this.

Issue #217, opened 2020-01-22
default nodesize


navidcy:
On raijin the default node size was, I believe, 16 and when we wanted to use `normalbw` queue we had to add in `config.yaml`

```bash
platform:
    nodesize: 32
```

Now with gadi, should we change the default nodesize to 48?

marshallward:
Maybe a better idea is to use `pbsnodes` to parse and get this number?

aidanheerdegen:
Whilst that information is available using `pbsnodes`:
```json
$ pbsnodes -F json gadi-cpu-clx-0470
{
    "timestamp":1579661565,
    "pbs_version":"19.2.4.20190830141245",
    "pbs_server":"gadi-pbs-01",
    "nodes":{
        "gadi-cpu-clx-0470":{
            "Mom":"gadi-cpu-clx-0470.gadi.nci.org.au",
            "ntype":"PBS",
            "state":"job-busy",
            "pcpus":96,
            "jobs":[
                "1079829.gadi-pbs"
            ],
            "resources_available":{
                "arch":"linux",
                "host":"gadi-cpu-clx-0470",
                "jobfs":"429496729600b",
                "mem":"213647360kb",
                "ncpus":48,
                "ngpus":0,
                "topology":"rack-23-ib2,rack-23,group-1,cpu-clx",
                "vmem":"213647360kb",
                "vnode":"gadi-cpu-clx-0470"
            },
            "resources_assigned":{
                "jobfs":"102400kb",
                "mem":"41943040kb",
                "ncpus":48
            },
            "comment":"offlined by hook 'begin_checknode' due to hook error",
            "resv_enable":"True",
            "sharing":"default_shared",
            "license":"l",
            "last_state_change_time":1579659651,
            "last_used_time":1579659544
        }
    }
}
```
I've had a look but it isn't clear to me how we would know which node to query for that information a priori, i.e. a mapping of queue name to nodes.

aidanheerdegen:
I'm wondering about a system config file with some of this information, something like:
```yaml
gadi.nci.org.au:
    normal:
          ncpus: 48
          mem: 256GB
    normalbw:
          ncpus: 28
          mem: 128GB
```
that could live somewhere in the {{payu}} directory. Maybe a {{platform}} directory?

navidcy:
And when the user selects a queue the rest is determined automatically by payu?
This sounds nice.

Include also `express` and `expressbw` in that case.

[...]

PR #223, opened 2020-02-06
Fix bug: messing with LD_LIBRARY_PATH when in conda environment

### LINKED ISSUE #221: Populating LD_LIBRARY_PATH can cause library conflicts when used in conda
NCI users on `gadi` experienced a sudden unexplained error with models that had previously been working
```
yatm_1bb8904.exe: symbol lookup error: yatm_1bb8904.exe: undefined symbol: netcdf_mp_nf90_open_
```

The issue occurred with `payu` in the `conda/analysis3-20.01` environment but was not present when `conda/analysis3-19.10` was used. 

This is because another package (`netcdf-fortran`) was upgraded in the `20.01` environment. The dependency changed from `libnetcdff.so.6` to `libnetcdff.so.7`, which is the same library the `yam` exe used.

This occurs because `LD_LIBRARY_PATH` is populated with the python library directory:
https://github.com/payu-org/payu/blame/73ba06b47ed25b2801ebcd2eb9d00c3ebc79b72b/payu/cli.py#L87-L88

so libraries in that directory are picked up preferentially.

When using python in a `conda` environment, the `python` executable has `RPATH` set, so it is not necessary to add to `LD_LIBRARY_PATH`, and within a `conda` environment can lead to bugs like this.

Proposed solution is to detect if the `python` interpreter is in a `conda` environment and branch past this code if so.

aidanheerdegen:
Added is_conda function to fsops and used to detect conda environment and not muck about with LD_LIBRARY_PATH

Fixes #221


aidanheerdegen:
Changed default node size which is a temporary fix for #217

PR #251, opened 2020-05-01
Removed scipy from dependencies

marshallward:
Scipy is listed as a dependency but does not appear to be used in payu.

Scipy depends on NumPy, which is a very large package, so we probably do
not want this to be installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,6 @@ setup(
         'yamanifest',
         'dateutil',
         'tenacity',
-        'scipy',
     ],
     install_requires=[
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
Well spotted.

You can just merge stuff like this if you want @marshallward

marshallward:
Bob and Alistair have a interesting policy where they are not permitted to merge their own code, and the other must merge it.

I don't think I'm back up to speed yet to review yours and hold up essential NCI business, but I like the idea of someone else merging my code at the moment :).

PR #252, opened 2020-05-01
Empty YAML config handled as empty dict

marshallward:
If one passes an empty YAML file, e.g. all lines commented out, then
PyYAML returns `None`, which breaks subsequent assumptions in the
handling of config.

We now parse an empty file as an empty dict, {}.

### DIFF
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -73,8 +73,6 @@ def read_config(config_fname=None):
         else:
             raise
 
-
-
     collate_config = config.pop('collate', {})
 
     # Transform legacy collate config options
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -61,6 +61,10 @@ def read_config(config_fname=None):
     try:
         with open(config_fname, 'r') as config_file:
             config = yaml.safe_load(config_file)
+
+        # NOTE: A YAML file with no content returns `None`
+        if config is None:
+            config = {}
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
Do we want to add a test to make sure this is picked up in case it gets accidentally removed/changed?

marshallward:
Do you mean just adding a test case with an empty YAML file?  Yeah, I can add that.

But you may mean something more significant, which I am missing.

aidanheerdegen:
Yeah I meant an empty yaml test. I don't think it is super important, but it is a good habit to get into to add to the tests when a bug is fixed.

aidanheerdegen:
I noticed this was still open, do you still feel like adding an empty yaml file test case @marshallward?

marshallward:
Seems I cannot even remember how to run the tests (or perhaps they have changed).  This may need a chat.

aidanheerdegen:
@marshallward I added a test for an empty config, and found to my surprise it isn't empty because of that auto translation of the old `collate` boolean to a `dict`. 

Are you happy with this behaviour, or do you want some special case logic around this

https://github.com/payu-org/payu/blob/master/payu/fsops.py#L72-L91

aidanheerdegen:
Sorry, had to rebase and force push for tests to be consistent with current code base.

marshallward:
All good with me!

PR #256, opened 2020-05-06
Added force option to setup and run

### LINKED ISSUE #255: Add option to overwrite existing directories
Users would like to be able to `payu run` with an existing `work` directory and for `payu` to run as usual, removing the `work` directory automatically.

There are some issues with implementing this interactively (#250) so a half way solution is to provide a command line argument to remove any existing `work` directory. The option `-f/--force` is probably appropriate, unless this is to broad and might want to be used in some other context?

If there is an option to overwrite an existing `work` directory should there also be an option to overwrite an existing `outputXXX` directory?

If yes then should they be separate options? Say `-w` for overwrite `work` and `-o` for overwrite `outputXXX`?

aidanheerdegen:
Added force option to setup and run. Automatically removes existing work directory. Closes #255.

Also hanged default of reproduce option to False from None.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -30,6 +30,20 @@ DEFAULT_CONFIG = 'config.yaml'
 def parse():
     """Parse the command line inputs and execute the subcommand."""
 
+    parser = generate_parser()
+
+    # Display help if no arguments are provided
+    if len(sys.argv) == 1:
+        parser.print_help()
+    else:
+        args = vars(parser.parse_args())
+        run_cmd = args.pop('run_cmd')
+        run_cmd(**args)
+
+
+def generate_parser():
+    """Parse the command line inputs generate and return parser."""
+
     # Build the list of subcommand modules
     modnames = [mod for (_, mod, _)
                 in pkgutil.iter_modules(payu.subcommands.__path__,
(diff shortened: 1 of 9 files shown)


aidanheerdegen:
This isn't finished yet. Have to make a test. Also haven't made output overwrite. @marshallward do we want to add that too? I'm broadly in favour, just wanted to check.

marshallward:
I don't know how output override would work, since if a run crashed then there ought not be an output directory.  (If there were then I'd consider it a bug).

But if there's some problem I can't see that you think it will solve, then I reckon it's alright.

aidanheerdegen:
Not really. The instructions say you can `payu run -i n` but don't mention you would have to remove the `output00n` directory to do this. Same with the `reproduce` flag. You can use it to re-run an experiment, but it won't work if there is an existing `output` directory.

aidanheerdegen:
So it is solving a different problem.

marshallward:
OK, then I could see how it could be useful.

I had never used the `payu run -i n` stuff much, but there ought to be better support for navigating the history, and wiping and re-running old jobs would certainly be a part of that.

aidanheerdegen:
Maybe do that in a different PR I guess, rather than overload this one right now?

marshallward:
Sounds good, smaller is better!

aidanheerdegen:
Still have to add a test, which I'll do tomorrow as I am off for now. Thanks for the feedback @marshallward

aidanheerdegen:
Well I went a bit mad and added a bunch of command line argument tests, which did show up a weird errant option that was not a bug but was not correct.

Pretty happy with the resulting tests. Not testing all commands, but easy enough to add them in if you feel so inclined.

aidanheerdegen:
I reckon I'll just go ahead and merge this @marshallward

PR #260, opened 2020-05-13
Changed to floating point division in MITgcm to support fractional timestep

### LINKED ISSUE #228: Cannot change timestep to fraction of a second in MITgcm
Report from user:

> Another MITgcm issue with changing timestep. It seems that you can't change a timestep to a fraction of a second e.g. here I tried to change the timestep from 0.1 to 0.2, and I get the error
> 
> 
> (base) [cjs157@raijin4 SOMs_hr1_200m_nc2]$ payu setup
> Use of these keys is deprecated: collate_walltime, collate_mem.
> Instead use collate dictionary and subkey without 'collate_' prefix
> laboratory path:  /short/v45/cjs157/mitgcm
> binary path:  /short/v45/cjs157/mitgcm/bin
> input path:  /short/v45/cjs157/mitgcm/input
> work path:  /short/v45/cjs157/mitgcm/work
> archive path:  /short/v45/cjs157/mitgcm/archive
> Use of these keys is deprecated: collate_walltime, collate_mem.
> Instead use collate dictionary and subkey without 'collate_' prefix
> Loading input manifest: manifests/input.yaml
> /g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/yamanifest/manifest.py:99: YAMLLoadWarning: calling yaml.load_all() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   self.header, self.data = yaml.load_all(file)
> Setting up mitgcm
> Calculated n_timesteps from starttime and endtime
> /g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/payu/models/mitgcm.py:133: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   restart_info = yaml.load(restart_file)
> payu : error: Timestep changed to 0.2. New timestep not integer multiple of start time: 3240.0
> 
> 
> 
> Clearly the error is false through - since 3240/0.2 is an integer... is this a a rounding issue in the payu logic? (Note that payu accepts a change to a timestep of 1.0 secs).

aidanheerdegen:
Changed to floating point division in MITgcm to avoid issues with floating point round off with fractional timesteps.

Added mitgcm model tests.

Fixes #228

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_script:
 script:
     - payu list
     - pylint --extension-pkg-whitelist=netCDF4 -E payu
-    - if [[ $TRAVIS_PYTHON_VERSION == 3.6 || $TRAVIS_PYTHON_VERSION == 3.7 ]]; then PYTHONPATH=. coverage run --source payu -m py.test -s test/*.py; fi;
+    - if [[ $TRAVIS_PYTHON_VERSION == 3.6 || $TRAVIS_PYTHON_VERSION == 3.7 ]]; then PYTHONPATH=. coverage run --source payu -m py.test -s test; fi;
     - cd docs && make html
 after_success:
     - coverage report -m
--- a/payu/models/mitgcm.py
+++ b/payu/models/mitgcm.py
@@ -99,9 +99,9 @@ class Mitgcm(Model):
                                              self.restart_calendar_file)
         # TODO: Sort this out with an MITgcm user
         try:
-            dt = data_nml['parm03']['deltat']
+            dt = float(data_nml['parm03']['deltat'])
         except KeyError:
-            dt = data_nml['parm03']['deltatmom']
+            dt = float(data_nml['parm03']['deltatmom'])
 
         # Runtime is set either by timesteps (ntimesteps) or physical
         # time (startTime and endTime).
@@ -120,7 +120,7 @@ class Mitgcm(Model):
 
                 if n_timesteps is None:
                     print("Calculated n_timesteps from starttime and endtime")
-                    n_timesteps = (t_end - t_start) / dt
+                    n_timesteps = round((t_end - t_start) / dt)
             else:
                 # Assume n_timesteps and dt set correctly
                 pass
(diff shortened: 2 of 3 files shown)


aekiss:
@aidanheerdegen are you sure you want `n_timesteps` and `n_iter0` to be floats? They aren't guaranteed to be whole numbers.

aidanheerdegen:
> @aidanheerdegen are you sure you want `n_timesteps` and `n_iter0` to be floats? They aren't guaranteed to be whole numbers.

It's a good point. Mostly I wanted to make some tests so I could change the MITgcm driver to support fractional time steps and not break existing functionality.

I guess that is why I had the `//` divisions to begin with.  I'll look into forcing those variables to be integer. Tah.

aekiss:
I guess if MITgcm supports fractional timesteps it would be best to do whatever MITgcm does, at the same numerical precision? I think python uses 64-bit double precision by default.

aidanheerdegen:
MITgcm supports arbitrary model time steps, but clearly the number of time steps is an integer.

I'm not attempting to cover all possible use cases, more coding to the tests, which I'll take as a reasonable representation of usage. If `payu` breaks because of different use cases then I'll look at adding those to the tests and changing the code appropriately. Otherwise it is pretty easy to lose your mind and a great deal of time.

aidanheerdegen:
Whaddya reckon @aekiss? Missed anything obvious?

aekiss:
Does `n_timesteps` need to be the actual number of timesteps that MITgcm will take? 

If MITgcm simply adds `dt` to `t_start` over and over until time exceeds `t_end` it could give a different number of iterations from your rounded answer when `dt` is a float, depending on whether the summation rounds high (reducing the iterations by one) or low (increasing number of iterations). (In the worst case (model time being much larger than `dt`), maybe MITgcm could fail to increment model time at all due to catastrophic roundoff and get in an endless loop?)

Anyway if you need to match the number of MITgcm timesteps, the surest way would be to do the summation in the same way as MITgcm, to the same precision, and count the number of iterations. Crazy I know, but I can't think of a better way.

Incidentally I think this sort of hassle is one reason that `numy.arange` excludes the limiting value.

aekiss:
Also if it were me I wouldn't force `t_start` and `t_end` to be floats. If you leave them as `int` you can have roundoff-free integer arithmetic in the (probably most common) case of integer `dt`.

aidanheerdegen:
Thanks. I don't like the idea of the code behaving in a different way depending on the implicit type of a number that it reads in. I think I'd prefer it to be consistent and fail, consistently. But point taken.

As for copying what MITgcm does, not going there unless it fails. It is just a driver program. If it fails it should do so loudly and then it can be fixed.

I'll merge this as Jim is waiting for a fix.

PR #261, opened 2020-05-14
Try and get coveralls working

aidanheerdegen:
Coveralls is reporting no coverage, try and get it working.


aidanheerdegen:
I have this half-remembered idea that coveralls won't work unless there is some valid data in `master`, so I'm going to merge this and hope for the best.

aidanheerdegen:
It is at least reporting that it is uploading something to coveralls.io

PR #262, opened 2020-05-14
Another attempt to get coveralls working

aidanheerdegen:
Added some debugging to check .coverage file persists


aidanheerdegen:
This works. Shame it doesn't allow for using from `after_success`

PR #296, opened 2021-02-12
Scan control path for storage flags

### LINKED ISSUE #264: Control directory is not included when checking for storage flag
`payu` parses all manifests to check for mount points to add to the `-l storage` option for PBS on `gadi`. It does not seem to check the path to the control directory which can lead to errors like this
```
  File "/home/502/aph502/.local/bin/payu-run", line 11, in <module>
    load_entry_point('payu', 'console_scripts', 'payu-run')()
  File "/home/502/aph502/code/python/payu/payu/subcommands/run_cmd.py", line 116, in runscript
    lab = Laboratory(run_args.model_type, run_args.config_path,
  File "/home/502/aph502/code/python/payu/payu/laboratory.py", line 30, in __init__
    raise ValueError('Cannot determine model type.')
ValueError: Cannot determine model type.


payu: warning: Configuration file config.yaml not found!
```

aidanheerdegen:
Scan control path for storage flags. Closes #264


aidanheerdegen:
This was far too complicated for a simple enough end result.

@marshallward can you check you're happy with pushing defining the default control path into `readconfig`?

I actually think it is a good idea to put as many default settings as possible in one place, as it is they are sprinkled around a bit.

aidanheerdegen:
I've tested it, seems to work ok,  so I'm going to merge.

marshallward:
Sorry that I was unable to review this before merge.

I'm not too familiar with what `storages` is all about, but I can't see too much harm in adding the control path to the list.

(I suppose it may encourage more binaries in control, which could break the paradigm of a lean control directory, but no doubt there's a genuine need here so no matter.)

Does `control_path` still default back to the current working directory?  I can't really tell from the patch.  (Perhaps it doesn't matter?)

And it looks like `PBS.generate_command()` got absorbed into `PBS.submit()`.   I guess both were hanging around?

aidanheerdegen:
Yeah `control_path` defaults to the directory in which the `config.yaml` resides. So in the case of running `payu` from the control directory it reproduces the previous behaviour. But it doesn't do a `getcwd`, instead it takes the `abspath` of the `config.yaml` and then uses that directory:

https://github.com/payu-org/payu/blob/2886517442f99d67ccf5c2c3a3ebd6169493ec41/payu/fsops.py#L95-L97

Now it has just occurred to me that `payu` has a `--config` command line option which that change might break?
```
  --config CONFIG_PATH, -c CONFIG_PATH
                        Configuration file path
```
Was the point of that option that you could run multiple experiments with the same `config.yaml` but in separate control directories?

Bugger.

Yeah I removed `PBS.generate_command()` because it wasn't being called by anything. I'd separated the functionality for testing, and then I think you made some changes with the scheduler stuff that made it redundant. At least I think it wasn't used ... :)

marshallward:
Don't sweat the `--config` flag, it was another one of those things  I put in but no one ever used (or so I assume).

marshallward:
(I think the point was to allow a generic YAML input.  I always hated the name `config.yaml`, maybe this was a pathway to phasing it out.)

aidanheerdegen:
I only changed it so it was easier for one test, and didn't seem to break anything, but now I'm wondering if I should change it back. 

Regarding the `-l storage` stuff, it is because NCI now only mounts `/home/` and `/scratch/$PROJECT` on PBS jobs, and all other `/scratch` and `/g/data` mounts you need access to have to be specified when the PBS job is submitted. This is a right royal PITA when, for example, people want to save data to a different `/scratch` area than the project code they are running with. This is very common with our researchers. 

So now `payu` scans a bunch of paths and all the paths in the manifests to look for anything that looks like a `/scratch` or `/g/data` path, extracts the info it needs and then constructs a `-l storage=...` option for the PBS job.

This latest fix was for when people decided to have a control directory no in the `/home`, and also not in the default `/scratch` space. There was no way `payu` knew to add this to the storage flag. The job would submit ok, but then `payu` would complain that it couldn't read the `config.yaml`, which was quite confusing for users.

PR #297, opened 2021-03-22
Add postscript action to non-collate jobs

### LINKED ISSUE #289: postscript only runs if collating
The `postscript` hook currently only runs if the model does a `collate` step. 

https://github.com/payu-org/payu/blob/4dc4a2b2d431b337a55743ee1c870edced3c0e2a/payu/subcommands/collate_cmd.py#L112-L113

There are models that do not collate, and situations were models that can collate don't need to.

Currently `postscript` is the only hook that is available that is guaranteed to run after all possible collation, but it should also be available where collation is not required, for consistency.

aidanheerdegen:
Closes #289 

Add logic to run the `postscript` action for jobs which do not collate, which ensures `postscript` will *always* run once an output directory is guaranteed not to change as a result of an action by `payu`.

Tested and works fine. Not sure if there is a simple `pytest` I could add for this unfortunately.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -44,7 +44,7 @@ jobs:
         run: payu list
 
       - name: Lint
-        run: pylint --extension-pkg-whitelist=netCDF4,backports -E payu
+        run: pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=backports -E payu
 
       - name: Run tests
         run: |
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
Dunno why it was still a draft, but I'll assume you're ok with this @marshallward and merge it as it's a pretty minor change.

aidanheerdegen:
I remembered why it was still a draft, needed to update the docs! Damn.

PR #300, opened 2021-03-30
Added mail on final run option + test

### LINKED ISSUE #299: Add mail feature
Following on from the discussion https://github.com/payu-org/payu/discussions/295 about mail features it should be possible to add some minimal support for mailing users.

While it is possible for a user to add specific flags to `qsub_flags` there are a couple of extra cases that would be good to cover:

1. Mail when a series of runs has finished, i.e. when payu is invoked like `payu run -n N` and the Nth run completes
2. Mail when the job fails. This is currently not supported in PBS, though the documentation suggests it does have this feature, the error condition it references is an internal PBS state, and not the exit code of the job

aidanheerdegen:
Closes #299 

Adds a config option to mail user when last run of a sequence completes.

### DIFF
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -97,16 +97,15 @@ class PBS(Scheduler):
 
         n_runs_per_submit = pbs_config.get('runspersub', 1)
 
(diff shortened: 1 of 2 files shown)


navidcy:
Can't really comment on the python implementation of this, but *I applaud* the feature.

PR #310, opened 2021-09-03
Fix FMS collate for >9999 files

### LINKED ISSUE #309: Collating regional outputs for core counts exceeding 10,000
I have an ACCESS-OM2-01 simulation where I am trying to save some regional diagnostics. The simulation uses Andrew’s 10461 core count for MOM, meaning that the regional diagnostics routine (which writes out 1 netcdf tile per core) now has 6 digits in the filename after the .nc — like `rregionocean-2d30m-vorticity_z-3-hourly-mean-ym_2012_01.nc.010431` . 

It seems that payu doesn't ask mppnccombine to collate these files, likely because of this:
https://github.com/payu-org/payu/blob/9348acdf92ca18aae229fc06b0b716d4cd85e1aa/payu/models/fms.py#L65-L66

Is there a nice way to generalise this bit of code?

aidanheerdegen:
Closes #309 

Modified payu.models.fms.get_uncollated_files to support > 9999 files.

Added tests.

### DIFF
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -6,6 +6,7 @@
 from __future__ import print_function
 
 from collections import defaultdict
+from pathlib import Path
 import multiprocessing
 import os
 import resource as res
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
As this is currently stopping work I will merge shortly, so speak up if anyone (@marshallward @aekiss ?) spots any problems

aekiss:
looks fine to me @aidanheerdegen

PR #321, opened 2022-02-07
Tell SIS2 about restart runs for MOM6

angus-g:
At some point last year, SIS2 stopped ignoring its namelist for input filenames. I think this effectively meant that every run segment was starting with no ice. If the `SIS_input_nml` is present, it should get the same input type (i.e. restart or new run) as MOM6 itself.


aidanheerdegen:
I've pushed a new tag to get a new `conda` version as this sounds like a possibly pernicious and difficult to diagnose bug.

angus-g:
Thanks, yes it was a surprising discovery!

PR #324, opened 2022-03-29
Flag a MOM6 run as a restart if using prior restart path

angus-g:
Using the mechanism to point to an external restart directory:

```yaml
restart: /path/to/restart
```

doesn't modify the experiment counter, because as far as payu is concerned, this is the first run. This breaks the logic the MOM6 module was using to determine whether to flag a run as a restart or not. In the case that the prior restart path has been set, we want to explicitly perform a restart run.

Thanks to @julia-neme for (unfortunately) finding this!

### DIFF
--- a/payu/models/mom6.py
+++ b/payu/models/mom6.py
@@ -67,7 +67,8 @@ class Mom6(Fms):
 
         input_nml = f90nml.read(input_fpath)
 
-        if self.expt.counter == 0 or self.expt.repeat_run:
+        if ((self.expt.counter == 0 or self.expt.repeat_run) and
+                self.prior_restart_path is None):
             input_type = 'n'
         else:
             input_type = 'r'


aidanheerdegen:
Seems a straightforward change. 

Just to be explicit, would this have let to different behaviour if `input_type` is set incorrectly?

angus-g:
That's right, `input_type = 'r'` will try to start the run from the restart files, and `input_type = 'n'` will ignore any restart files and start a fresh run.

aidanheerdegen:
Huh, interesting. Not the same behaviour as MOM5 and other FMS models I guess. I'll merge.

angus-g:
What do the other models do with that parameter?

aidanheerdegen:
MOM5 at least didn't use it. Just reads in restart files that are either initial conditions or restarts from a previous run. In some cases if the restarts don't exist then it initialises to an "ideal state".

PR #326, opened 2022-04-08
Porting to pawsey

### LINKED ISSUE #323: Porting payu to non-NCI machinery
There has been some discussion in the context of a MOM6 project about the potential for porting Payu to non-NCI hardware. 

I would be interested in getting payu up and running on the Pawsey HPC system. This system uses Cray architecture with slurm as a scheduler. 

I notice that there appears to be some support for slurm in Payu, as there is a slurm schedular class. However, I'm unsure of the process for porting payu to another system.

aidanheerdegen:
Wrap ldd in try/except as executables on pawsey seem to be statically linked. 
Also for the same reason don't assume LD_LIBRARY_PATH is set.

Commented out call to load_modules. Pawsey has a lot of default modules
that it relies on, so can't reliably monkey with that.

Removed a couple of the bespoke flags Marshall added to the slurm
scheduler, and also explicitly pass through the PAYU environment
variables. Also set in the current environment, but that didn't seem
to make it through to the submitted job.

Closes #323

### DIFF
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -100,8 +100,11 @@ def lib_update(bin_path, lib_name):
 
     # TODO: Use objdump instead of ldd
     cmd = 'ldd {0}'.format(bin_path)
-    ldd_output = subprocess.check_output(shlex.split(cmd)).decode('ascii')
-    slibs = ldd_output.split('\n')
+    try:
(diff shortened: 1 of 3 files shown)


aidanheerdegen:
This is a place-holder to make it easier to find and focus attention on what is required to port to magnus @ pawsey. Some of the changes are only to get it working, and aren't intended to be merged into `master`.

PR #351, opened 2023-07-31
Default to add '-wdir' arguments 

### LINKED ISSUE #341: missing -wdir arguments
When testing a Spack build of `access-om2` using Payu, I was receiving the following errors:
```
ice: error reading coupling_nml
...
assertion failed: Input atm.nml does not exist.
```

I noticed that `-wdir` is missing from the arguments given to `mpirun`:
``
mpirun --mca io ompio --mca io_ompio_num_aggregators 1 -np 1  $SCRATCH/access-om2/work/1deg_jra55_ryf_spackv1.git/atmosphere/yatm.exe : -np 216  $SCRATCH//access-om2/work/1deg_jra55_ryf_spackv1.git/ocean/fms_ACCESS-OM.x : -np 24  $SCRATCH/access-om2/work/1deg_jra55_ryf_spackv1.git/ice/cice_auscom_360x300_24x1_24p.exe
``

### LINKED ISSUE #350: MVAPICH run option incompatible with submodels
This logic

https://github.com/payu-org/payu/blob/aacfd92570d6f33487ec4de47f9ad8b7a7fa8f12/payu/experiment.py#L594-L596

is incompatible with submodels, as it would require doing an `os.chdir()` multiple times, with the final command running in the final submodel `work_path`. 

We're in the process of changing the logic for when `-wdir` is used, so just wanted to check we could get rid of this @marshallward

jo-basevi:
Change logic in experiments  to default to adding `-wdir` to the arguments given to mpirun. 
Should close #341 and close #350

### DIFF
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -104,7 +104,8 @@ def lib_update(bin_path, lib_name):
     slibs = ldd_output.split('\n')
 
     for lib_entry in slibs:
-        if lib_name in lib_entry:
+        if lib_name in lib_entry and 'spack' not in lib_entry:
+            # Only load enviroment modules available - assuming spack built libs includes 'spack' in full path of library
             lib_path = lib_entry.split()[2]
 
             # pylint: disable=unbalanced-tuple-unpacking
@@ -112,7 +113,3 @@ def lib_update(bin_path, lib_name):
 
             module('unload', mod_name)
             module('load', os.path.join(mod_name, mod_version))
-            return '{0}/{1}'.format(mod_name, mod_version)
-
-    # If there are no libraries, return an empty string
-    return ''
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -92,24 +92,21 @@ def module(command, *args):
     exec(envs)
 
 
-def lib_update(bin_path, lib_name):
+def lib_update(required_libs, lib_name):
     # Local import to avoid reversion interference
     # TODO: Bad design, fixme!
     # NOTE: We may be able to move this now that reversion is going away
     from payu import fsops
 
-    # TODO: Use objdump instead of ldd
-    cmd = 'ldd {0}'.format(bin_path)
-    ldd_output = subprocess.check_output(shlex.split(cmd)).decode('ascii')
-    slibs = ldd_output.split('\n')
-
-    for lib_entry in slibs:
-        if lib_name in lib_entry and 'spack' not in lib_entry:
-            # Only load enviroment modules available - assuming spack built libs includes 'spack' in full path of library
-            lib_path = lib_entry.split()[2]
-
+    for lib_filename, lib_path in required_libs.items():
+        if lib_filename.startswith(lib_name) and lib_path.startswith('/apps/'): 
+            # Load nci's /apps/ version of module if required 
             # pylint: disable=unbalanced-tuple-unpacking
             mod_name, mod_version = fsops.splitpath(lib_path)[2:4]
 
             module('unload', mod_name)
             module('load', os.path.join(mod_name, mod_version))
+            return '{0}/{1}'.format(mod_name, mod_version)
+
+    # If there are no libraries, return an empty string
+    return ''
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -527,7 +527,7 @@ class Experiment(object):
 
             wdir_arg = '-wdir'
             if self.config.get('scheduler') == 'slurm':
-                # Slurm's launcher controls the working directory
+                # Option to set the working directory differs in slurm
                 wdir_arg = '--chdir'
             model_prog.append(f'{wdir_arg} {model.work_path}')
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -25,7 +25,7 @@ import yaml
 
 # Local
 from payu import envmod
-from payu.fsops import mkdir_p, make_symlink, read_config, movetree
+from payu.fsops import mkdir_p, make_symlink, read_config, movetree, required_libs
 from payu.schedulers.pbs import get_job_info, pbs_env_init, get_job_id
 from payu.models import index as model_index
 import payu.profilers
@@ -137,6 +137,10 @@ class Experiment(object):
         self.models = []
 
         submodels = self.config.get('submodels', [])
+
+        # Inject information about required dynamically loaded libraries into submodel configuration
+        for sm in submodels:
+            sm['required_libs'] = required_libs(sm['exe'])
 
         solo_model = self.config.get('model')
         if not solo_model:
@@ -515,7 +519,7 @@ class Experiment(object):
             # TODO: Check for MPI library mismatch across multiple binaries
             if mpi_module is None:
                 envmod.lib_update(
-                    model.exec_path_local,
+                    model.config.get('required_libs'),
                     'libmpi.so'
                 )
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -514,24 +514,18 @@ class Experiment(object):
             # Update MPI library module (if not explicitly set)
             # TODO: Check for MPI library mismatch across multiple binaries
             if mpi_module is None:
-                mpi_module = envmod.lib_update(
+                envmod.lib_update(
                     model.exec_path_local,
                     'libmpi.so'
                 )
 
             model_prog = []
 
-            # Our MVAPICH wrapper does not support working directories
-            if mpi_module is not None and mpi_module.startswith('mvapich'):
-                curdir = os.getcwd()
-                os.chdir(self.work_path)
-            else:
-                curdir = None
-                wdir_arg = '-wdir'
-                if self.config.get('scheduler') == 'slurm':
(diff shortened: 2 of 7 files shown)


jo-basevi:
Is the MVIPICH wrapper still used? If so, I could revert the changes and only add ‘-wdir’ arguments if mpi_module does not start with ‘mvapich’ rather than previously which was checking mpi_module starts with `openmpi`?

aidanheerdegen:
Looks like MPICH supports `-wdir` in any case, so this code is redundant and should be removed

https://www.mpich.org/static/docs/v3.4.x/www1/mpiexec.html

See https://github.com/payu-org/payu/issues/350

payu/envmod.py
```diff
@@ -104,15 +104,12 @@ def lib_update(bin_path, lib_name):
...
 
             # pylint: disable=unbalanced-tuple-unpacking
             mod_name, mod_version = fsops.splitpath(lib_path)[2:4]
 
             module('unload', mod_name)
-            module('load', os.path.join(mod_name, mod_version))
-            return '{0}/{1}'.format(mod_name, mod_version)
-
-    # If there are no libraries, return an empty string
-    return ''
```
aidanheerdegen, line 118: I can see that the return value is no longer used here, or anywhere else in the code. I'm not sure we should remove that functionality though. It doesn't do any harm, and it may be something that @marshallward would expect to be available, and I can see the utility in returning the result of this sort of inspection.

In addition, if we wanted to test this new functionality we should split out the logic for finding a `lib_name` in the `ldd` output, and return `mod_name, mod_version` from that function.

Then we could add a `pytest` test with some pre-generated `ldd` output that included `spack` libraries and check we get the correct return values.

aidanheerdegen (approved):

payu/envmod.py
jo-basevi: Ok, I've reverted those changes and added in a couple of tests. I've also pulled the changes recommended here: https://github.com/payu-org/payu/issues/341#issuecomment-1641192337

payu/fsops.py
```diff
@@ -171,3 +173,41 @@ def is_conda():
...
+    # libmpi_usempif08.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_usempif08.so.40 (0x00007f12671c0000)
+    # libmpi_usempi_ignore_tkr.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_usempi_ignore_tkr.so.40 (0x00007f1266fb4000)
+    # libmpi_mpifh.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_mpifh.so.40 (0x00007f1266d44000)
+    # libmpi.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi.so.40 (0x00007f12667c3000)
+
+    needed_libs = {}
+    cmd = 'ldd {0}'.format(bin_path)
+    ldd_out = subprocess.check_output(shlex.split(cmd)).decode('ascii')
+
+    # awk '$2 ~ /=>/{print $1" "$3}'
```
aidanheerdegen, line 207: Can probably get rid of this comment.

payu/fsops.py
```diff
@@ -171,3 +173,41 @@ def is_conda():
...
+    # libmpi.so.40 => /apps/openmpi/4.0.2/lib/libmpi.so.40 (0x00007fa493665000)
+    # libmpi_usempif08_Intel.so.40 => /apps/openmpi/4.0.2/lib/libmpi_usempif08_Intel.so.40 (0x00007fa492a7c000)
+    # libmpi_usempi_ignore_tkr_Intel.so.40 => /apps/openmpi/4.0.2/lib/libmpi_usempi_ignore_tkr_Intel.so.40 (0x00007fa492863000)
+    # libmpi_mpifh_Intel.so.40 => /apps/openmpi/4.0.2/lib/libmpi_mpifh_Intel.so.40 (0x00007fa4925cb000)
+
+    # Example 2:
+    # libmpi_usempif08.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_usempif08.so.40 (0x00007f12671c0000)
+    # libmpi_usempi_ignore_tkr.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_usempi_ignore_tkr.so.40 (0x00007f1266fb4000)
+    # libmpi_mpifh.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi_mpifh.so.40 (0x00007f1266d44000)
+    # libmpi.so.40 => $SPACKDIR/opt/spack/linux-rocky8-cascadelake/intel-2019.5.281/openmpi-4.1.5-ooyg5wc7sa3tvmcpazqqb44pzip3wbyo/lib/libmpi.so.40 (0x00007f12667c3000)
```
aidanheerdegen, line 201: In my opinion these examples should be moved into a test of `required_libs`.

payu/envmod.py
```diff
@@ -92,21 +92,15 @@ def module(command, *args):
...
-    cmd = 'ldd {0}'.format(bin_path)
-    ldd_output = subprocess.check_output(shlex.split(cmd)).decode('ascii')
-    slibs = ldd_output.split('\n')
-
-    for lib_entry in slibs:
-        if lib_name in lib_entry:
-            lib_path = lib_entry.split()[2]
-
+    for lib_filename, lib_path in required_libs.items():
+        if lib_filename.startswith(lib_name) and lib_path.startswith('/apps/'):
```
aidanheerdegen, line 102: `.startswith` definitely better than `in`

payu/experiment.py
```diff
@@ -138,6 +138,9 @@ def init_models(self):
 
         submodels = self.config.get('submodels', [])
 
+        for sm in submodels:
```
aidanheerdegen, line 141:
```suggestion
        # Inject information about required dynamically loaded libraries into submodel configuration
        for sm in submodels:
```
Maybe just add a comment here to help the (code) reader

aidanheerdegen (approved):
More feedback!

payu/experiment.py
jo-basevi: Thanks, I've added it in

payu/fsops.py
jo-basevi: Ok, I've separating the parsing of ldd output into a separate file so its simpler to test. The other option could be to mock out the ldd call?

payu/experiment.py
```diff
@@ -590,13 +593,6 @@ def run(self, *user_flags):
         if self.config.get('coredump', False):
             enable_core_dump()
 
-        # Our MVAPICH wrapper does not support working directories
-        if mpi_module.startswith('mvapich'):
-            curdir = os.getcwd()
-            os.chdir(self.work_path)
-        else:
-            curdir = None
-
```
aidanheerdegen, line 599: Just wanted to leave a comment to say deleting this relates directly to https://github.com/payu-org/payu/issues/350. 

This just can't work with submodels and there is no attempt to test for that, nor do we want to support something that is so limited, so it is being removed.

payu/experiment.py
```diff
@@ -514,19 +518,18 @@ def run(self, *user_flags):
...
 
             model_prog = []
 
-            if mpi_module.startswith('openmpi'):
-                # Our MPICH wrapper does not support a working directory flag
-                model_prog.append('-wdir {0}'.format(model.work_path))
-            elif self.config.get('scheduler') == 'slurm':
+            wdir_arg = '-wdir'
+            if self.config.get('scheduler') == 'slurm':
                 # Slurm's launcher controls the working directory
```
aidanheerdegen, line 530: I've never really understood this comment. Probably better to just do

```suggestion
                # Option to set the working directory differs in slurm
```

aidanheerdegen (approved):
Agreed you could mock `ldd` as an alternative to storing the output in a file, but seems a lot of hassle and I kind of like the resulting design as it isolates functionality, which is not only good for your testing, but allows someone else to call a function like `parse_ldd_output` independently of running `payu`, which can help with checking for errors, or ensuring the code is doing what you expect.

aidanheerdegen (approved):
Feel free to merge.

PR #353, opened 2023-08-02
Setting MODULEPATH via config.yaml

### LINKED ISSUE #347: Support setting MODULEPATH via config.yaml
Our ACCESS-OM3 configurations (e.g. [MOM6-CICE6](https://github.com/COSIMA/MOM6-CICE6)) rely on the user doing
```
module use /g/data/ik11/spack/0.20.1/share/modules/linux-rocky8-cascadelake
```
prior to `payu run` in order to put the right versions of dependencies in the `MODULEPATH` environment variable (see https://github.com/COSIMA/access-om3/wiki/Quick-start#running and https://github.com/payu-org/payu/issues/337).

This is error-prone, especially as we will likely require particular spack versions to support particular model releases. 

Would it be possible to modify payu to support a `moduleuse` list in `config.yaml`? It [already supports](https://payu.readthedocs.io/en/latest/config.html#miscellaneous) `modules`, so it seems appropriate for `config.yaml` to also record the `module use`. This way it will show up in the git runlog etc.

jo-basevi:
Add an option to add module directories to MODULEPATH via config.yaml. 

Should close #347


jo-basevi:
So a simple solution could be to add a list of directories to add to the MODULEPATH, e.g. `moduledirs`
`module use` by default prepends to the MODULEPATH, and it doesn't add the directory if it is already in MODULEPATH. 
So there could be ordering differences. For example, if prior to `payu run`:
```
$ module use dir1 
$ module use dir2 
$ echo $MODULEPATH 
dir2:dir1
```
And in config.yaml the order was  
```
moduledirs:
    - dir2 
    - dir1
```
MODULEPATH with remain `dir2:dir1` while if `module use` wasn't run before `payu run`, the order would be `dir1:dir2`

jo-basevi:
I prefer a list of additional module directories than to defining an entire MODULEPATH, however redefining MODULEPATH will have the benefit of a concretised order of modules. 
E.g. in config.yaml
```
env:
    MODULEPATH: /g/data/ik11/spack/0.20.1/share/modules/linux-rocky8-cascadelake:/g/data/hh5/public/modules:/etc/scl/modulefiles:/opt/Modules/modulefiles:/opt/Modules/v4.3.0/modulefiles:/apps/Modules/modulefiles
```

jo-basevi:
It could be nice grouping the user-defined enviroment modules and modulues directories together however that would add a breaking change to modules option.. 
e.g.
```
modules:
   use:
       - /g/data/ik11/spack/0.20.1/share/modules/linux-rocky8-cascadelake
   load:
       - parallelio/2.5.10-intel-2021.6.0
```

aidanheerdegen:
> It could be nice grouping the user-defined environment modules and modules directories together however that would add a breaking change to modules option

Agreed that does sound like nicest option. 

It is possible to keep backward compatibility by testing the return value of popping the `modules` value off the config dictionary and updating it to the new syntax. This is how it was done when `collate` transitioned from a boolean to a dictionary of options:

https://github.com/payu-org/payu/blob/c8e74243dd27c42129901efbcc82916e8113da84/payu/fsops.py#L80-L82

jo-basevi:
Thanks for that suggestion @aidanheerdegen! I've added that in

test/test_payu.py
```diff
@@ -222,8 +223,8 @@ def test_parse_ldd_output():
     with open(ldd_output_path, 'r') as f:
         ldd_output = f.read()
     required_libs = payu.fsops.parse_ldd_output(ldd_output)
-    assert(len(required_libs), 4) 
-    assert(required_libs['libmpi.so.40'], '/apps/openmpi/4.0.2/lib/libmpi.so.40')
+    assert(len(required_libs) == 4) 
+    assert(required_libs['libmpi.so.40'] == '/apps/openmpi/4.0.2/lib/libmpi.so.40')
```
aidanheerdegen, line 237: This is a fix to tests from #351? 

Oops! Sorry I missed those when reviewing.

test/test_payu.py
```diff
@@ -148,8 +148,19 @@ def test_read_config():
...
+    # Test transform legacy modules option 
+    with open(config_tmp, 'w') as f:
+        f.write('modules:\n   - test_user_defined_module')
+    
+    config = payu.fsops.read_config(config_tmp)
+    modules_config = config.get('modules', {})
+    
+    assert(modules_config.get('load', []) == ['test_user_defined_module'])
+    assert(modules_config.get('use', []) == [])
+
```
aidanheerdegen, line 163: It would be good to have some tests for `modules` in both the new and old syntax, and testing `use` as well.

It might be consistent to move the config file to the resources subdirectory you created?

docs/source/config.rst
```diff
@@ -427,17 +427,21 @@ Miscellaneous
...
+            - /path/to/module/dir
+         load:
+            - netcdf-c-4.9.0
+            - parallel-netcdf-1.12.3
+            - xerces-c-3.2.3
 
    This is seldom needed, because payu is good at automatically determining
    the environment modules required by model executables. If the modules
-   require `module use` inorder to be found, do this prior to `payu run`,
+   require `module use` inorder to be found, this command can also be run
```
aidanheerdegen, line 443:
```suggestion
   require `module use` in order to be found, this command can also be run
```

aidanheerdegen (approved):
Looks good, but I do think we should add a few more tests for the additional functionality.

test/test_payu.py
jo-basevi: Yes they are, I should've picked up on it sooner as it was picked up on the test warnings..

test/test_payu.py
jo-basevi: Ok, I've added a couple of tests and moved that config file. Generally is there a preference for creating/removing temporary test files when running the tests or using test files in the resources subdirectory?

aidanheerdegen: > Generally is there a preference for creating/removing temporary test files when running the tests or using test files in the resources subdirectory?

I'd say it depends. If there was a large number of test files based on, say, combinations of test variables, then it would make sense to generate them dynamically in a `setup` routine, and then remove them in `teardown`. In this case there aren't many files, and they're not large, and there is something appealing simple about being able to discover the file just by browsing the repo. Yes if it was encoded as a `dict` and written to a file it would be visible in the source code, but would be a little trickier to discover.

In the end there is a level of user preference as well.

aidanheerdegen (approved):
Looks good!

jo-basevi:
Nothings changed/added, I just squashed the commits so it's a little cleaner

PR #362, opened 2023-09-12
Add pypi build and publish to CD

### LINKED ISSUE #361: Remove payu from pypi?
The `payu` version on pypi is very out of date, it is 0.5.1 and doesn't support `python3`

https://pypi.org/project/payu/

This isn't a massive problem IMO. As [previously discussed](https://github.com/payu-org/payu/issues/339) users either use the `conda` channel or self-install a modified version from source.

However it seems that recently it [did trip someone up](https://forum.access-hive.org.au/t/payu-issues-on-setonix/1333/5). 

So I think we should either remove it from `pypi`, or automatically publish to pypi from the CI. Removing it means losing the name on pypi, which if it was desirable to have in the future would be bad.

@marshallward you own the pypi project, so it's probably up to you.

aidanheerdegen:
Add automated PyPI publishing to CD Workflow.

Closes #361

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -37,9 +37,9 @@ jobs:
     if: github.repository == 'payu-org/payu' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: 3.11
 
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -31,3 +31,37 @@ jobs:
           user: accessnri
           label: main
           token: ${{ secrets.anaconda_token }}
+
+  pypi-build:
+    name: Build package for PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: 3.11
+
+      - run: |
+          python3 -m pip install --upgrade build && python3 -m build
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./dist
+
+  pypi-publish:
+    # Split build and publish to restrict trusted publishing to just this workflow
+    needs: ['pypi-build']
+    name: Publish to PyPI.org
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v3
+
+      - name: Publish package distributions to PyPI
+        # This is version v1.8.10
+        uses: pypa/gh-action-pypi-publish@b7f401de30cb6434a1e19f805ff006643653240e
+        with:
+          packages_dir: artifact/
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -61,6 +61,8 @@ jobs:
       id-token: write
     steps:
       - uses: actions/download-artifact@v3
+        with:
+          path: artifact/
 
       - name: Publish package distributions to PyPI
         # This is version v1.8.10
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -11,7 +11,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       - name: Checkout source
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup conda environment
         uses: conda-incubator/setup-miniconda@v2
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
Seeing as I managed to accidentally publish 1.0.19 to PyPI I guess that bit actually works. 

https://pypi.org/project/payu/1.0.19/

I'll leave that published version up, as we currently have no candidate for a tagged release.

.github/workflows/CD.yml
```diff
@@ -31,3 +31,39 @@ jobs:
           user: accessnri
           label: main
           token: ${{ secrets.anaconda_token }}
+
+  pypi-build:
+    name: Build package for PyPI
+    if: github.repository == 'payu-org/payu' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
dougiesquire, line 40: This can now be v4 (also above)

.github/workflows/CD.yml
```diff
@@ -31,3 +31,39 @@ jobs:
...
+
+      - uses: actions/setup-python@v3
```
dougiesquire, line 42: This can now be v4

.github/workflows/CD.yml
```diff
@@ -31,3 +31,39 @@ jobs:
...
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v3
+
+      - name: Publish package distributions to PyPI
+        # This is version v1.8.10
+        uses: pypa/gh-action-pypi-publish@b7f401de30cb6434a1e19f805ff006643653240e
+        with:
+          packages_dir: artifact/
```
dougiesquire, line 69: I don't recall, is an `artifact` directory created by `download-artifact@v3`? [The docs](https://github.com/actions/download-artifact#compatibility-between-v1-and-v2v3) aren't very clear about what happens when a name is not provided, but

> With v2 and v3, when an artifact is specified by the name input, there is no longer an extra directory that is created if the path input is not provided. All the contents are downloaded to the current working directory.

dougiesquire (commented):
A couple of comments/questions

Also, is there any value in having the conda build use the PyPI dist, e.g. [here](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/main/.conda/meta.yaml#L9)? This ensures that the PyPI and conda dists are always the same (though I don't know how they wouldn't be...) and makes the conda build dependent on a successful PyPI build (which could be seen as good or bad...)

.github/workflows/CD.yml
aidanheerdegen:
```suggestion
      - uses: actions/setup-python@v4
```
Suggestions are great for this sort of small change.

.github/workflows/CD.yml
aidanheerdegen:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/CD.yml
aidanheerdegen: Ugh. Just sloppy from me, cutting n' pasting from others examples. Will check and fix.

aidanheerdegen:
Thanks for the review. Lots of good feedback.

> Also, is there any value in having the conda build use the PyPI dist

Probably there is yes, though this is mostly about not having a completely out of date version on PyPI, but that is not intended to be the official distribution channel. 

Having said which, I always found it confusing that `conda` and `pip install` seemed to use slightly different, but overlapping, configuration/meta-data.  If they can be made to be completely compatible then yeah, probably a good idea. 

Not sure I am the person to do that, so someone else** could either make those changes in this PR, or make a subsequent PR to align them.


** @dougiesquire

.github/workflows/CD.yml
dougiesquire: True. Annoyingly, I don't think I can make suggestions on code that hasn't changed (e.g. the `actions/checkout@v3` used in the `conda` job)?

dougiesquire:
> could either make those changes in this PR, or make a subsequent PR to align them

Let's merge this since it does the job (hopefully), and maybe leave that for another PR?

.github/workflows/CD.yml
aidanheerdegen: Hrm. That is annoying, I can't batch commit my own suggestions.

aidanheerdegen: @dougiesquire can you batch up and commit my suggestions? Not being lazy (am), but also curious if that is the problem, that they're my suggestions.

.github/workflows/CD.yml
aidanheerdegen: > I don't think I can make suggestions on code that hasn't changed

Yeah I think you're right.

aidanheerdegen: https://github.com/github/roadmap/issues/347

.github/workflows/CD.yml
aidanheerdegen: Fixed. Specify download directory for artifact. It is bad to rely on magic default behaviour anyway.

dougiesquire (approved):

PR #364, opened 2023-09-18
Add a check for MOM6 parameter filenames defined in input.nml

### LINKED ISSUE #325: Read MOM6 parameter filenames from experiment
At the moment, the list of parameter filenames for MOM6/SIS2 is hardcoded: https://github.com/payu-org/payu/blob/3f00d4b72390053ab9182e340c582fbd0a67683f/payu/models/mom6.py#L39-L55

For the non-FMS components of these models, the filenames are actually specified in the `input.nml`, under the `parameter_filename` entry of the `MOM_input_nml` and `SIS_input_nml` namelists:

```&MOM_input_nml
        output_directory = '.',
        input_filename = 'n',
        restart_input_dir = 'INPUT',
        restart_output_dir = 'RESTART',
        parameter_filename = 'MOM_input'
/

&SIS_input_nml
        output_directory = '.',
        input_filename = 'n',
        restart_input_dir = 'INPUT',
        restart_output_dir = 'RESTART',
        parameter_filename = 'SIS_input'
/
```

jo-basevi:
In this branch, I've tagged on some code that runs at the end of `Mom6.setup()` that checks if files listed in `parameter_file` of `MOM_input_nml` and `SIS_input_nml` namelists in input.nml, are included in mom6's hard coded `config_files` or `optional_config_files`.

If the file is not present, I've added a warning, added the file to `config_files` and then copied the file to the workpath.
Or is just a warning/error printed out during setup sufficient? If not, should the files be treated as `config_files` or `optional_config_files` - i.e. raises an exception if file not found or not?

From what I can tell, the only usages of `optional_config_files` and `config_files` are:
- to be added to the runlog manifest (which happens during Experiment.run())
- config files are copied from the control path to the work path (at Model.setup())

Should close #325

### DIFF
--- a/payu/models/mom6.py
+++ b/payu/models/mom6.py
@@ -10,6 +10,7 @@
 
 # Standard library
 import os
+import shutil
 
 # Extensions
 import f90nml
@@ -59,6 +60,7 @@ class Mom6(Fms):
         super(Mom6, self).setup()
 
         self.init_config()
+        self.add_parameter_config_files()
 
     def init_config(self):
         """Patch input.nml as a new or restart run."""
@@ -78,3 +80,31 @@ class Mom6(Fms):
             input_nml['SIS_input_nml']['input_filename'] = input_type
 
         f90nml.write(input_nml, input_fpath, force=True)
+
+    def add_parameter_config_files(self):
+        """Check that the parameter files listed in input.nml are in the
+        model's configuration files"""
+        input_nml = f90nml.read(os.path.join(self.work_path, 'input.nml'))
+
+        # Set of all configuration files
+        config_files = set(self.config_files).union(self.optional_config_files)
+
+        for input in ['MOM_input_nml', 'SIS_input_nml']:
+            input_namelist = input_nml.get(input, {})
+            parameter_files = input_namelist.get('parameter_filename', [])
+
+            if isinstance(parameter_files, str):
+                parameter_files = [parameter_files]
+
+            for filename in parameter_files:
+                if filename not in config_files:
+                    print(f"payu: warning: parameter file {filename} "
+                          f"listed under {input} in input.nml is not in "
+                          "mom6's configuration files")
+
+                    # Extend config files
+                    self.config_files.append(filename)
+
+                    # Copy parameter file from control path to work path
+                    file_path = os.path.join(self.control_path, filename)
+                    shutil.copy(file_path, self.work_path)
--- a/payu/models/mom6.py
+++ b/payu/models/mom6.py
@@ -32,27 +32,12 @@ class Mom6(Fms):
 
         self.config_files = [
             'input.nml',
-            'MOM_input',
             'diag_table',
         ]
 
-        # TODO: Need to figure out what's going on here with MOM6
         self.optional_config_files = [
             'data_table',
-            'data_table.MOM6',
-            'data_table.OM4',
-            'data_table.SIS',
-            'data_table.icebergs',
-
-            'field_table',
-
-            'MOM_override',
-            'MOM_layout',
-            'MOM_saltrestore',
-
-            'SIS_input',
-            'SIS_override',
-            'SIS_layout',
+            'field_table'
         ]
 
     def setup(self):
@@ -60,7 +45,7 @@ class Mom6(Fms):
         super(Mom6, self).setup()
 
         self.init_config()
-        self.add_parameter_config_files()
+        self.add_config_files()
 
     def init_config(self):
         """Patch input.nml as a new or restart run."""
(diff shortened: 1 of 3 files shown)


aidanheerdegen:
I would have asked @angus-g for a review, but he wasn't a team member. Have added him, so he should be able to review this.

aidanheerdegen:
I've pinged you for a review @angus-g

payu/models/mom6.py
```diff
@@ -78,3 +80,31 @@ def init_config(self):
...
+                    print(f"payu: warning: parameter file {filename} "
+                          f"listed under {input} in input.nml is not in "
+                          "mom6's configuration files")
+
+                    # Extend config files
+                    self.config_files.append(filename)
+
+                    # Copy parameter file from control path to work path
+                    file_path = os.path.join(self.control_path, filename)
+                    shutil.copy(file_path, self.work_path)
```
angus-g, line 110: This makes me think we should fix up the (optional) config files definitions. For an ocean-only run, the only _required_ files are `input.nml` and `diag_table`. In fact, soon (but as far as I can tell, not yet), this could be either `diag_table` or `diag_table.yaml`, if using a YAML-enabled FMS.

I don't think we have the ability to distinguish between coupled ice-ocean and ocean-only runs, so `data_table` and `field_table` should be the only optional config files. I don't think the `data_table.xxx` files are ever read by the model, they were probably included because they exist in the [MOM6-examples](https://github.com/NOAA-GFDL/MOM6-examples) repository. Once again, with a YAML-enabled FMS, it might be the case that `data_table.yaml` could be valid.

The `MOM_override`, `MOM_layout`, `MOM_saltrestore`, `SIS_input`, `SIS_override` and `SIS_layout` entries could probably all be removed: they're now dynamically determined by this code. Instead of raising a warning if a file is specified in `input.nml` but _not_ in the (optional) config files, I think it should raise a warning if there aren't any entries in `MOM_input_nml` (or `SIS_input_nml`) at all. Perhaps in the latter case, it would be if the key is present but empty, as an ocean-only run wouldn't have that key present at all, and we probably don't want to be raising a warning in that case.

test/models/test_mom6.py
```diff
@@ -0,0 +1,144 @@
...
+            ['MOM_input', 'MOM_layout'],
+            ['SIS_input', 'New_SIS_file'],
+            ['New_SIS_file']
+        ),
+        (
+            ['New_MOM_file', 'MOM_input'],
+            None,
+            ['New_MOM_file']
+        )
+    ])
```
angus-g, line 115: Maybe one of these should test the case where the `mom_parameter_files` or `sis_parameter_files` is a bare string, rather than a list (wrt lines 96-97 in mom6.py)?

angus-g (commented):
This is a much nicer way of dealing with things! I've left a comment on how to take this a bit further to tidy up the lists of config files.

payu/models/mom6.py
jo-basevi: Thanks for the review! I am looking into implementing the suggested changes now. Regarding diag_table/diag_table.yaml (and data_table/data_table.yaml), would there be any way to know what option is required other than just checking that there is one of them present in the control directory? Also, would I leave `MOM_input` out of the hard-coded `config_files` or is it useful enough for guaranteeing that `MOM_input` is always present?

test/models/test_mom6.py
jo-basevi: I think the `f90nml.write()` method, writes lists of one elements as a string. But I can make it more clear in the code, and explicitly use test inputs of strings rather than lists with 1 element.

payu/models/mom6.py
angus-g: > Regarding diag_table/diag_table.yaml (and data_table/data_table.yaml), would there be any way to know what option is required other than just checking that there is one of them present in the control directory?

I think the YAML configs are very cutting-edge, I'd probably not worry about it for now. I don't know that there would be a good way to figure out what the model will read though, given it depends on whether FMS is compiled with/without a specific preprocessor define!

> Also, would I leave `MOM_input` out of the hard-coded `config_files` or is it useful enough for guaranteeing that `MOM_input` is always present?

It would be fairly convention-defying to use a different name for the main configuration file, but I don't see anything wrong with doing that. Since you're populating `self.config_files` at this point anyway, I guess there'll be a guarantee that _all_ the required parameter files are present, regardless of the default? I suppose that's why I suggested a warning if there aren't any parameter files in the namelist at all.

test/models/test_mom6.py
angus-g: It's probably not too big an issue, just a suggestion to hit that extra code path. In fact, given that namelists are typed and the `parameter_filename` field is an array, it's probably just a quirk of f90nml and doesn't explicitly need to be tested...

payu/models/mom6.py
jo-basevi: Thanks, I've added a commit that tidies up the config_files and optional_config_files, and refactored it slightly so if the `diag_table/diag_table.yaml` logic could easily be added to the `add_config_files` method if needed. I've added in a warning if there aren't any parameter files for MOM_input_nml (and  SIS_input_nml if present).

angus-g (approved):
Excellent, thanks @jo-basevi, this looks good now!

PR #366, opened 2023-09-27
Refactor experiment.set_counter() and some minor bug fixes

### LINKED ISSUE #285: Crash if restart exists in archive
If a file or directory named `restart` exists in the `archive` directory `payu` crashes with this error:
```
$ touch archive/restart
$ ls archive/
output000  output001  pbs_logs  restart  restart000  restart001
$ payu sweep
laboratory path:  /scratch/v45/aph502/access-om2
binary path:  /scratch/v45/aph502/access-om2/bin
input path:  /scratch/v45/aph502/access-om2/input
work path:  /scratch/v45/aph502/access-om2/work
archive path:  /scratch/v45/aph502/access-om2/archive
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-20.10/bin/payu", line 10, in <module>
    sys.exit(parse())
  File "/home/502/aph502/code/python/payu/payu/cli.py", line 42, in parse
    run_cmd(**args)
  File "/home/502/aph502/code/python/payu/payu/subcommands/sweep_cmd.py", line 16, in runcmd
    expt = Experiment(lab)
  File "/home/502/aph502/code/python/payu/payu/experiment.py", line 88, in __init__
    self.set_counters()
  File "/home/502/aph502/code/python/payu/payu/experiment.py", line 191, in set_counters
    self.counter = 1 + max([int(d.lstrip('restart'))
  File "/home/502/aph502/code/python/payu/payu/experiment.py", line 191, in <listcomp>
    self.counter = 1 + max([int(d.lstrip('restart'))
ValueError: invalid literal for int() with base 10: ''
```
even if correctly named restart files exist in `archive`, e.g. `restart000`

### LINKED ISSUE #306: Missing executable does not throw an exception
Recent issue with a malformed executable path only warned user and `payu run` submitted ok, and then the model error with a message that didn't give much information about the real problem
```
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):
 PML add procs failed
 --> Returned "Error" (-1) instead of "Success" (0)
```

During `payu setup` the malformed path triggered this warning:
```
Setting up atmosphere
File not found: /g/data/ik11/inputs/acce ss-om2/bin/yatm_1bb8904.exe
```
but this should be an exception which stops `setup` finishing for executables.

Maybe add some logic here:

https://github.com/payu-org/payu/blob/master/payu/models/model.py#L289

### LINKED ISSUE #307: archive step assumed LD_LIBRARY_PATH populated
In the `archive` step `payu` assumed `LD_LIBRARY_PATH` is populated and errors if it doesn't:

https://github.com/payu-org/payu/blob/master/payu/experiment.py#L781

```python
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-21.07/bin/payu", line 10, in <module>                                          
    sys.exit(parse())
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-21.07/lib/python3.9/site-packages/payu/cli.py", line 42, in parse               
    run_cmd(**args)
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-21.07/lib/python3.9/site-packages/payu/subcommands/archive_cmd.py", line 18, in runcmd
    expt.archive()
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-21.07/lib/python3.9/site-packages/payu/experiment.py", line 781, in archive     
    ld_libpaths = os.environ['LD_LIBRARY_PATH']
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-21.07/lib/python3.9/os.py", line 679, in __getitem__                            
    raise KeyError(key) from None
```

Should use `get` and supply an appropriate default if it doesn't exist.

jo-basevi:
- Fix to payu crashing with empty restart files/dirs
- Add default values of LD_LIBRARY_PATHS
- Raise exceptions for missing model executables

Closes #285, closes #307, closes #306

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -13,6 +13,7 @@ import datetime
 import errno
 import getpass
 import os
+import re
 import resource
 import sys
 import shlex
@@ -172,40 +173,38 @@ class Experiment(object):
 
         # Initialize counter if unset
         if self.counter is None:
-            # TODO: this logic can probably be streamlined
-            try:
-                restart_dirs = [d for d in os.listdir(self.archive_path)
-                                if d.startswith('restart')]
-            except EnvironmentError as exc:
-                if exc.errno == errno.ENOENT:
-                    restart_dirs = None
-                else:
-                    raise
-
-            # First test for restarts
-            if restart_dirs:
-                self.counter = 1 + max([int(d.lstrip('restart'))
-                                        for d in restart_dirs
-                                        if d.startswith('restart')])
+            # Check for restart index
+            max_restart_index = self.max_output_index(output_type="restart")
+            if max_restart_index:
+                self.counter = 1 + max_restart_index
             else:
-                # repeat runs do not generate restart files, so check outputs
-                try:
-                    output_dirs = [d for d in os.listdir(self.archive_path)
-                                   if d.startswith('output')]
-                except EnvironmentError as exc:
-                    if exc.errno == errno.ENOENT:
-                        output_dirs = None
-                    else:
-                        raise
-
-                # First test for restarts
-                # Now look for output directories
-                if output_dirs:
-                    self.counter = 1 + max([int(d.lstrip('output'))
-                                            for d in output_dirs
-                                            if d.startswith('output')])
+                # Now look for output directories,
+                # as repeat runs do not generate restart files.
+                max_output_index = self.max_output_index()
+                if max_output_index:
+                    self.counter = 1 + max_output_index
                 else:
                     self.counter = 0
+
+    def max_output_index(self, output_type="output"):
+        """Given a output directory type (output or restart),
+        return the maximum index of output directories found"""
+        try:
+            output_dirs = self.list_output_dirs(output_type)
+        except EnvironmentError as exc:
+            if exc.errno == errno.ENOENT:
+                output_dirs = None
+            else:
+                raise
+
+        if output_dirs and len(output_dirs):
+            return max([int(d.lstrip(output_type)) for d in output_dirs])
+
+    def list_output_dirs(self, output_type="output"):
+        """Return a list of restart or output directories in archive"""
+        naming_pattern = re.compile(fr"^{output_type}[0-9][0-9][0-9]$")
+        return [d for d in os.listdir(self.archive_path)
+                if naming_pattern.match(d)]       
 
     def set_stacksize(self, stacksize):
 
@@ -749,8 +748,7 @@ class Experiment(object):
                                           default_restart_history)
 
         # Remove any outdated restart files
-        prior_restart_dirs = [d for d in os.listdir(self.archive_path)
-                              if d.startswith('restart')]
+        prior_restart_dirs = self.list_output_dirs(output_type="restart")
 
         for res_dir in prior_restart_dirs:
 
@@ -766,10 +764,12 @@ class Experiment(object):
                     shutil.rmtree(res_path)
 
         # Ensure dynamic library support for subsequent python calls
-        ld_libpaths = os.environ['LD_LIBRARY_PATH']
+        ld_libpaths = os.environ.get('LD_LIBRARY_PATH', None)
         py_libpath = sysconfig.get_config_var('LIBDIR')
-        if py_libpath not in ld_libpaths.split(':'):
-            os.environ['LD_LIBRARY_PATH'] = ':'.join([py_libpath, ld_libpaths])
+        if ld_libpaths is None:
+            os.environ['LD_LIBRARY_PATH'] = py_libpath
+        elif py_libpath not in ld_libpaths.split(':'):
+            os.environ['LD_LIBRARY_PATH'] = f'{py_libpath}:{ld_libpaths}'
 
         collate_config = self.config.get('collate', {})
         collating = collate_config.get('enable', True)
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -561,7 +561,7 @@ class Experiment(object):
 
             for prof in self.profilers:
                 if prof.runscript:
-                    model_prog = model_prog.append(prof.runscript)
+                    model_prog.append(prof.runscript)
 
             model_prog.append(model.exec_prefix)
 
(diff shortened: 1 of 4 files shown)


payu/experiment.py
```diff
@@ -172,41 +173,39 @@ def set_counters(self):
...
+
+        if output_dirs and len(output_dirs):
+            return max([int(d.lstrip(output_type)) for d in output_dirs])
+
+    def list_output_dirs(self, output_type="output"):
+        """Return a list of restart or output directories in archive"""
+        naming_pattern = re.compile(fr"^{output_type}[0-9][0-9][0-9]$")
+        return [d for d in os.listdir(self.archive_path)
+                if naming_pattern.match(d)]       
+
```
aidanheerdegen, line 208: Big improvement! Thanks.

payu/models/model.py
```diff
@@ -286,6 +286,12 @@ def setup(self):
             # If have exe manifest this implies exe reproduce is True. Do not
             # want to overwrite exe manifest in this case
             if not self.expt.manifest.have_manifest['exe']:
+                # Check whether executable path exists
+                if not os.path.isfile(self.exec_path):
```
aidanheerdegen, line 290: I wonder if we're going to the trouble of checking, if we shouldn't also check if it is executable. I've tested and it gives the same error when the executable doesn't have the executable permissions:
```
mpirun was unable to launch the specified application as it could not access
or execute an executable:
```

`os.access` can be used to test for executable status

https://docs.python.org/3/library/os.html#os.access

So 
```python
os.access(self.exec_path, os.X_OK)
```
should do the trick. While you can do the existence check and the executable check in one step with `os.access` the message to the user should be different for the two cases, so probably best to check them separately, and sequentially.

payu/models/model.py
```diff
@@ -286,6 +286,12 @@ def setup(self):
             # If have exe manifest this implies exe reproduce is True. Do not
             # want to overwrite exe manifest in this case
             if not self.expt.manifest.have_manifest['exe']:
+                # Check whether executable path exists
```
aidanheerdegen, line 289: I'm slightly concerned about putting the checks here, even though I suggested this location I think I was wrong. 

The tests are only done if there is not an existing manifest. An existing manifest means we want to check for reproducibility. It will do the hash checks, which would fail if it was missing, but would not detect a change in file permissions.

I'd suggest moving your tests above line 288:
```python
if not self.expt.manifest.have_manifest['exe']:
```

aidanheerdegen (changes requested):
Looks great, thanks for cleaning that up and closing a bunch of issues. 

Just a couple of tweaks and it should be good to go.

payu/experiment.py
jo-basevi: Thanks!

payu/models/model.py
jo-basevi: I initially had the check above line 288. But it would fail this [manifest test](https://github.com/payu-org/payu/blob/edf8d8c538040447dcd3bdd944cba5bab86937c1/test/test_manifest.py#L217C1-L230) which makes the executable unfindable but tests that it still works because it has the same name as executable in the manifest. Do you think it would it be ok if I changed/removed that test?

payu/models/model.py
jo-basevi: Thanks, I'll add that in!

jo-basevi: I've added it in and then most of the tests broke because all the test exe files weren't executable. I changed the permission of the test files in the test code to be executable, even though their content is random bytes so any attempt to actually execute the files would result in an error..

jo-basevi:
A previous CI action failed at the lint stage because of ```payu/experiment.py:564:20: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)``` where the code was  in the payu run command:
```python
for prof in self.profilers:
      if prof.runscript:
          model_prog = model_prog.append(prof.runscript)
```
I think it found a legitimate error so I fixed it in the previous commit (though could be wrong). But I am bit confused as to how that error isn't picked up when running `pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=backports -E payu` locally or in other CI builds in other PRs..

payu/models/model.py
aidanheerdegen: Failing tests that are fixed are the best kind of tests! :)

aidanheerdegen:
> I think it found a legitimate error so I fixed it in the previous commit (though could be wrong). But I am bit confused as to how that error isn't picked up when running `pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=backports -E payu` locally or in other CI builds in other PRs..

Yeah that is weird. It is a not often used part of the code, but you're right that a limiter should have picked that up before. Is computing supposed to be deterministic?

payu/models/model.py
aidanheerdegen: Ok, maybe I've changed my mind *again* and we should remove that test, as it is testing to make sure the manifest is still correct even if the `config.yaml` has been changed so the paths to the executables is broken. In retrospect I don't think we should support that use case, and I doubt anyone is relying on it.

There is a tension between where the source of truth lies. It was always in `config.yaml`, but the manifests now also hold information, and when we want to reproduce a run, or make sure the executables haven't changed then the manifests become the source of truth. I think I need to think about this a bit more!

jo-basevi:
>Yeah that is weird. It is a not often used part of the code, but you're right that a limiter should have picked that up before. Is computing supposed to be deterministic?

Well it is now failing in another PR, so it'll be good to fix here..

payu/models/model.py
jo-basevi: Ok, I agree with not supporting invalid paths in `config.yaml` as that will at least help catch any mistakes. Unless it was practice to put invalid paths in to check that exe paths in the manifests were used? But in that case should the practice be to remove the exe path from the config.yaml?

aidanheerdegen (approved):
Good to merge I think. Thanks.

PR #368, opened 2023-09-29
Symlink instead of copy the `mod_def.ww3` input in the cesm_cmeps driver

### LINKED ISSUE #365: Copying input files anti-pattern in CESM_CMEPS driver
### Problem

The CESM_CMEPS driver copies an input file to the `work` directory (and renames to `mod_def.ww3`).

https://github.com/payu-org/payu/blob/master/payu/models/cesm_cmeps.py#L181

This is an anti-pattern for `payu` for a couple of reasons:

1. Input files are symbolically linked, that way the links can be removed during the `archive` step. The current implementation will save multiple identical copies of the `mod_def.ww3` file in the `outputXXX` directories. Each one is quite large, e.g. `26M` for a 320x384 domain.
2. The file is copied to the top level of the `work` directory, and so is not captured by the input manifest

### Possible fixes

These fixes assume the `mod_def.ww3` files *is not altered* when the model is run. If it is altered, and the updated version needs to be retained then it is effectively an output or restart file, and the following options won't work.

#### Option 1: File can be read from `input` dir

Ideally the model could be be made to read the `mod_def.ww3` from the `input` directory. In that case I would advise removing the code above entirely and making a symbolic link to the `mod_def.ww3` file in the original input directory.

For example, the config provided by @ezhilsabareesh8 uses this file:
```
/g/data/ik11/inputs/access-om3/0.1.0/cime/wav/ww3/ww3a.mod_def.ww3.wwver7.14.220506
```
so making a symbolic link to it in the same same directory would work
```
ln -s /g/data/ik11/inputs/access-om3/0.1.0/cime/wav/ww3/ww3a.mod_def.ww3.wwver7.14.220506 /g/data/ik11/inputs/access-om3/0.1.0/cime/wav/ww3/mod_def.ww3
```
Then the `config.yaml` could use either the directory as an input, `/g/data/ik11/inputs/access-om3/0.1.0/cime/wav/ww3`, or use the full path to the symbolic link `/g/data/ik11/inputs/access-om3/0.1.0/cime/wav/ww3/mod_def.ww3`.

#### Option 2: File cannot be read from the `input` dir

If the path to `mod_def.ww3` is hard-coded (and there is resistance to that being made user-settable parameter) then I would suggest altering the code linked above, and rather than copy the file, make a symbolic link to the linked file in the `input` directory.

A further version of this would be to follow the same steps above, and create a correctly named symbolic link in the original `input` directory, and then link to *that* link. This has the benefit that the manifest entry matches the name used by the model, so makes it a bit more obvious that this is the file that is actually used.

#### Summary

Both options avoid needless copying of an identical file, so are an improvement.

I favour option 1 if possible.  While both options capture the `mod_def.ww3` file used by the model in the input manifest, option 1 makes it clear that this was the file actually used by the model. Option 2 requires some inspection of the `payu` driver and/or the `work` directory to make this connection.

dougiesquire:
The WW3 model definition file input needs to be in the work directory and called `mod_def.ww3`. Previously this driver looked for a file in the inputs with "mod_def.ww3" in the filename and then copied that file to the work directory. This is bad practice. Now, the name of the file is assumed to be "mod_def.ww3" and this file is symlinked instead. Those setting up the configuration will need to ensure that a file with the right name is included in the inputs (e.g. using a symlink if the original input has a different name).

Closes #365

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -168,7 +168,8 @@ class CesmCmeps(Model):
 
         self.runconfig.write()
 
-        # Horrible hack to make a link to the mod_def.ww3 input in the work directory
+        # Horrible hack to make a link to the mod_def.ww3 input in the work
+        # directory
         # The ww3 mod_def input needs to be in work_path and called mod_def.ww3
         if "ww3dev" in self.components.values():
             f_name = "mod_def.ww3"
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -17,7 +17,7 @@ import glob
 import shutil
 import multiprocessing
 
-from payu.fsops import mkdir_p
+from payu.fsops import mkdir_p, make_symlink
 from payu.models.model import Model
 from payu.models.fms import fms_collate
 
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -168,17 +168,15 @@ class CesmCmeps(Model):
 
         self.runconfig.write()
 
-        # TODO: Should have a better way to do this
+        # Horrible hack to make a link to the mod_def.ww3 input in the work directory
         # The ww3 mod_def input needs to be in work_path and called mod_def.ww3
         if "ww3dev" in self.components.values():
-            files =  glob.glob(
-                os.path.join(self.work_input_path, "*mod_def.ww3*"),
-                recursive=False
-            )
-            if files:
-                assert len(files) == 1, "Multiple mod_def.ww3 files found in input directory"
-                f_dst = os.path.join(self.work_path, "mod_def.ww3")
-                shutil.copy(files[0], f_dst)
+            f_name = "mod_def.ww3"
+            f_src = os.path.join(self.work_input_path, f_name)
+            f_dst = os.path.join(self.work_path, f_name)
+
+            if os.path.isfile(f_src):
+                make_symlink(f_src, f_dst)
             else:
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')


payu/models/cesm_cmeps.py
```diff
@@ -168,17 +168,16 @@ def setup(self):
...
-                os.path.join(self.work_input_path, "*mod_def.ww3*"),
-                recursive=False
-            )
-            if files:
-                assert len(files) == 1, "Multiple mod_def.ww3 files found in input directory"
-                f_dst = os.path.join(self.work_path, "mod_def.ww3")
-                shutil.copy(files[0], f_dst)
+            f_name = "mod_def.ww3"
+            f_src = os.path.join(self.work_input_path, f_name)
+            f_dst = os.path.join(self.work_path, f_name)
```
aidanheerdegen, line 177: I like how straightforward this is. 

It will break some existing configs, so that should be communicated to those affected, but likely only a few people at this stage.

Just checking, there is no way to set the path of this file to be `input/mod_def.ww3`? That would be clearer, but I understand if this is hard-coded and not configurable then this is as good as it gets.

aidanheerdegen (approved):
Just a quickie question to make something clear, otherwise fine to merge if that is not something that can be changed.

payu/models/cesm_cmeps.py
dougiesquire: > It will break some existing configs, so that should be communicated to those affected, but likely only a few people at this stage.

Already done: https://github.com/COSIMA/MOM6-CICE6-WW3/issues/2

> Just checking, there is no way to set the path of this file to be input/mod_def.ww3? That would be clearer, but I understand if this is hard-coded and not configurable then this is as good as it gets.

Not that Ezhil or I know of. But we're still learning about WW3. If I do discover it's possible, I'll fix this up.

aidanheerdegen:
Merge away.

PR #370, opened 2023-10-04
Fix bug in set_counter

### LINKED ISSUE #369: Bug in set_counter
There's a bug in set counter which doesn't check max_output_index/max_restart_index for None, so counter is set incorrectly for the second run (i.e when max_output_index is 0)

Just needs if-statements in set_counter to be tweaked.

jo-basevi:
Explicitly check for None in set_counter if-statements, to set counter correctly when max restart/output index is 0 

Should close #369

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -88,6 +88,7 @@ class Experiment(object):
         self.set_expt_pathnames()
         self.set_counters()
 
+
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
Good catch!

PR #373, opened 2023-10-17
Read mom6 parameter files in cesm cmeps driver

### LINKED ISSUE #371: Read MOM6 parameter filenames from experiment in cesm_cmeps driver
The same issue described in #325 is also in the `cesm_cmeps` driver:

https://github.com/payu-org/payu/blob/cfd694aa4744b1c7578e18a8993650df9bd88bfe/payu/models/cesm_cmeps.py#L26-L32

In addition, `field_table` (and possibly `data_table`??) needs to be added to the `"optional_config_files"` key.

jo-basevi:
Refactored the code added in #364, so it can also be re-used for the `mom` component in `cesm-cmeps` driver.

I didn't add in `mixin` class in the end, as it just needed only one function (`mom6_add_parameter_files`) which I broke out of `mom6` driver class. Though please let me know if using a `mixin` class would still be better!

Should close #371

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -20,16 +20,19 @@ import multiprocessing
 from payu.fsops import mkdir_p, make_symlink
 from payu.models.model import Model
 from payu.models.fms import fms_collate
+from payu.models.mom6 import mom6_add_parameter_files
 
 # Add as needed
 component_info = {
     "mom": {
         "config_files": [
             "input.nml",
-            "MOM_input",
             "diag_table",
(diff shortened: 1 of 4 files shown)


aidanheerdegen:
Neat! I've pinged @dougiesquire and @angus-g for a review.

Angus, don't worry if you have no opinion, just wanted to keep you in the loop as this reuses the code @jo-basevi just added to the mom6 driver.

dougiesquire (approved):
LGTM. Thanks @jo-basevi!

PR #376, opened 2023-10-23
Fix date-based restart pruning breaking on empty restart dirs

### LINKED ISSUE #375: date-based restart pruning breaks with empty restart directories
Date-based restart pruning currently breaks with empty restart directories. Empty (or near empty) directories could arise from local delete files option in the (un-released) `payu sync`, or in cosima's [sync_data script](https://github.com/COSIMA/01deg_jra55_iaf/blob/9d980a63e6bfba64048733b7663a73ca56f306ba/sync_data.sh#L125). This is because of rsync's `--remove-local-files` flag which leaves the local directories present but removes any synced files. Date-based restart pruning then throws a NotImplementedError as it can't find the `ocean_solo.res` which it uses to parse the datetimes. 

The idea of the NotImplementedError was to catch it being used for models that don't have an `ocean_solo.res` restart file. It could be worth moving date-based parsing from the `fms` model to a Mixin class that can be used with both `mom6` and `mom5` model drivers. Then if `ocean_solo.res` isn't present, it could then just print a warning and exclude the directory from any restart pruning.

jo-basevi:
Fix issue described in #375, Closes #375 

Instead of raising an exception when restart file can't be found, a warning is printed and the restart is ignored from date-based restart pruning. 

Also, moved code for restart file parsing for `ocean_solo.res` to a mixin mom class. This mixin class is then pulled into the `mom` and `mom6` model drivers.


payu/experiment.py
```diff
@@ -1050,6 +1050,10 @@ def get_restarts_to_prune(self,
                           'model. To use integer based restart pruning, '
                           'set restart_freq to an integer value.')
                     raise
+                except FileNotFoundError as e:
+                    print(f'payu: warning: Ignoring {restart} from date-based '
+                          f'restart pruning. Error: {e}')
+                    continue
                 except Exception:
```
aidanheerdegen, line 1057: Hope you don't hate me ... but it's always good practice if you're fixing a specific bug to write a test for it, confirm the test fails and then implement the fix and confirm the test passes.

Now maybe it isn't necessary in this case, but it is a good habit to get into, to confirm

1. the bug is what you think it is (by writing a test that trips it)
2. the code fix does what you think it's doing, i.e. fixing the bug

I have learnt this the hard way.

aidanheerdegen (approved):
Looks good to me.

payu/experiment.py
jo-basevi: Ok I had tested it with some test model configuration and modifying local archive files. But I just added in an small test

aidanheerdegen: Awesome. Thanks.

test/test_prune_restarts.py
```diff
@@ -222,3 +222,38 @@ def test_prune_restarts(restarts,
...
+    # Create empty restart dirs
+    make_expt_archive_dir(type='restart', index=0)
+    make_expt_archive_dir(type='restart', index=2)
+
+    with cd(ctrldir):
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
+        expt = payu.experiment.Experiment(lab, reproduce=False)
+
+        # Function to test
+        restarts_to_prune = expt.get_restarts_to_prune()
```
aidanheerdegen, line 252: We don't use warnings in `payu` (and probably should) then it would be simple to check that this emitted the warning you added.

https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#warns

I'm not suggesting you add this, just sayin'

aidanheerdegen (approved):
Good to merge I think. Sorry it has taken a while to get back to it.

test/test_prune_restarts.py
aidanheerdegen: Feel free to resolve this convo once you've had a chance to read it.

jo-basevi: Thanks, that is good to know about warnings. I'll look into adding it with future work

PR #389, opened 2023-11-29
Fix for Issue #388 which broke executable reproducibility

### LINKED ISSUE #388: Specifying executables to reproduce fails to create links
If I set
```yaml
manifest:
  reproduce:
    exe: True
```
the symlinks for the executables are not created. If I remove that option, or use the command line `-r` option, the symlinks are created correctly. Also works if I set all to `True`
```yaml
manifest:
  reproduce:
    exe: True
    input: True
    restart: True
```
This is with the the [`1deg_jra55_ryf`](https://github.com/ACCESS-NRI/access-om2-configs/tree/1deg_jra55_ryf) ACCESS-OM2 model configuration.

```
$ payu --version
payu 1.0.19
```

aidanheerdegen:
Closes #388 which caused executable reproducibility to break in some situations.

Bug was due to a `break` rather than `continue` in `Manifest.setup()`. Links for other mainfests weren't created if `scaninputs=True`.

Modified existing test to trigger error. Test now passes.

### DIFF
--- a/test/test_manifest.py
+++ b/test/test_manifest.py
@@ -167,7 +167,7 @@ def test_exe_reproduce():
 
     assert(manifests == get_manifests(ctrldir/'manifests'))
     assert((ctrldir / local_exe_path).exists())
-    assert(Path(manifests['exe.yaml'][local_exe_path]['fullpath']) == 
+    assert(Path(manifests['exe.yaml'][local_exe_path]['fullpath']) ==
            (ctrldir / local_exe_path).resolve())
 
     bindir = labdir / 'bin'


aidanheerdegen:
Hey @jo-basevi. I know you're super busy, but this is one of the fixes I'd like to get in before we merge the UUID/clone/branch stuff so would appreciate a review.

It shouldn't be a big deal, as it is a pretty simple fix for a fairly obvious error (mine as it turns out).

jo-basevi (approved):
Changes look good to me!

PR #391, opened 2023-11-29
add cftime dependency to conda meta.yaml

jo-basevi:
Small change here adding cftime run dependency to `conda/meta.yaml`. This in used in date based restart pruning.

There's also a fix for a small bug in testing which creates a `work` directory in payu. Moved setup call so it runs inside the temporary test control directory.


aidanheerdegen (approved):
Looks good. Weird that test didn't throw an error.

Issue #332, opened 2022-11-18
Update docs to refer to Gadi not Raijin


dougiesquire:
The docs still refer to Raijin in a number of places. In some cases, instructions given don't work on Gadi, e.g. https://payu.readthedocs.io/en/latest/install.html#nci-users

PR #394, opened 2023-12-15
Docs update

Mark360s:
Issue #332
The Payu module could not be located using the Gadi directly, while we could use the 'conda' module from 'hh5' data storage to have the payu functions work.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -12,7 +12,7 @@ of this document.
 NCI Users
 =========
 
-Payu is available for all users on Raijin.
+Payu is available for all users on Gadi.
 
 To load payu, load the environment module::
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -13,10 +13,10 @@ NCI Users
 =========
 
 Payu is available for all users on Gadi.
-
-To load payu, load the environment module::
-
-   module load payu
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -12,11 +12,11 @@ of this document.
 NCI Users
 =========
 
-Payu is available for all users on Raijin.
-
-To load payu, load the environment module::
-
-   module load payu
+Payu is available for all users on Gadi.
+But it needs to access the 'hh5' data in NCI at first, we could use that conda module to work payu functions.
```
aidanheerdegen, line 16: Users need to be a member of the `hh5` project. It is probably best to just point to the [instructions on the ACCESS-Hive](https://access-hive.org.au/models/run-a-model/run-access-esm/#model-specific-prerequisites), as they are being actively updated and supported.

aidanheerdegen (changes requested):
Thanks for the contribution. As outlined in my comment, I think it is best to point to instructions on ACCESS-Hive.

https://access-hive.org.au/models/run-a-model/run-access-esm/#model-specific-prerequisites

docs/source/install.rst
Mark360s: Yes, thanks for your remind of this instructions.

PR #400, opened 2024-01-08
Update Packaging Logic

### LINKED ISSUE #398: Update packaging tools and logic
`payu` deploys to `pypi` and `conda`, but the configuration metadata is stored in separate files:

https://github.com/payu-org/payu/blob/master/setup.py

https://github.com/payu-org/payu/blob/master/conda/meta.yaml

It seems this is a [well known problem](https://github.com/conda/conda/issues/12462). Some solutions include:

https://github.com/basnijholt/unidep

https://stackoverflow.com/a/76722681

@dougiesquire suggested we could even utilise the PyPI build in `conda` packaging https://github.com/payu-org/payu/pull/362 but this left for later. Later has arrived.

The "modern" approach is to use a `pyproject.toml` file to specify dependencies

Related issue, the version is hard-coded:

https://github.com/payu-org/payu/blob/master/payu/__init__.py#L6

and used in a few locations

https://github.com/search?q=repo%3Apayu-org%2Fpayu%20__version__&type=code

so the pypi distribution only installs the version in that file. Hard-coding the version string in a file is a bad approach when using `git` tags for versioning, as it requires putting the version in the file, then tagging with the version in the file as a subsequent step. Nasty.

There are a number of options for "single-sourcing" the package version:

https://packaging.python.org/en/latest/guides/single-sourcing-package-version/#single-sourcing-the-version

But option 7 is the one that is most compatible with using `git` tags for versioning, and it uses [setuptools_scm](https://pypi.org/project/setuptools-scm/).

NB: A couple of the places the hard-coded version string is used are in some old install scripts that should be removed (see https://github.com/payu-org/payu/issues/396).

CodeGat:
In this PR:
* Update some of the CI/CD logic to be more clear
* Centralize the version of `payu` to a single location
* Probably more stuff as the PR develops

Closes #398

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -1,11 +1,14 @@
 name: CD
 
-on: [push, pull_request]
+on:
+  push:
+    tags:
+      - '*'
 
 jobs:
   conda:
     name: build and deploy to conda
-    if: github.repository == 'payu-org/payu' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    if: github.repository == 'payu-org/payu'
     runs-on: ubuntu-latest
 
     # Steps represent a sequence of tasks that will be executed as part of the job
@@ -34,7 +37,7 @@ jobs:
 
   pypi-build:
     name: Build package for PyPI
-    if: github.repository == 'payu-org/payu' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    if: github.repository == 'payu-org/payu'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


CodeGat:
Hey @dougiesquire I believe you are working on this issue linked in this PR? I'll close my PR as I didn't exactly get very far into it :D just a bit of CD refactoring...thanks for taking this one on.

dougiesquire:
Whoops sorry @CodeGat - didn't mean to tread on your toes. Aidan put a call out for help on this issue. Please feel free to add the changes you made here into the other PR I started

dougiesquire:
I went ahead and included your changes here as part of https://github.com/payu-org/payu/pull/403/commits/631c414397ce11ce45e0aa921d9aca7fda55d1b8 in #403. I hope you don't mind.

PR #407, opened 2024-01-22
Fix PyPI upload

### LINKED ISSUE #398: Update packaging tools and logic

dougiesquire:
The PyPI upload stopped working after #403 was merged. I can't say I understand why as #403 didn't touch the workflow steps that are causing the problem... Regardless, this PR should fix things.

The update to `actions/upload-artifact@v4` (from `v3`) is not necessary to fix the issue, but good to do anyway.

Closes #398 (again)

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -23,9 +23,10 @@ jobs:
       - run: |
           python3 -m pip install --upgrade build && python3 -m build
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          path: ./dist
+          name: release
+          path: dist
 
   pypi-publish:
     # Split build and publish to restrict trusted publishing to just this workflow
@@ -36,9 +37,10 @@ jobs:
       # IMPORTANT: this permission is mandatory for trusted publishing
       id-token: write
     steps:
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          path: artifact/
+          name: release
+          path: dist
 
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@2f6f737ca5f74c637829c0f5c3acd0e29ea5e8bf # v1.8.11
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -44,8 +44,6 @@ jobs:
 
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@2f6f737ca5f74c637829c0f5c3acd0e29ea5e8bf # v1.8.11
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/CD.yml
```diff
@@ -36,9 +37,10 @@ jobs:
       # IMPORTANT: this permission is mandatory for trusted publishing
       id-token: write
     steps:
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          path: artifact/
+          name: release
+          path: dist
```
aidanheerdegen, line 43: I noticed these didn't match, but just assumed this was correct. I guess it wasn't?

aidanheerdegen (approved):
Thanks for fixing so promptly. 

Happy for you to merge if you think it fixes the issue.

dougiesquire:
Sorry forgot one thing. Just tested publishing to TestPyPI from my fork and these changes fixed the issue.

.github/workflows/CD.yml
dougiesquire: Yeah. The strange thing is that I didn't touch these...

dougiesquire: Best guess is it had to do with the version change of the pypi-publish action

aidanheerdegen (approved):
Looks good to merge.

PR #423, opened 2024-03-06
Add bug fix for restart history with empty archive directory

### LINKED ISSUE #422: Payu date-based restart issue with MITgcm
Hi, I've been running MITgcm on gadi with payu. [This repo](https://github.com/claireyung/mitgcm-pig) contains my model config directory. As mentioned in [this post](https://forum.access-hive.org.au/t/payu-and-mitgcm-crashes-with-restart-history-counting-error/1884), the model runs with the gadi analysis3-23.07 conda environment but not analysis3-23.10: in that case I get the error message in https://github.com/claireyung/mitgcm-pig/blob/main/mitgcmice2.e110118980, relating to restart date counting, copied below:
```
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-23.10/bin/payu-run", line 10, in <module>
    sys.exit(runscript())
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-23.10/lib/python3.10/site-packages/payu/subcommands/run_cmd.py", line 133, in runscript
    expt.setup()
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-23.10/lib/python3.10/site-packages/payu/experiment.py", line 454, in setup
    self.get_restarts_to_prune()
  File "/g/data/hh5/public/apps/miniconda3/envs/analysis3-23.10/lib/python3.10/site-packages/payu/experiment.py", line 1056, in get_restarts_to_prune
    index_bound = max_index - restart_history
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

 This was surprising to me because I am not using restarts, since startTime in [my data input file ](https://github.com/claireyung/mitgcm-pig/blob/main/data) (see &PARM03 for time stepping parameters) is zero.

jo-basevi:
There's a bug in the `get_restarts_to_prune()` method which raises an error when `restart_history` is configured and an archive directory exists but it contains no restarts. 

I've added a test case that was able to recreate the error.

I could also do an early return if there is no restarts, but I've left the check for it later in the method to after checks for valid `config.yaml` values. But it would be an easy change to add that. 

Closes #422

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -498,7 +498,6 @@ class Experiment(object):
             else:
                 mpi_flags = []
 
-
         if not isinstance(mpi_flags, list):
             mpi_flags = [mpi_flags]
 
@@ -968,9 +967,12 @@ class Experiment(object):
         if not os.path.exists(self.archive_path):
             return []
 
-        # List all restart directories in archive
+        # Sorted list of restart directories in archive
         restarts = list_archive_dirs(archive_path=self.archive_path,
                                      dir_type='restart')
+        restart_indices = {}
+        for restart in restarts:
+            restart_indices[restart] = int(restart.lstrip('restart'))
 
         # TODO: Previous logic was to prune all restarts if self.repeat_run
         # Still need to figure out what should happen in this case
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM. Thanks.

PR #425, opened 2024-03-15
Add check for UUIDs when checking for existence of archives

### LINKED ISSUE #420: Existing experiment is not detected if archive does not exist
If an existing experiment does not have an archive path but a legacy archive path exists the current logic will assume it is a legacy experiment.

I discovered this when I made a new experiment, incorrectly configured it when I ran, so did a `payu sweep --hard` to remove the unwanted outputs and start again.

For the logic contained here:

https://github.com/payu-org/payu/blob/b6db50b8502ebc7cfbaa9bc12d8920517f332ac7/payu/metadata.py#L178-L196

```python
is_new_experiment == False
archive_path.exists() == False
legacy_archive_path.exists() == True
```
So the branch at https://github.com/payu-org/payu/blob/b6db50b8502ebc7cfbaa9bc12d8920517f332ac7/payu/metadata.py#L181 is the one taken.

As soon as I created `archive_path` for my new experiment `payu` takes the branch above and it works.

jo-basevi:
In #420, there’s a bug with using branching logic, where an existing experiment is not detected if its archive does not exist. So if an experiment was created then deleted on a branch in a control directory that has a corresponding legacy archive,  subsequent `payu setup/run` calls, it’ll use the legacy archive. 

This PR so far adds an additional check when checking for the existence of archives, also check for metadata with a matching UUID. 

This isn’t a perfect check as it requires `payu setup/payu run/ payu checkout` to be run on the legacy experiment which will then subsequently create a UUID and metadata in the legacy archive if it doesn’t already exist. Otherwise, if there is no metadata associated with the legacy archive and a new UUID has not been generated, it’s difficult to know whether to use a legacy archive or not.

Closes #420

TODO:
    - Write up a short explainer for "recommendations for use with legacy experiments"

### DIFF
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -155,7 +155,7 @@ class Metadata:
                   self.experiment_name)
             return
 
-        # Legacy experiment name and archive path
+        # Legacy experiment name
         legacy_name = self.control_path.name
 
         if not self.enabled:
@@ -167,7 +167,6 @@ class Metadata:
             return
 
         branch_uuid_experiment_name = self.new_experiment_name()
-
         if is_new_experiment or self.has_archive(branch_uuid_experiment_name):
             # Use branch-UUID aware experiment name
             self.experiment_name = branch_uuid_experiment_name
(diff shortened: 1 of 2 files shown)


jo-basevi:
I've simplified the logic in the check to only fail when there exists a UUID in the archive metadata but it is not equal to the UUID in the control directory metadata. 

So it'll still work for legacy archives with no UUIDs or metadata. So the recommendation when running a legacy experiment but want to run a related experiment on a different branch in the same control directory is to firstly run: `payu setup` on the legacy experiment branch to ensure a UUID is generated, and metadata is copied across to the legacy archive.

aidanheerdegen:
Doesn't this interact rather heavily with https://github.com/payu-org/payu/pull/427? I'm seeing code I think was modified in the other PR.

jo-basevi:
Yes, I'll need rebase the changes onto main and then run a couple tests

aidanheerdegen (approved):
LGTM. All much cleaner and easier to follow.

PR #427, opened 2024-03-15
Add fix and test for metadata enable false bug

### LINKED ISSUE #426: Metadata/UUIDs disable config option bug
There’s a bug with configuring `metadata: enable: false`. This was added late on in the metadata/UUID changes. The idea behind this configuration option is that it would prevent the creation of new metadata files and UUIDs, and experiment name would either be:
 - The configured `experiment` value in `config.yaml`
- Or, `$CONTROL_DIR-$BRANCH` or just `$CONTROL_DIR` if not a git repo or on main/master branch.

The bug raises an `AttributeError:AttributeError: 'Metadata' object has no attribute 'repo’` when `enable: false`. 

Going through the metadata code again, it got me thinking more that this option doesn’t check for legacy archives. It is documented as `If set to False, the UUID is left out of the experiment name used for archival.`

I wonder whether the logic should be more consistent with the experiment naming with UUIDs and check for existence of archives. The idea was that disabling metadata generation could be useful for development as branches can still be used but it keeps metadata/uuids out of the git commits.

Or, should `metadata: enable: false` only use control directory names for experiment names to make the logic more simple. Another pro of using “legacy” names is if `metadata: enable` is set to `true` or removed, payu can still find the existing archive. Otherwise the logic probably should be extended to search for `$CONTROL_DIR-$BRANCH` archives.

jo-basevi:
This PR fixes the config option to disable metadata bug #426.
I’ve added a test that replicates the error. 

Closes #426

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -489,8 +489,8 @@ Experiment Tracking
 
    ``enable`` (*Default:* ``True``)
       Flag to enable/disable creating/updating metadata files and UUIDs.
-      If set to False, the UUID is left out of the experiment name used
-      for archival.
+      If set to False, the experiment name used for archival is either the
+      control directory name or the configured ``experiment`` name.
 
    ``model`` (*Default: The configured model value*)
       Model name used when generating metadata for new experiments.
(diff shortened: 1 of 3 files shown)


jo-basevi:
As discussed, with @aidanheerdegen, have modified the the `config.yaml`, `metadata: enable: false` to set the experiment name for archival to either to the legacy control directory name or the configured `experiment` value in `config.yaml`. So payu will effectively behave as it did before UUID+Branching changes, i.e. no branching and no updating/generating metadata with UUIDs.

PR #432, opened 2024-03-27
Allow for multiple restarts from each component in `cesm_cmeps` driver

### LINKED ISSUE #430: Incorrect Restart File Names for MOM6 in payu 1.1
In the new PayU version 1.1, it has been observed that the restart file names for the MOM6 are incorrect. This issue causes MOM6 to look for files with incorrect filenames, leading to warnings such as:
```
WARNING from PE 0: MOM_restart: Unable to find restart file : ./GMOM_JRA.mom6.r.1900-01-02-00000_1.nc.nc
WARNING from PE 0: MOM_restart: Unable to find restart file : ./GMOM_JRA.mom6.r.1900-01-02-00000_2.nc.nc
WARNING from PE 0: MOM_restart: Unable to find restart file : ./GMOM_JRA.mom6.r.1900-01-02-00000_3.nc.nc
WARNING from PE 0: MOM_restart: Unable to find restart file : ./GMOM_JRA.mom6.r.1900-01-02-00000_4.nc.nc
```
As seen in the warning messages, the file extension .nc.nc is incorrect and seems to be duplicated, resulting in MOM6 being unable to locate the required restart files.

dougiesquire:
MOM6 may split its restart across multiple files. This wasn't accounted for in the `cesm_cmeps` driver, but with this PR it is.

Closes #430

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -217,12 +217,12 @@ class CesmCmeps(Model):
         for pointer in pointer_files:
             try:
                 with open(pointer,"r") as f:
-                    restart = f.readline().rstrip()
-                    restart_files.append(
-                        os.path.join(
-                            self.work_path, restart,
+                    for restart in f.readlines():
+                        restart_files.append(
+                            os.path.join(
+                                self.work_path, restart.rstrip(),
+                                )
                             )
-                        )
             except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
Looks good. Normally at this point I would suggest that tests are almost always a good idea, but I understand the issues with creating tests with `payu`.

Issue #429, opened 2024-03-18
Updated metadata schema URL


aidanheerdegen:
The location of the metadata schema is changing

https://github.com/ACCESS-NRI/schema/pull/12

This will require an update to the URL in `metadata.py`

https://github.com/payu-org/payu/blob/421431b52c24dd57fa0ad023d46b739667133af7/payu/metadata.py#L39

We will need flag that this is being deprecated and when that file is removed it will break existing deployed versions of `payu`.

Note for the future, if we had used a specific commit then there would be no problem with deleting/moving the file in the schema repo, i.e.

https://raw.githubusercontent.com/ACCESS-NRI/schema/36abe2fe28eb2853a54f41c5eedfd964617d9d68/experiment_asset.json

rather than 

https://raw.githubusercontent.com/ACCESS-NRI/schema/main/experiment_asset.json

aidanheerdegen:
Metadata URL was updated in #434 but we should keep this open until we deprecate deployed versions of  `payu` with old URL location.

AFAICT this only affects a single deployed version:
```
/g/data/vk83/apps/payu/1.1
```

So once we have a couple of versions deployed over the top it should be fine to remove the deprecated schema file and close this issue.

aidanheerdegen:
Should consider deprecating payu/1.1 and doing what is necessary to close this issue.

jo-basevi:
Closing this as `payu/1.1` module has already been removed.

PR #434, opened 2024-04-05
Update metadata schema url

### LINKED ISSUE #424: `model` field in `metadata.yaml` should be an array
The `model` field in the `metadata.yaml` file should be an array to allow for experiments that contain multiple models. Currently Payu [writes a string](https://github.com/payu-org/payu/blob/421431b52c24dd57fa0ad023d46b739667133af7/payu/metadata.py#L273), which means that the auto-generated `metadata.yaml` files don't validate against the [`experiment_asset.json` schema](https://github.com/ACCESS-NRI/schema/blob/main/experiment_asset.json#L22-L25).

It could possibly be worth automatically checking somewhere that Payu is generating valid `metadata.yaml` files.

dougiesquire:
This PR updates the metadata schema used by `metadata.py` to [experiment-metadata/1-0-3.json](https://github.com/ACCESS-NRI/schema/tree/main/au.org.access-nri/model/output/experiment-metadata/1-0-3.json)

Closes #424. Related #429

### DIFF
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -36,8 +36,7 @@ MODEL_FIELD = "model"
 METADATA_FILENAME = "metadata.yaml"
 
 # Metadata Schema
-SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/main/experiment_asset.json"
-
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM. Thanks @dougiesquire

PR #439, opened 2024-05-03
Get module executables from path loaded by environment modules

### LINKED ISSUE #379: Get model executables from path
Recently we added the option to install ACCESS-OM3 with Spack.  Usually the location where Spack installs the packages includes a hash in the naming scheme, which makes it a bit tedious to keep the full executable path up to date in the payu `config.yaml` file.  On the other hand, we are using Spack to automatically create environment modules for OM3 that add the corresponding path to the OM3 binaries to the PATH environment variable. Therefore it would be nice if payu could get the executable from the path.

Basically, instead of doing:

```yaml
exe: /g/data/ik11/spack/0.20.1/opt/linux-rocky8-cascadelake/intel-2021.6.0/access-om3-main-7cryrsy/bin/access-om3-MOM6-CICE6

modules:
  use:
      - /g/data/ik11/spack/0.20.1/modules/access-om3/0.3.0/linux-rocky8-cascadelake
  load:
      - access-om3
```

we would like to simply write:

```yaml
exe: access-om3-MOM6-CICE6

modules:
  use:
      - /g/data/ik11/spack/0.20.1/modules/access-om3/0.3.0/linux-rocky8-cascadelake
  load:
      - access-om3
```

jo-basevi:
- Move setting executable paths in Model class to Experiment.setup()
- Inspect user modules (specified in config.yaml) to changes to PATH - The modules are restricted to user module directories defined in config.yaml
- If exe is not an absolute path, search along above paths

Modules are still loaded the same way in experiment.run so to not change users env vars when running `payu setup` 

Closes #379

### DIFF
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -33,6 +33,7 @@ commands, e.g.:
     $ module unuse dir # Remove dir from $MODULEPATH
 """
 
+
 def setup(basepath=DEFAULT_BASEPATH):
     """Set the environment modules used by the Environment Module system."""
 
@@ -153,7 +154,7 @@ def setup_user_modules(user_modules, user_modulepaths):
 
         # Extract out the modulefiles available
         modules = [line for line in output.strip().splitlines()
-                    if not (line.startswith('/') and line.endswith(':'))]
+                   if not (line.startswith('/') and line.endswith(':'))]
 
         if len(modules) > 1:
             # Modules are used for finding model executable paths - so check
@@ -176,7 +177,7 @@ def env_var_set_by_modules(user_modules, env_var):
     if 'MODULESHOME' not in os.environ:
         print('payu: warning: No Environment Modules found; skipping '
               f'inspecting user module changes to ${env_var}')
-        return []
+        return
 
     # Note: Using subprocess shell to isolate changes to environment
     load_commands = [f'load {module}' for module in user_modules]
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -17,6 +17,21 @@ if not hasattr(subprocess, 'check_output'):
 DEFAULT_BASEPATH = '/opt/Modules'
 DEFAULT_VERSION = 'v4.3.0'
 
+MODULE_NOT_FOUND_HELP = """ To fix module not being found:
+- Check module name and version in config.yaml (listed under `modules: load:`)
+- If module is found in a module directory, ensure this path is listed in
+config.yaml under `modules: use:`, or run `module use` command prior to running
+payu commands.
+"""
+
+MULTIPLE_MODULES_HELP = """ To fix having multiple modules available:
+- Add version to the module in config.yaml (under `modules: load:`)
+- Modify module directories in config.yaml (under `modules: use:`)
+- Or modify module directories in user environment by using module use/unuse
+commands, e.g.:
+    $ module use dir # Add dir to $MODULEPATH
+    $ module unuse dir # Remove dir from $MODULEPATH
+"""
 
 def setup(basepath=DEFAULT_BASEPATH):
     """Set the environment modules used by the Environment Module system."""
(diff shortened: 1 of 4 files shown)


payu/models/fms.py
```diff
@@ -76,9 +76,16 @@ def fms_collate(model):
...
     if mpi:
-        # Must use envmod to be able to load mpi modules for collation
-        envmod.setup()
-        model.expt.load_modules()
+        # TODO: I've added module setup and load functions to
+        # Experiment.collate - as user modules needed to be use & loaded
+        # to expand executable paths. However, is loading mpi module
+        # for non-mpi collate jobs going to cause errors?
+        # If so, need to only use/load user modules in Experiment.collate
+        # and only load additional modules for mpi here
```
aidanheerdegen, line 84: I don't think it is a problem to load if not required, but best if we don't have to.

payu/envmod.py
```diff
@@ -109,4 +109,74 @@ def lib_update(required_libs, lib_name):
...
+    # Remove inital paths and convert into a set
+    return set(paths).difference(set(init_paths))
+
+
+def paths_post_module_commands(commands):
+    """Runs subprocess module command and parse out the resulting
+    PATH environment variable"""
+    # Use modulecmd as module command is not available on compute nodes
+    module_cmds = [
+        f"eval `{os.environ['MODULESHOME']}/bin/modulecmd bash {c}`"
```
aidanheerdegen, line 166: Currently wondering about security of doing `eval` on an environment variable.

aidanheerdegen (commented):
Some comments, but looks great, thanks.

jo-basevi:
Hi Aidan, thanks for the review! I've changed the logic a little bit to actually load user modules as part of `payu setup` rather than `payu run`(other modules will still get loaded in `payu run`).

 The idea was to actually check if the module will be found, and check if more than one module is available. It errors out if more than one module is available - this may be a breaking change for configs thats haven't added a version for modules? I am not sure if there exists a case when the same module and version exists in two modulepaths, but still want one of them to be loaded? If so, can change it to an warning.. 

I wanted to also load the actual modules that will be loaded as part of `payu run`. There could be a confusing case where one modulefile was loaded because `module use` was used, but that was different to the module loaded only using `module: use:` in config.yaml`. Hopefully this case be caught with the above check.

payu/models/model.py
```diff
@@ -198,6 +177,54 @@ def get_prior_restart_files(self):
...
+            module_added_paths = self.expt.user_modules_set_paths
+
+        # Search for exe inside paths added to $PATH by user-defined modules
+        exec_paths = []
+        for path in module_added_paths:
+            exec_path = os.path.join(path, exec)
+            if os.path.exists(exec_path):
+                exec_paths.append(exec_path)
+
+        if len(exec_paths) > 1:
```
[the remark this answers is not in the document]
jo-basevi: Yeah I wanted to double check incase two modules with the same model executables name got loaded

payu/models/model.py
```diff
@@ -198,6 +177,54 @@ def get_prior_restart_files(self):
...
+        if not search_module_path:
+            module_added_paths = set()
+        else:
+            module_added_paths = self.expt.user_modules_set_paths
+
+        # Search for exe inside paths added to $PATH by user-defined modules
+        exec_paths = []
+        for path in module_added_paths:
+            exec_path = os.path.join(path, exec)
+            if os.path.exists(exec_path):
```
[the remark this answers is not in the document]
jo-basevi: Thanks, have added that check

payu/models/model.py
```diff
@@ -339,6 +366,8 @@ def collate(self):
         raise NotImplementedError
 
     def build_model(self):
+        # Don't search user modules for executable paths
+        self.setup_executable_paths(search_module_paths=False)
```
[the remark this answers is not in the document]
jo-basevi: I was just adding that there so exec_path values were defined, otherwise it would break the function (because exec paths are no longer set in `Model.init()`.

But if the intention is to remove it, that's great as I don't have to worry about figuring out how to test it

payu/models/fms.py
jo-basevi: I've split off loading the user modules from loading the other modules - so user modules will always get loaded as part of collate and `load_modules` - which loads mpi is only loaded if needed.

payu/envmod.py
jo-basevi: Yeah.. Would it be better to run it in python? e.g.: https://github.com/payu-org/payu/blob/91f2831a55b61d20a0da3d5f933b6b269e0fb875/payu/envmod.py#L86-L92

payu/envmod.py
```diff
@@ -109,4 +125,79 @@ def lib_update(required_libs, lib_name):
...
+        module('use', modulepath)
+
+    for modulefile in user_modules:
+        # Check module exists and there is not multiple available
+        module_subcommand = f"avail --terse {modulefile}"
+        output = run_cmd(module_cmd(module_subcommand)).stderr
+
+        # Extract out the modulefiles available
+        modules = [line for line in output.strip().splitlines()
+                   if not (line.startswith('/') and line.endswith(':'))]
```
aidanheerdegen, line 157: This is stripping out lines like `/apps/Modules/modulefiles:` indicating the path to the listed modulefiles. Might be worth stating this explicitly.

payu/envmod.py
```diff
@@ -109,4 +125,79 @@ def lib_update(required_libs, lib_name):
...
+
+        if len(modules) > 1:
+            # Modules are used for finding model executable paths - so check
+            # for unique module -TODO: Could be a warning rather than an error?
```
aidanheerdegen, line 161: Yeah there will be multiple results if there are multiple versions of a module. For non-model modules that is probably fine. For models it's probably not desirable, e.g. a user specifying `access-om2` rather than a specific version `access-om2/2024.03.0`.

Not sure what to do about that. Probably best to insist on users specifying a version at this stage, so leave as error but the error message should tell them what to do: namely use a specific version.

payu/models/model.py
```diff
@@ -198,6 +177,59 @@ def get_prior_restart_files(self):
...
+
+        # Check if path set by loading user modules has been defined
+        module_added_path = self.expt.user_modules_path
+        if module_added_path is None:
+            print("payu: warning: Skipping searching for model executable " +
+                  "in $PATH set by user modules")
+            module_added_paths = []
+        elif module_added_path == '':
+            module_added_paths = []
+        else:
```
aidanheerdegen, line 194: I'm not clear on the difference between these two cases, namely `None` or `''`. Don't they both indicate search will be skipped?

aidanheerdegen: In both cases we're going to fall through to the `return` condition right?

payu/models/model.py
```diff
@@ -198,6 +177,59 @@ def get_prior_restart_files(self):
...
+                f"user-defined modules in `config.yaml`. Paths: {exec_paths}")
+        elif len(exec_paths) == 1:
+            return exec_paths[0]
+
+        # Else prepend the lab bin path to exec
+        return os.path.join(self.expt.lab.bin_path, exec)
+
+    def setup_executable_paths(self):
+        """Set model executable paths"""
+        self.exec_prefix = self.config.get('exe_prefix', '')
```
aidanheerdegen, line 216: I don't really understand `exec_prefix`. I don't recall it ever being used in a configuration, and it is just randomly added in front of the `exe` path in `experiment.run` so it feels orthogonal to this new functionality. Maybe I'm misunderstanding it's purpose.

payu/envmod.py
```diff
@@ -109,4 +125,79 @@ def lib_update(required_libs, lib_name):
...
+              f'inspecting user module changes to ${env_var}')
+        return
+
+    # Note: Using subprocess shell to isolate changes to environment
+    load_commands = [f'load {module}' for module in user_modules]
+    commands = ['purge'] + load_commands
+    module_cmds = [f"eval `{module_cmd(c)}`" for c in commands]
+    module_cmds += [f'echo ${env_var}']
+    command = ' && '.join(module_cmds)
+    output = run_cmd(command)
```
aidanheerdegen, line 188: If this was a separate function it would be easily testable by checking the return string matched expectations.

That test might be quite uninformative though. I'll leave it up to you if you think it is worth it.

aidanheerdegen (approved):
Looks good, but as usual questions questions questions ...

payu/models/fms.py
```diff
@@ -92,8 +91,7 @@ def fms_collate(model):
                 mppnc_path = os.path.join(model.expt.lab.bin_path, f)
                 break
     else:
-        if not os.path.isabs(mppnc_path):
-            mppnc_path = os.path.join(model.expt.lab.bin_path, mppnc_path)
+        mppnc_path = model.expand_executable_path(mppnc_path)
```
aidanheerdegen, line 94: This is a random comment, but was chatting to @dougiesquire and @ezhilsabareesh8 today about their requirements for wave-watch three, and one of these was that there was an input file (a sort of derived config file) that must be generated every time the model is run, and it is a separate executable from the model executable, but should be built from the same source version as the model exe. So it was necessary to have this available in the `work` directory. So it would be a kind of generalisation of this, looking for tools in an executable path. 

I think it would have to be some new functionality, but basically just listing additional binary tools for a model and resolving their location and adding a link to it in `work`

payu/envmod.py
aidanheerdegen: On second thoughts, it would be great if we didn't have to do this at all.

Is there a reason we can't just save the `env` state before loading the user modules, and then comparing the `env` after we load the modules and see what has changed in `$PATH`?

payu/models/model.py
jo-basevi: This is the commit that it was added in: https://github.com/payu-org/payu/commit/8df851bb845147f8ac35c0071a596774a63e0bd6. I left it in as I don't know if it is still useful for e.g. gdb, gprof and valgrind?

payu/envmod.py
jo-basevi: Yeah thats a good point. I had previously kept them separate as the user modules were loaded in `experiment.run`. But as they are being loaded in setup and `module purge` isn't needed, I changed it to inspect the `$PATH` and `$LOADEDMODULES` in the same function that loads the user modules. I've added in code that will `unload` the module is it was already loaded before `payu setup`, so it can pick up any changes to those environment variables

payu/models/model.py
jo-basevi: As experiment now has a field `self.user_modules_paths` which is a set or None, I've updated to just check if it `None` with the error message. If loading user modules, and there were no changes to $PATH, `self.user_modules_paths` will just be an empty set()

payu/envmod.py
jo-basevi: Thanks, I've added that into a comment

payu/models/model.py
aidanheerdegen: I guess so. This is a pretty specialist requirement, so we'll assume whomever is using it knows what they're doing and we don't have to hold their hand, so leave as-is.

payu/models/model.py
```diff
@@ -184,15 +184,11 @@ def expand_executable_path(self, exec):
             return exec
 
         # Check if path set by loading user modules has been defined
-        module_added_path = self.expt.user_modules_path
-        if module_added_path is None:
+        module_added_paths = self.expt.user_modules_paths
+        if module_added_paths is None:
```
aidanheerdegen, line 188: Should `self.expt.user_modules_paths` just be initialised to an empty list? Or do is it important to be `None` at some stage?

In which case the `module_added_paths` temporary variable wouldn't be required I guess.

aidanheerdegen (commented):
Were you planning on adding tests?

aidanheerdegen (approved):
Ok, tests are probably too complicated for the return, so LGTM!

payu/models/model.py
jo-basevi: `None` just indicates that user modules were not loaded (either `expt.setup_modules()` has not been called yet or `$MODULESHOME` is not defined). If `self.expt.user_modules_paths` is used elsewhere for other things, it might be useful for it to stay None.

But it also could simplify the code a little if it was just initialised to `[]`...

payu/envmod.py
jo-basevi: Yeah, I've added a longer log message about using specific version, or change what modules directories are used. So will leave as an error for now

Issue #440, opened 2024-05-06
Enhance Workflow Flexibility in WW3 by Implementing Dynamic Generation of Binary Files


ezhilsabareesh8:
**Description:**
The WW3 model's workflow is intricate, involving the generation of binary files through the `ww3_grid` grid preprocessor. Currently, these binary files are created using predefined namelist parameters in `ww3_grid.nml` and `namelists_Global.nml`, alongside specific grid files (`OM2_1.Lon`, `OM2_1.Lat`, `OM2_1.Dpt`, `OM2_1.Mask`, and `OM2_1.Obstr`).

<img width="772" alt="Screenshot 2024-05-06 at 12 29 44 pm" src="https://github.com/payu-org/payu/assets/108497463/1f6c5de9-69c2-468e-a2e3-403a360d94d6">

To improve user experience and workflow flexibility, we propose implementing a solution where the binary files (specifically `mod_def.ww3`) are dynamically generated during runtime. This approach would enable users to modify most input namelist parameters without needing to rebuild `ww3_grid`, thus providing greater control and understanding of the binary file's provenance.

**Proposed Solution:**
1. Integrate dynamic generation of binary files (`mod_def.ww3`, `restart.ww3`) within the WW3 workflow using Payu.
2. Allow users to modify input namelist parameters without requiring a rebuild of `ww3_grid` or `ww3_strt` (for generating restart), except for major changes that necessitate modifying switches.

PS: The challenge arises when users need to modify critical parameters such as propagation schemes, source schemes, attenuation and scattering models. While most changes can be accommodated within the namelists, certain adjustments necessitate modifying switches and rebuilding `ww3_grid` other wise modifications in the namelists will be ignored by WW3.

aekiss:
Should we also look at the approach [NorESM](https://github.com/NorESMhub/NorESM) are taking to generate `mod_def.ww3` at runtime, as mentioned by @mvertens last week?

aekiss:
> certain adjustments necessitate modifying switches and rebuilding `ww3_grid` other wise modifications in the namelists will be ignored by WW3.

Is there an option to make WW3 abort if it receives an unrecognised namelist parameter, as we do in MOM6?

ezhilsabareesh8:
> Is there an option to make WW3 abort if it receives an unrecognised namelist parameter, as we do in MOM6?

[...]

PR #442, opened 2024-05-16
Generate mod_def.ww3 input file on the go

ezhilsabareesh8:
This PR addresses part of issue #440 by implementing a feature to generate the mod_def.ww3 file on the go.

- Added functionality to generate `mod_def.ww3` on the go.
- Updated the list of configuration files for WW3.


anton-seaice:
What's the status of this @ezhilsabareesh8 ?

ezhilsabareesh8:
Thanks @anton-seaice. This PR became stale after the original review requests and I did not get follow-up feedback, so it has not moved forward since then.

Since it is now far behind `master`, I think the best option is to first confirm whether this functionality is still wanted. If it is, I can rebase the branch onto the current master and update it as needed, or open a fresh PR based.

anton-seaice:
Thanks @ezhilsabareesh8 - ill give it a first review and then we can sync it up

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +177,35 @@ def setup(self):
...
+            
+             # Change the current working directory to f_dst
+             os.chdir(f_dst)
+
+             # Extract the directory name from self.exec_path without the base name
+             exec_dir = os.path.dirname(self.exec_path)
+
+             # Run the ww3_grid command using the defined path
+             cmd = os.path.join(exec_dir, "ww3_grid")
+             result = os.system(cmd)
```
anton-seaice: I think you need to check here that ww3_grid exists, and return an error otherwise

And check that result is successful, and return an error otherwise

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +177,35 @@ def setup(self):
...
+             os.chdir(original_dir)
+
+             # Check if ww3_grid command executed successfully
+             if result == 0:
+                print("mod_def.ww3 generated successfully.")
+             else:
+                print("Error generating mod_def.ww3 file.")
+                raise SystemExit(1)  # Terminate the program with exit code 1
+        else:
+                print("ww3dev component not found in self.components.")
```
anton-seaice: We don't want this print statement, most of the time ww3 isn't present

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +177,35 @@ def setup(self):
...
+
+             # Run the ww3_grid command using the defined path
+             cmd = os.path.join(exec_dir, "ww3_grid")
+             result = os.system(cmd)
+
+             # Change back to the original directory
+             os.chdir(original_dir)
+
+             # Check if ww3_grid command executed successfully
+             if result == 0:
```
anton-seaice: This check should be at line 195 - it would be easier to understand there

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +177,35 @@ def setup(self):
...
+                print("mod_def.ww3 generated successfully.")
+             else:
+                print("Error generating mod_def.ww3 file.")
+                raise SystemExit(1)  # Terminate the program with exit code 1
```
anton-seaice: raise and Error with a message would be better

e.g. 

`raise RuntimeError("Failed to generate WW3 input files")`

?

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +177,35 @@ def setup(self):
...
+        else:
+                print("ww3dev component not found in self.components.")
```
anton-seaice, line 199: I think you should commit the new mod_def.ww3 file at this point

anton-seaice (commented):
Some general comments @ezhilsabareesh8 

Is is slow to run this everytime payu setup is run ?

payu/models/cesm_cmeps.py
ezhilsabareesh8: Updated in this commit dcd7f66 , to validate that `ww3_grid` exists and is executable before running it, and to raise a error if the command fails or does not produce `mod_def.ww3`.

payu/models/cesm_cmeps.py
ezhilsabareesh8: Fixed in this commit dcd7f66 , WW3 setup now only runs when `ww3dev` is actually present

payu/models/cesm_cmeps.py
ezhilsabareesh8: Fixed in this commit https://github.com/payu-org/payu/commit/dcd7f663c71444e7b428986d0a75ea571200cb9e, added [here](https://github.com/ACCESS-NRI/payu/blob/dcd7f663c71444e7b428986d0a75ea571200cb9e/payu/models/cesm_cmeps.py#L268-L269)

payu/models/cesm_cmeps.py
ezhilsabareesh8: Fixed in this commit https://github.com/payu-org/payu/commit/dcd7f663c71444e7b428986d0a75ea571200cb9e, see [here](https://github.com/ACCESS-NRI/payu/blob/dcd7f663c71444e7b428986d0a75ea571200cb9e/payu/models/cesm_cmeps.py#L269-L270)

payu/models/cesm_cmeps.py
ezhilsabareesh8: Could you clarify where you want mod_def.ww3 committed? It’s a binary file.

anton-seaice: Oh sorry - I forgot it's treated as an input.

How long does it take to generate ?

ezhilsabareesh8: It will take just a few seconds to generate for `100km` grid

payu/models/cesm_cmeps.py
```diff
@@ -169,19 +179,9 @@ def setup(self):
...
         self.runconfig.write()
 
-        # Horrible hack to make a link to the mod_def.ww3 input in the work
-        # directory
-        # The ww3 mod_def input needs to be in work_path and called mod_def.ww3
         if "ww3dev" in self.components.values():
-            f_name = "mod_def.ww3"
-            f_src = os.path.join(self.work_input_path, f_name)
-            f_dst = os.path.join(self.work_path, f_name)
+            self._setup_ww3_mod_def()
```
anton-seaice, line 198:
```suggestion
            # WW3 uses a bespoke binary file for configuration (mod_def.ww3),
            # Generate mod_def.ww3 on the fly based on the WW3 files in the config being run
            self._setup_ww3_mod_def()
```

?

anton-seaice (commented):
This looks ok to me - be good to get someone from the release team to look over it

PR #445, opened 2024-06-12
Save experiment outputs + counters as environment variables

### LINKED ISSUE #443: Providing the experiment counter as an environment variable
I've been investigating ways to run a UM file to netcdf converter as a payu userscript for the ESM1.5 release. To convert the files in the correct output directory (e.g. archive/output005), it would be helpful to access either the experiment's counter, or the current output directory held in `expt.counter` and `expt.dir_path` respectively. From what I can understand, neither of these are currently easily accessible by a userscript. 

Would it be reasonable to export these as environment variables, perhaps after  `self.set_counters()` and `self.set_output_paths()` in experiment.py? It would then be easy to provide these to the conversion job during its submission.

jo-basevi:
Add environment variables of experiment output paths, and run counters so they can be easily accessed via user-scripts.

Currently have added `PAYU_CURRENT_RUN`, `PAYU_ARCHIVE_DIR`, `PAYU_CURRENT_OUTPUT_DIR` and `PAYU_CURRENT_RESTART_DIR`.

Closes #443

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -98,6 +98,7 @@ class Experiment(object):
             model.set_input_paths()
 
         self.set_output_paths()
+        self.set_environment_vars()
 
         # Create metadata file and move to archive
         self.metadata.write_metadata(restart_path=self.prior_restart_path)
@@ -374,6 +375,18 @@ class Experiment(object):
 
         for model in self.models:
             model.set_model_output_paths()
+
+    def set_environment_vars(self):
+        """Save information of output directories and current run to
+        environment variables, so they can be accessed via user-scripts"""
+        os.environ.update(
+            {
+                'PAYU_CURRENT_OUTPUT_DIR': self.output_path,
+                'PAYU_CURRENT_RESTART_DIR': self.restart_path,
+                'PAYU_ARCHIVE_DIR': self.archive_path,
+                'PAYU_CURRENT_RUN': str(self.counter)
+            }
+        )
 
     def build_model(self):
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -98,7 +98,6 @@ class Experiment(object):
             model.set_input_paths()
 
         self.set_output_paths()
-        self.set_environment_vars()
 
         # Create metadata file and move to archive
         self.metadata.write_metadata(restart_path=self.prior_restart_path)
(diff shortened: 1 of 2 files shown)


payu/experiment.py
```diff
@@ -375,6 +376,18 @@ def set_output_paths(self):
         for model in self.models:
             model.set_model_output_paths()
 
+    def set_environment_vars(self):
```
blimlim, line 379: While trying to understand some parts of payu I saw there's an existing function `set_env_vars` in `cli.py` with a similar name. I'm wondering whether we should use a different name here to avoid the two getting mixed up. If the exported variables will only be used by userscripts, would something like `set_userscript_env_vars` be ok?

payu/subcommands/run_cmd.py
```diff
@@ -147,6 +147,7 @@ def runscript():
             expt.manifest = Manifest(expt.config.get('manifest', {}),
                                      reproduce=False)
             expt.set_output_paths()
+            expt.set_environment_vars()
```
blimlim, line 150: Just a small idea: wondering whether it would work if the new function was called within `run_userscript` from `experiment.py`, e.g:

```python
def run_userscript(self, script_cmd):
        self.set_environment_vars()

        # First try to interpret the argument as a full command:
        try:
            sp.check_call(shlex.split(script_cmd))
```
I was thinking this could make it so that the variables are only exported in situations when they might be needed and it might make the purpose of the new function a bit clearer when skimming through the code.

blimlim (commented):
Thanks for putting these changes together – I think they should work well. I just had a couple of small ideas, though also happy to approve if you think it's better to keep as is!

payu/experiment.py
jo-basevi: Yeah I think that's a great idea - I've added that in that change

payu/subcommands/run_cmd.py
jo-basevi: Yeah that makes more sense for  it to be run there and ensures the env vars have been set before any user-script is run. I've now added that in

jo-basevi:
Thanks for the review @blimlim! I've added in the code suggestions, so it's good for another review

blimlim (approved):
Thanks for adding the changes, looks great!

blimlim (approved):

CodeGat (approved):

PR #447, opened 2024-06-24
Add cmd-line flag to disable metadata + UUID generation and commits

### LINKED ISSUE #446: Command line argument to disable metadata/UUID generation
Currently the only way to disable new UUIDs being generated and metadata files being committed, is to disable it in `config.yaml`, e.g.
```
metadata:
   enable: false
```

(Note: If set to False, the experiment name used for archival is either the control directory name or the configured experiment name - so basically how payu worked before UUID/branching changes)

For developing released configurations, it would be useful to easily disable metadata when running payu commands using a command line argument.

jo-basevi:
Add cmd line flag `--metadata-off`/`-M` (better names?) to payu subcommands to disable metadata files or UUIDs being created. This uses control directory for experiment names. (so equivalent to `metadata: enable: false` in `config.yaml`)

Added this flag to sub-commands that create an Experiment class instance - so archive, collate, profile, run, setup, sweep and sync. 

Closes #446

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -49,7 +49,7 @@ default_restart_freq = 5
 
 class Experiment(object):
 
-    def __init__(self, lab, reproduce=False, force=False):
+    def __init__(self, lab, reproduce=False, force=False, metadata_off=False):
         self.lab = lab
 
         if not force:
@@ -61,7 +61,7 @@ class Experiment(object):
         self.start_time = datetime.datetime.now()
 
         # Initialise experiment metadata - uuid and experiment name
-        self.metadata = Metadata(Path(lab.archive_path))
+        self.metadata = Metadata(Path(lab.archive_path), disabled=metadata_off)
         self.metadata.setup()
 
         # TODO: replace with dict, check versions via key-value pairs
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -58,13 +58,18 @@ class Metadata:
         config_path : Optional[Path]
             Configuration Path. The default is config.yaml in the current
             working directory. This is also set in fsop.read_config
+        disabled : bool, default False
+            Flag to disable metadata and UUID generation and commits. The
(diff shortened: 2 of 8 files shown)


payu/experiment.py
```diff
@@ -794,6 +794,10 @@ def archive(self, force_prune_restarts=False):
                 payu=self.payu_path,
                 expt=self.counter
             )
+
+            if not self.metadata.enabled:
+                cmd += f' --metadata-off'
+
```
aidanheerdegen, line 800: Would it be possible, and therefore easier, to propagate this option by setting the `PAYU_METADATA_OFF` environment variable?

aidanheerdegen (approved):
Wow that touched a lot! I appreciate the time it took to make all those changes. 

Looks good to me. I have a question about the best way to propagate the argument through to subsequent commands, but trust your judgement and am happy to leave as-is, just wanted to point it out.

payu/experiment.py
jo-basevi: Yes it is possible! I've added it in f89bff0. Only error I can see is when running commands like `payu-run` say that spins up a `collate`/`sync` job - would that be an issue?

jo-basevi:
As the intention was to add a command-line argument to make it easier to update manifests for released configurations, running `payu setup` (and maybe `payu sweep`) would be the only commands required to have `--metadata-off`. If wishing to disable metadata for `payu run`/`sync`/`collate` jobs using the `metadata: enable: false` in `config.yaml` would be the best option. So that simplifies this PR quite a bit!

payu/experiment.py
jo-basevi: Ok resolving this as removed this option from `run`/`sync`/`collate` commands

aidanheerdegen (approved):
LGTM. Thanks, and sorry for the extra work that wasn't required.

Issue #451, opened 2024-06-26
Fix metadata warning formatting 


jo-basevi:
The metadata related warnings look a bit broken with unclosed parentheses.
e.g.
```
/g/data/vk83/prerelease/apps/payu/dev/lib/python3.9/site-packages/payu/metadata.py:125: MetadataWarning: No experiment uuid found in metadata. Generating a new uuid
  warnings.warn("No experiment uuid found in metadata. "
Found experiment archive: /scratch/PROJECT/USER/access-esm/archive/access-esm1.5-configs
```
So it has the filename, line number, name of the warning and the line of the source code where warning get raised. The line of source code e.g. `warnings.warn(..`, can probably be left out of the log messages. 

Wondering whether it'll be worth using `logging.warning` instead of `warnings.warn` for metadata warnings as reading more python documentation (e.g. https://docs.python.org/3/howto/logging.html), `warnings.warn` is for if the warning can be avoided in the clients application, while `logging.warning` is when the situation should be noted. This makes more sense when warning the user that a new experiment uuid is being generated.

aidanheerdegen:
> Wondering whether it'll be worth using `logging.warning` instead of `warnings.warn` for metadata warnings as reading more python documentation (e.g. https://docs.python.org/3/howto/logging.html), `warnings.warn` is for if the warning can be avoided in the clients application, while `logging.warning` is when the situation should be noted. This makes more sense when warning the user that a new experiment uuid is being generated.

Sounds like good idea.

jo-basevi:
Closed by #755

PR #452, opened 2024-07-01
User-scripts error-handling, commands and tests

### LINKED ISSUE #444: Shell options when running userscripts and examples in config.rst
Following on from the [discussion about exporting environment variables for use in userscripts](https://github.com/payu-org/payu/issues/443), I've been playing around with running different userscripts and came across a small issue. The example in `config.rst` of a subcommand userscript: `echo "some_data" > input.nml` currently doesn't work.

Placing this in `config.yaml`
```yaml
userscripts:
   archive: echo "some_data" > input.txt
```
ends up just printing "some_data > input.txt" to the output rather than writing to a file:

```console
...
qsub -q copyq -P tm70 -l ncpus=1 -l mem=4GB -N pre-industria_c -l wd -j n -v PAYU_PATH=/g/data/hh5/public/apps/miniconda3/envs/analysis3-24.01/bin,PAYU_CURRENT_RUN=0,MODULESHOME=/opt/Modules/v4.3.0,MODULES_CMD=/opt/Modules/v4.3.0/libexec/modulecmd.tcl,MODULEPATH=/g/data/hh5/public/modules:/etc/scl/modulefiles:/apps/Modules/restricted-modulefiles/matlab_monash:/opt/Modules/modulefiles:/opt/Modules/v4.3.0/modulefiles:/apps/Modules/modulefiles -W umask=027 -l storage=gdata/access+gdata/hh5+gdata/tm70+gdata/vk83 -- /g/data/hh5/public/apps/miniconda3/envs/analysis3-24.01/bin/python3.10 /g/data/hh5/public/apps/miniconda3/envs/analysis3-24.01/bin/payu-collate
some_data > input.txt
Writing manifests/restart.yaml
...
```
I think this might be due to the default setting of `shell=False` used in the `sp.check_call` call in the `run_userscript` method though am not completely sure:
```python
def run_userscript(self, script_cmd):
        # First try to interpret the argument as a full command:
        try:
            sp.check_call(shlex.split(script_cmd))
```
I'm wondering whether it's preferable to update `run_userscript` to make commands like the example from `config.rst` possible, or instead to update `config.rst` to clarify the types of commands which can be run?

I think having `shell=True` would make it easier to provide arguments like environment variables and storage flags when running userscripts, but it still should be possible otherwise.

### LINKED ISSUE #448: Handling errors from user scripts
At the moment errors from user scripts just result in a pretty feeble warning message

https://github.com/payu-org/payu/blob/8a4a1e2e9171f01e37bd34bfd8bb6f209f550873/payu/experiment.py#L926

which is easily missed in a log file (and doesn't even seem to appear with https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/22).

I think payu should raise an error instead.

jo-basevi:
- Raise RuntimeErrors when user-scripts exit with an error (previously it was just a warning)
- Move run user-script code to payu.fsops out of Experiment class so it's easier to test
- Use shell=True for subprocess commands needing a shell, e.g. file redirections, pipes
- Add tests for running userscripts

To be able to run commands such as `echo "some_data" > input.txt` that are in payu documentation, it requires use of `shell=True`. So there's some security issues with using `shell=True` (https://docs.python.org/3/library/subprocess.html#security-considerations). However as the user scripts have typically been user defined `.sh` files run with `/bin/bash`, it could contain malicious code anyway..

I added a `needs_subprocess_shell` function that checks for any shell specific values. The list I've added might not be exhaustive. The question I am not sure on is that function even needed, and should all the user script commands just be run with `shell=True`?

Closes #448, closes #444

### DIFF
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -27,6 +27,11 @@ DEFAULT_CONFIG_FNAME = 'config.yaml'
 # Delete this once this bug in Lustre is fixed
 CHECK_LUSTRE_PATH_LEN = True
 
+# File extensions to script interpreters
+EXTENSION_TO_INTERPRETER = {'.py': sys.executable,
+                            '.sh': '/bin/bash',
+                            '.csh': '/bin/tcsh'}
+
 
 def mkdir_p(path):
     """Create a new directory; ignore if it already exists."""
@@ -229,3 +234,91 @@ def list_archive_dirs(archive_path: Union[Path, str],
 
     dirs.sort(key=lambda d: int(d.lstrip(dir_type)))
     return dirs
+
+
+def run_script_command(script_cmd: str, control_path: Path) -> None:
+    """Run a user defined script or command.
+
+    Parameters
+    ----------
+    script_cmd : string
+        String of user-script command defined in configuration file
+    control_path : Path
+        The control directory of the experiment
+
+    Raises
+    ------
+    RuntimeError
+        If there's was an error running the user-script
+    """
+    try:
+        _run_script(script_cmd, control_path)
+    except Exception as e:
+        error_msg = f"User defined script/command failed to run: {script_cmd}"
+        raise RuntimeError(error_msg) from e
+
+def needs_subprocess_shell(command: str) -> bool:
+    """Check if command contains shell specific values. For example, file
+    redirections, pipes or logical operators.
+
+    Parameters
+    ----------
+    command: string
+        String of command to run in subprocess call
+
+    Returns
+    -------
+    bool
+        Returns True if command requires a subprocess shell, False otherwise
+    """
+    shell_values = ['>', '<', '|', '&&', '$', '`']
+    for value in shell_values:
+        if value in command:
+            return True
+    return False
+
+def _run_script(script_cmd: str, control_path: Path) -> None:
+    """Helper recursive function to attempt running a script command.
+
+    Parameters
+    ----------
+    script_cmd : string
+        The script command to attempt to run in subprocess call
+    control_path: Path
+        The control directory to use for resolving relative filepaths, if file
+        is not found
+    """
+    # First try to interpret the argument as a full command:
+    try:
+        if needs_subprocess_shell(script_cmd):
+            subprocess.check_call(script_cmd, shell=True)
+        else:
+            subprocess.check_call(shlex.split(script_cmd))
+    except EnvironmentError as e:
+        # Now try to run the script explicitly
+        if e.errno == errno.ENOENT:
+            # Check if script is a file in the control directory
+            cmd = control_path / script_cmd
+            if cmd.is_file():
+                _run_script(str(cmd), control_path)
+            else:
+                raise
+
+        # If we get a "non-executable" error, then guess the type
+        elif e.errno == errno.EACCES:
+            _, file_ext = os.path.splitext(script_cmd)
+            shell_name = EXTENSION_TO_INTERPRETER.get(file_ext, None)
+            if shell_name:
+                print('payu: warning: Assuming that {0} is a {1} '
+                      'script based on the filename extension.'
+                      ''.format(os.path.basename(script_cmd),
+                                os.path.basename(shell_name)))
+
+                cmd = f'{shell_name} {script_cmd}'
+                _run_script(cmd, control_path)
+            else:
+                raise
+
+        # Otherwise re-raise the error
+        else:
+            raise
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -290,37 +290,33 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
         The control directory to use for resolving relative filepaths, if file
         is not found
     """
-    # First try to interpret the argument as a full command:
+    # First try to interpret the argument as a full command
     try:
         if needs_subprocess_shell(script_cmd):
             subprocess.check_call(script_cmd, shell=True)
         else:
             subprocess.check_call(shlex.split(script_cmd))
-    except EnvironmentError as e:
-        # Now try to run the script explicitly
-        if e.errno == errno.ENOENT:
-            # Check if script is a file in the control directory
-            cmd = control_path / script_cmd
-            if cmd.is_file():
-                _run_script(str(cmd), control_path)
-            else:
-                raise
+            print(script_cmd)
 
-        # If we get a "non-executable" error, then guess the type
-        elif e.errno == errno.EACCES:
-            _, file_ext = os.path.splitext(script_cmd)
-            shell_name = EXTENSION_TO_INTERPRETER.get(file_ext, None)
-            if shell_name:
-                print('payu: warning: Assuming that {0} is a {1} '
-                      'script based on the filename extension.'
-                      ''.format(os.path.basename(script_cmd),
-                                os.path.basename(shell_name)))
+    except FileNotFoundError:
+        # Check if script is a file in the control directory
+        cmd = control_path / script_cmd
+        if cmd.is_file():
+            _run_script(str(cmd), control_path)
+        else:
+            raise
 
-                cmd = f'{shell_name} {script_cmd}'
-                _run_script(cmd, control_path)
-            else:
-                raise
+    except PermissionError:
+        # Guess the type of interpreter using the file extension
+        _, file_ext = os.path.splitext(script_cmd)
+        shell_name = EXTENSION_TO_INTERPRETER.get(file_ext, None)
+        if shell_name:
+            print('payu: warning: Assuming that {0} is a {1} '
+                  'script based on the filename extension.'
+                  ''.format(os.path.basename(script_cmd),
+                            os.path.basename(shell_name)))
 
-        # Otherwise re-raise the error
+            cmd = f'{shell_name} {script_cmd}'
+            _run_script(cmd, control_path)
         else:
             raise
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -257,6 +257,7 @@ def run_script_command(script_cmd: str, control_path: Path) -> None:
         error_msg = f"User defined script/command failed to run: {script_cmd}"
         raise RuntimeError(error_msg) from e
 
+
 def needs_subprocess_shell(command: str) -> bool:
     """Check if command contains shell specific values. For example, file
     redirections, pipes or logical operators.
@@ -276,6 +277,7 @@ def needs_subprocess_shell(command: str) -> bool:
         if value in command:
             return True
     return False
+
 
 def _run_script(script_cmd: str, control_path: Path) -> None:
     """Helper recursive function to attempt running a script command.
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -296,7 +296,7 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
             subprocess.check_call(script_cmd, shell=True)
         else:
             subprocess.check_call(shlex.split(script_cmd))
-            print(script_cmd)
+        print(script_cmd)
 
     except FileNotFoundError:
         # Check if script is a file in the control directory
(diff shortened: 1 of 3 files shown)


payu/fsops.py
```diff
@@ -229,3 +234,93 @@ def list_archive_dirs(archive_path: Union[Path, str],
...
+        The control directory to use for resolving relative filepaths, if file
+        is not found
+    """
+    # First try to interpret the argument as a full command:
+    try:
+        if needs_subprocess_shell(script_cmd):
+            subprocess.check_call(script_cmd, shell=True)
+        else:
+            subprocess.check_call(shlex.split(script_cmd))
+    except EnvironmentError as e:
```
truth-quark, line 299: Adding random late night thoughts - this will likely require a revisit:
* Not having seen `EnvironmentError` before, I discovered it's [an alias of OSError](https://docs.python.org/3/library/exceptions.html#EnvironmentError)
* `errno.ENOENT` [maps to FileNotFoundError](https://docs.python.org/3/library/errno.html#errno.ENOENT)
* `errno.EACCES` [maps to PermissionError](https://docs.python.org/3/library/errno.html#errno.EACCES)

Would this func look better if reworked to a structure like:
```Python
try:
    # subprocess stuff
except OSError:
    # handler
except FileNotFoundError:
    # handler
except PermissionError:
    # handler

# then else or finally block if needed?
```
Another thought: would it be cleaner to detect the input type outside the exception handler block, then call the relevant processing func?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+@pytest.fixture
+def script_control_dir():
+    # Create a temporary control directory
+    control_dir = tmpdir / 'script_control_dir'
+    control_dir.mkdir()
+
+    yield control_dir
+
+    # Tear down
+    shutil.rmtree(control_dir)
```
truth-quark, line 315: Does [pytest's tmp-path fixture](https://docs.pytest.org/en/latest/reference/reference.html#tmp-path) provide this functionality?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+
+
+def test_run_userscript_python_script(script_control_dir: Path):
+    # Create a simple python script
+    python_script = script_control_dir / 'test_script.py'
+    with open(python_script, 'w') as f:
+        f.writelines([
+            f"with open('{script_control_dir}/output.txt', 'w') as f:\n",
```
truth-quark, line 323: Is there any need to test the chance of a script with a python shebang "header"?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+    with pytest.raises(RuntimeError):
+        payu.fsops.run_script_command('unknown_userscript.sh',
+                                      script_control_dir)
+
+
+def test_userscript_non_existent_file(script_control_dir: Path):
+    # Test userscript raises an error
+    with pytest.raises(RuntimeError):
+        payu.fsops.run_script_command('unknown_userscript.sh',
+                                      script_control_dir)
```
truth-quark, line 374: This test looks duplicated, unless there's an intent to test another branch of the run script code?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+
+
+def test_run_userscript_python_script_eror(script_control_dir):
+    # Create a python script that'll exit with an error
+    python_script = script_control_dir / 'test_script.py'
```
truth-quark, line 379: As a minor stylistic observation, what are your thoughts on naming this `test_script_error.py` to distinguish from the other `test_script.py`?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+
+
+def test_userscript_non_existent_file(script_control_dir: Path):
+    # Test userscript raises an error
+    with pytest.raises(RuntimeError):
+        payu.fsops.run_script_command('unknown_userscript.sh',
+                                      script_control_dir)
+
+
+def test_run_userscript_python_script_eror(script_control_dir):
```
truth-quark, line 377: Should there be a test to cover bash scripts returning with non-zero?

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+    cmd = 'echo -n "some_data" > test.txt'
+
+    # Test payu userscript
+    with cd(script_control_dir):
+        payu.fsops.run_script_command(cmd, script_control_dir)
+
+    # Check userscript output
+    with open((script_control_dir / 'test.txt'), 'r') as f:
+        content = f.read()
+        assert content == "some_data"
```
truth-quark, line 400: Minor/optional consistency thing, the other tests have `assert f.read() == "some_data"`.

test/test_payu.py
```diff
@@ -301,3 +301,116 @@ def test_list_archive_dirs():
...
+
+
+@pytest.mark.parametrize("command, expected", [
+    ('echo "Some Data" > test.txt', True),
+    ('ls -l | grep "test"', True),
+    ('cmd1 && cmd2', True),
+    ('cmd1 || cmd2', True),
+    ('echo "Data"', False),
+    ('ls -l', False),
+    ('some_python_script.py', False),
```
truth-quark, line 410: Is there any need to add a test for a relative path, like `./some/dir/app.py <some> <args>`?

payu/fsops.py
```diff
@@ -27,6 +27,11 @@
 # Delete this once this bug in Lustre is fixed
 CHECK_LUSTRE_PATH_LEN = True
 
+# File extensions to script interpreters
+EXTENSION_TO_INTERPRETER = {'.py': sys.executable,
+                            '.sh': '/bin/bash',
+                            '.csh': '/bin/tcsh'}
+
 
 def mkdir_p(path):
```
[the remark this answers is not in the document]
jo-basevi: Yeah I think it would actually, I'll make a separate issue for it as it is used widely throughout the code

payu/fsops.py
jo-basevi: Thanks, I've updated the function with the using `FileNotFoundError` and `PermissionError` and it does look cleaner!

jo-basevi: > Another thought: would it be cleaner to detect the input type outside the exception handler block, then call the relevant processing func? 

Do you mean like instead of the `FileNotFoundError`, check first if the cmd is a file in the control path?

test/test_payu.py
jo-basevi: Yes it does! I've updated to use the tmp_path fixture

test/test_payu.py
jo-basevi: Yeah it was duplicated, I should've have spotted that.. I've removed it now

test/test_payu.py
jo-basevi: Added that in

test/test_payu.py
jo-basevi: I've added in a test that is a bash script that raises an error

test/test_payu.py
jo-basevi: I've just added that in, thanks

test/test_payu.py
jo-basevi: Just added in a test for that

test/test_payu.py
jo-basevi: Yeah, I've just added in a new test for that

payu/fsops.py
truth-quark: > > Another thought: would it be cleaner to detect the input type outside the exception handler block, then call the relevant processing func?
> 
> Do you mean like instead of the `FileNotFoundError`, check first if the cmd is a file in the control path?

Yep, I imagined something like
```Python
def detect_script_type(script_cmd):
    if is_a_script_path(script_cmd):  # runs some detection mechanism
        return something  # return a function that runs scripts???
    else:
        # check for user commands etc.
        return something_else

def workflow(script_cmd):
    # ... processing ...
    func = detect_script_type(script_cmd)
    func(script_cmd)
    # ... processing ...    
```
This may not gain much beyond simply extracting some logic from the new exception handler blocks. Do you see any other reason(s) to modify the exception block approach?

payu/fsops.py
```diff
@@ -27,6 +27,11 @@
 # Delete this once this bug in Lustre is fixed
 CHECK_LUSTRE_PATH_LEN = True
 
+# File extensions to script interpreters
+EXTENSION_TO_INTERPRETER = {'.py': sys.executable,
```
truth-quark, line 31: Does line 31 need to detect the Python `sys.executable` or potentially a _virtual environ_ executable?

payu/fsops.py
```diff
@@ -229,3 +234,89 @@ def list_archive_dirs(archive_path: Union[Path, str],
...
+
+    except PermissionError:
+        # Guess the type of interpreter using the file extension
+        _, file_ext = os.path.splitext(script_cmd)
+        shell_name = EXTENSION_TO_INTERPRETER.get(file_ext, None)
+        if shell_name:
+            print('payu: warning: Assuming that {0} is a {1} '
+                  'script based on the filename extension.'
+                  ''.format(os.path.basename(script_cmd),
+                            os.path.basename(shell_name)))
```
truth-quark, line 317: For minor styling, is there a preference for _f-strings_ in the project?

Also, is this OK as a `print` or would it be better in `warnings.warn()` call?

truth-quark (approved):
Looks good to go. Virtually all the comments are small things.

The biggest thing is suggesting moving `subprocess` security to a separate issue for ease of handling.

payu/fsops.py
```diff
@@ -229,3 +234,89 @@ def list_archive_dirs(archive_path: Union[Path, str],
...
+    script_cmd : string
+        The script command to attempt to run in subprocess call
+    control_path: Path
+        The control directory to use for resolving relative filepaths, if file
+        is not found
+    """
+    # First try to interpret the argument as a full command
+    try:
+        if needs_subprocess_shell(script_cmd):
+            subprocess.check_call(script_cmd, shell=True)
```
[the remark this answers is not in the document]
jo-basevi: Unfortunately I think `shlex.quote()` breaks commands like `echo -n "some_data" > test.txt`. I think it's used when you are expecting like one value or arguments like `echo -n {shlex.quote(some_user_input)} > test.txt`

payu/fsops.py
jo-basevi: Yeah I briefly had a go at that, but then ended writing a bit more code than the current recursive approach.

payu/fsops.py
jo-basevi: I might keep as print for now and revisit it as part of this issue which aims to fix some of the warning formatting in the project  https://github.com/payu-org/payu/issues/451

payu/fsops.py
jo-basevi: I think it detects the python that is being used in the environment used to run payu.

truth-quark: For future reference:
* From the system Python, `sys.executable` resolves to something `/usr/bin/python3`
* Inside a virtualenv, `sys.executable` resolves to the virtualenv `bin` dir & the symlink to the python binary

test/test_payu.py
```diff
@@ -301,3 +301,134 @@ def test_list_archive_dirs():
...
+    with open(python_script, 'w') as f:
+        f.write('raise ValueError("Test that script exits with error")')
+
+    # Test userscript raises an error
+    with pytest.raises(RuntimeError):
+        payu.fsops.run_script_command('test_script_error.py',
+                                      tmp_path)
+
+
+def test_run_userscript_bash_script_eror(tmp_path):
```
CodeGat, line 392: `test_run_userscript_bash_script_eror` should be `error`

CodeGat (approved):
A spelling error was all I could pull you up on!

test/test_payu.py
jo-basevi: Oops! Have just fixed that

CodeGat (approved):

Issue #419, opened 2024-02-19
Unnecessary files in UM config_files list.


MartinDix:
The config_files list includes a number of files that the model doesn't actually use. Model configurations would be clearer without the duplicated namelists. See https://github.com/coecms/access-esm/issues/14

aidanheerdegen:
A simple change would be to move them to the `optional_config_files`

https://github.com/payu-org/payu/blob/master/payu/models/um.py#L48

blimlim:
I've had a go at removing the extra files [listed here](https://github.com/coecms/access-esm/issues/14) in a branch of payu. In addition to shortening the `um.py` config list, I had to remove the following lines:

https://github.com/payu-org/payu/blob/8a4a1e2e9171f01e37bd34bfd8bb6f209f550873/payu/models/um.py#L144-L150

As far as I can tell, `parexe` is empty (at least in the pre-industrial configuration), and it appears to remain so after these lines are run. I just wanted to get confirmation that this would be safe to do, as I'm unsure of the original purpose of `parexe`.

---

Removing the extra config files and running for a few months produces identical simulations to unmodified configurations. The following shows the difference in surface temperature for a simulation with the configuration files removed and one without, on March 31:

![download-5](https://github.com/payu-org/payu/assets/88933912/1b8aeec1-e4c6-41ca-8b20-2ca31a2217b7)

---

If this sounds ok I'll add pull requests to add the changes to payu and the configurations.

blimlim:
Closing this issue and have added a [new one](https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/27) to remove the extra files in the configurations.

blimlim:
Closing this issue and have added a [new one](https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/27) to remove the extra files in the configurations.

PR #455, opened 2024-07-03
Remove unnecessary UM config files

blimlim:
Changes related to #419.

Payu currently requires several UM configuration files which aren't required and duplicate settings in the required ones, making it confusing to modify the correct namelists. 

I've removed the unneeded files from the `config_files` in `um.py`, though kept `parexe` as an `optional_configuration_file` in case it is needed by older configurations.

### DIFF
--- a/payu/models/um.py
+++ b/payu/models/um.py
@@ -140,15 +140,15 @@ class UnifiedModel(Model):
                                            work_path=self.work_path)
         os.environ.update(um_vars)
 
-        # parexe removed from newer configurations - retain the old processing if 
-        # it exists for backwards compatibility 
+        # parexe removed from newer configurations - retain the
+        # old processing if file exists for backwards compatibility
         try:
             # The above needs to be done in parexe also.
-            # FIXME: a better way to do this or remove.  
+            # FIXME: a better way to do this or remove.
             parexe = os.path.join(self.work_path, 'parexe')
             for line in fileinput.input(parexe, inplace=True):
                 line = line.format(input_path=self.input_paths[0],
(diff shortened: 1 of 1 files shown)


payu/models/um.py
```diff
@@ -141,13 +140,18 @@ def setup(self):
...
+        try:
+            # The above needs to be done in parexe also.
+            # FIXME: a better way to do this or remove.
+            parexe = os.path.join(self.work_path, 'parexe')
+            for line in fileinput.input(parexe, inplace=True):
+                line = line.format(input_path=self.input_paths[0],
+                                   work_path=self.work_path)
+                print(line, end='')
+        except FileNotFoundError:
+            pass
```
jo-basevi, line 154: Alternatively `os.path.isfile()` could be used to see if parexe file exists, instead of the try except block - but both work fine
```suggestion
        # The above needs to be done in parexe also, if it exists
        parexe = os.path.join(self.work_path, 'parexe')
        if os.path.isfile(parexe):
            # FIXME: a better way to do this or remove.
            for line in fileinput.input(parexe, inplace=True):
                line = line.format(input_path=self.input_paths[0],
                                   work_path=self.work_path)
                print(line, end='')
```

jo-basevi (approved):
The changes look good to me!

payu/models/um.py
blimlim: Great idea, I think that makes it clearer to read!

jo-basevi (approved):
Re-approving the changes 👍

aidanheerdegen (approved):
LGTM2. Thanks @blimlim

PR #461, opened 2024-07-11
Add staged CABLE driver for CABLE configurations

Whyborn:
This driver is intended to facilitate spin-up CABLE configurations, which involve usually many stages with differing science configurations. An example experiment exists here, with some documentation: [Staged_CABLE](https://github.com/CABLE-LSM/Staged-CABLE-in-payu/tree/main)


payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+                    for step_id, _ in enumerate(step_names):
+                        if step_counts[step_id] > step:
+                            cable_stages.append(step_names[step_id])
+            # Finish handling of multistep stage
+
+            else:
+                # A single step stage, in general we only want to run this
+                # once, but check for the count anyway
+                for _ in range(stage_opts['count']):
+                    cable_stages.append(stage_name)
```
aidanheerdegen, line 121: I think this can be replaced with something like:
```python
for stage, repeat in self.stage_config.items();
   cable_stages.extend([stage] * repeat))
```
Possibly could be done as a one-liner list comprehension, but I got a bit bogged down in the syntax, and this is relatively clean and understandable.

payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+    def setup(self):
+        super(StagedCable, self).setup()
+
+        # Directories required by CABLE for outputs
+        os.makedirs(os.path.join(self.work_output_path, 'logs'),
+                    exist_ok=True)
+        os.makedirs(os.path.join(self.work_output_path, 'restart'),
+                    exist_ok=True)
+        os.makedirs(os.path.join(self.work_output_path, 'outputs'),
+                    exist_ok=True)
```
aidanheerdegen, line 135: Could reduce code duplication
```suggestion
        for dir in ['logs', 'restart', 'outputs']:
           os.makedirs(os.path.join(self.work_output_path, dir),
                    exist_ok=True)
```

payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+        namelists = os.listdir(stage_name)
+
+        for namelist in namelists:
+            with open(namelist, 'r') as master_nml_f:
+                master_namelist = f90nml.read(master_nml_f)
+
+            with open(os.path.join(stage_name, namelist), 'r') as stage_nml_f:
+                stage_namelist = f90nml.read(stage_nml_f)
+
+            deep_update(master_namelist, stage_namelist)
```
aidanheerdegen, line 177: `f90nml` has a `patch` function that might do what you want

https://github.com/marshallward/f90nml/blob/21db6a7afc6b5df6e53cdc1980f1301660732d01/f90nml/namelist.py#L654-L668

The docs suggest you an call a higher level `patch` function without needing to do any reading of input and output namelists, just pass the filenames and the patch and it does the rest:
```
f90nml.patch('sample.nml', patch_nml, 'new_sample.nml')
```

payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Update the configuration log and save it to the working directory
+        completed_stage = self.configuration_log['current_stage']
+        self.configuration_log['current_stage'] = ''
+        self.configuration_log['completed_stages'].append(completed_stage)
+
+        self.save_configuration_log()
+
+        if int(os.environ['PAYU_N_RUNS']) == 1:
+            os.remove('configuration_log.yaml')
```
aidanheerdegen, line 195: I'm wondering if we can give an error message if a user attempts to set the number of runs with `-n`?

It would probably require inspecting the value of `PAYU_N_RUNS` at the beginning.

Perhaps it is sufficient to print to STDOUT when the number of runs is set, to let them know it is being overridden.

aidanheerdegen (commented):
I have some questions and suggestions.

payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+            self._build_new_configuration_log()
+        else:
+            # Read the current configuration log
+            self._read_configuration_log()
+
+        # Add the restart directories to inputs
+        num_completed_stages = len(self.configuration_log['completed_stages'])
+        for restart_dir in reversed(range(num_completed_stages)):
+            config['input'].append(os.path.realpath(
+                f'archive/output{restart_dir:03d}/restart'))
```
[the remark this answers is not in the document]
Whyborn: I couldn't work out precisely whether this function would give me the functionality I needed i.e. what is done with the list of files returned from this function. Is it simply a list of files that ended up being added to the manifest?

payu/models/staged_cable.py
Whyborn: That will work for the single step stages, but I don't think it will for the multi-step stages. An example of a multistep stage in the config file would be
```
multistep_demo_stage:
    demo_stage_1:
        count: 4
    demo_stage_2:
        count: 2
```
The desired output is ```["demo_stage_1, "demo_stage_2", "demo_stage_1", "demo_stage_2", "demo_stage_1", "demo_stage_1"]```, but I think the one liner would give ```["demo_stage_1", "demo_stage_1", "demo_stage_1", "demo_stage_1", "demo_stage_2", "demo_stage_2"]```. There is perhaps a cleaner way to do it than the way I did it, but I couldn't think of it at the time.

payu/models/staged_cable.py
Whyborn: I tried this originally, but it wasn't handling struct attributes correctly. Turns out that the object method ```patch``` does not work as expected, but calling the library ```patch``` function as you recommend here does work as expected.

payu/models/staged_cable.py
aidanheerdegen: Ah, sorry my misunderstanding. This is where tests are super-useful (some might even say essential), as it gives a very clear indication of what the expected output would be. 

Depending on what is required [itertools](https://docs.python.org/3/library/itertools.html) provides a bunch of useful functions for these sorts of shenanigans.

Whyborn:
I'll add some tests before requesting another review.

payu/models/staged_cable.py
Whyborn: Is cutting down the number of lines so important? I think it's not a such a bad idea to have it explicitly written with simple methods, but then again, the implementation didn't make clear what was going on. Perhaps a better idea is to find a convenient method and make the explanation in the comments clearer.

ccarouge: I agree with @aidanheerdegen that I'm not sure what is going on in the code here. What about using `itertools.zip_longest()`? You pad the shortest stages with a given missing value, then you can use list comprehension to remove this missing value from the list.

This has the advantage that people know what `zip` does in Python.

aidanheerdegen: Agreed clarity is the most important factor, and I didn't read the code carefully enough to understand what it was trying to do which was poor, sorry. But as you say, perhaps it could be expressed more clearly. Also tests. Did I mention tests?

Using something from `itertools` at least means using a well documented function, with a widely understood purpose.

Adapted from https://stackoverflow.com/a/48200012 this shows one way of doing it.
```python
>>> list(filter( lambda x: x is not None, chain.from_iterable(zip_longest(['a', 'b', 'c', 'd'], [1, 2]))))
['a', 1, 'b', 2, 'c', 'd']
```

By itself maybe not that readable, but as you say with a comment line "Build a list interleaving steps using `zip_longest`, and use `chain` and `filter` to flatten and remove missing elements"

Or you could use a [recipe from the itertools docs](https://docs.python.org/3/library/itertools.html#itertools-recipes):
```python
def roundrobin(*iterables):
    "Visit input iterables in a cycle until each is exhausted."
    # roundrobin('ABC', 'D', 'EF') → A D E B F C
    # Algorithm credited to George Sakkis
    iterators = map(iter, iterables)
    for num_active in range(len(iterables), 0, -1):
        iterators = cycle(islice(iterators, num_active))
        yield from map(next, iterators)
```

Whyborn: I have adapted the first suggestion with some minor changes, compacting the handling of the multi-step stage to a few lines.

payu/models/staged_cable.py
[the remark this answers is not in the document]
aidanheerdegen: The way manifests work (in general) is that they reflect the state of whatever is in the `work` directory after all the model setup steps. This was deliberate, rather than inject manifest stuff in every time a model driver made some modification to what was in the `work` directory, just wait until it all washes out and then record the result.

So I believe `get_prior_restart_files` just gets called and then loops through the files creating symlinks into `work`

https://github.com/payu-org/payu/blob/83001a2b45ef786e178894f77dd9fbc0878e73c1/payu/models/model.py#L261-L270

payu/models/staged_cable.py
aidanheerdegen: Weird we overlapped with identical suggestions @ccarouge. I think I must have started my comment before yours, so didn't see it, otherwise I clearly wouldn't have bothered. Yours was much more succinct! :)

Whyborn: Ended up going with an adaptation of the first solution.

payu/models/staged_cable.py
```diff
@@ -0,0 +1,205 @@
...
+
+# Local
+from payu.models.model import Model
+
+
+def deep_update(d_1, d_2):
+    """Deep update of namelists."""
+    for key, value in d_2.items():
+        if isinstance(value, dict):
+            deep_update(d_1[key], d_2[key])
```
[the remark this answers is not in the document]
Whyborn: Removed for `f90nml.patch`.

payu/models/staged_cable.py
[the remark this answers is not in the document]
Whyborn: The issue I see here is that (I think) `self.prior_restart_path` points strictly the most recent output directory in the `archive`. This precludes us from pulling restarts from earlier stages, which we often do.

aidanheerdegen:
Are you ready for another review @Whyborn?

Whyborn:
I had requested another one, but I have been doing some thinking since the Payu workshop session to improve the handling of restarts. I'll cancel the current request and add a new one after I make (or not make) some changes.

Whyborn:
I've made some adjustments to the way we retrieve the restarts, see the unresolved conversation. I've added tests to capture most of the desired functionality, but I was struggling to build a good test for retrieving restarts.

payu/models/staged_cable.py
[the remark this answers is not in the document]
Whyborn: I don't think there was a workable solution built around extending the ```self.get_prior_restart_files()```, because of the way it handles the files returned by this method. To have the correct file added to the manifest, files from restarts that are not the most recent would need to be prepended by ```"../restart{x}"```, which would have unintended consequences for the link location.

The solution I went with runs through the previous restart directories, from newest to oldest, and adds the files in the restart directories to the manifest if said file doesn't already exist in the manifest. This way we don't write over newer versions of a given restart, but still have the ability to mix and match restarts from different science modules.

payu/models/staged_cable.py
```diff
@@ -152,7 +195,7 @@ def _prepare_stage(self):
             yaml.dump(self.configuration_log, conf_log_f)
 
         # Ensure the directory exists
-        if not os.path.isdir(stage_name):
+        if not os.path.isdir(os.path.join(self.control_path, stage_name)):
             errmsg = f"""Directory containing namelists for stage {stage_name}
              does not exist."""
             raise FileNotFoundError(errmsg)
```
ccarouge, line 201: This might be better done earlier within/after the `__init__()` when we first read the stage configuration file. This makes sure any configuration error is picked right at the start of the process.

As it stands, if there is a directory missing for step 5, one has to run the 4 first steps to know about it, correct?

ccarouge (changes requested):
I haven't looked at the tests and I'm running out of time so I'll send what I've picked up.

I would refactor `_prepare_stage()`. I don't like that it calls `_apply_stage_namelists()`. I would transform `_prepare_stage()` into a `_get_stage_name()` function that returns the current stage name. Then call that from `_apply_stage_namelists()` or from within `setup()`. 
Then I would introduce a `deal_with_configuration_log()` function (name to be determine) that is called at the end of `setup()` and deals with saving the configuration log file (call the function you have for that) and copying it to the work directory.

One final comment for @aidanheerdegen , do you have any aspiration to use `Pathlib` instead of `os.path` in payu? If so, maybe it's a bad idea to introduce a new driver based on `os.path`. Your call @aidanheerdegen .

aidanheerdegen:
I think I'm at the end of my utility as a reviewer. I see @ccarouge has a few more suggestions, but after that probably best to merge it and start using it and iterate if there are problems, or improvements required. You're not touching any other drivers, so it's fairly side-effect free and you can test it, as can others.

payu/models/staged_cable.py
Whyborn: This is true, although there could be advantages to not checking that all stage directories exist during initialisation, as it permits restarting from a later stage even if the prior stage directories no longer exist (or maybe we don't to allow this?). As it stands, one can simply ```payu sweep``` and ```payu run``` after addressing errors at a given stage, and it will just restart at the failed stage as if we had just run the prior stages. I think it makes sense to check this when we determine the stage name.

aidanheerdegen (approved):
I've approved as Claire is away in case you're ready to merge @Whyborn

PR #465, opened 2024-07-23
Adding date based pruning for ESM

### LINKED ISSUE #464: Date based pruning for ESM1.5
Date based restart pruning is not currently implemented for ESM1.5, leading to a `NotImplementedError` if it's enabled in the `config.yaml` file. This shouldn't be too hard to add, especially as it's currently implemented for MOM, which is a submodel for ESM1.5.

blimlim:
Closes #464 

This pull requests adds support for date based restart pruning for ESM1.5, which was previously implemented for the MOM models. As ESM1.5 uses MOM as a submodel, and produces the same ocean date files `ocean_solo.res` in the restart directories, I've just copied the `get_restart_datetime` method from `mom_mixin.py` into `access.py`. 

I've manually tested it with with different monthly and yearly restart frequencies. E.g. running in month long sections, with `restart_freq: 3MS` for a total of 10 month results in the following archive:

```
metadata.yaml  output000  output001  output002  output003  output004  output005  output006  output007  output008  output009  pbs_logs  restart000  restart003  restart006  restart009
```

Likewise running for 6 months with `restart_freq: 2MS` results in 
```
metadata.yaml  output000  output001  output002  output003  output004  output005  restart000  restart002  restart004  restart005
```

Creating a fake group of restart directories with yearly gaps in their `ocean_solo.res` dates, and running the pruning with `restart_freq: 7YS` also looks like it works (note that restart010 was the first of the fake directories):
```
metadata.yaml  pbs_logs    restart017  restart031  restart045  restart059  restart073  restart087  restart101 
output102      restart010  restart024  restart038  restart052  restart066  restart080  restart094  restart102
``` 

The implementation and manual tests are both a bit ad-hoc, and so it would be great to get another opinion on whether what's here is sufficient or if its better to try and do it differently.

### DIFF
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -235,7 +235,7 @@ class Access(Model):
                 return model.get_restart_datetime(mom_restart_path)
 
         raise NotImplementedError(
-            'Cannot find mom sub-model: access-om2 date-based restart pruning '
+            'Cannot find mom sub-model: access-ESM1.5 date-based restart pruning '
             'requires the mom sub-model to determine restart dates')
 
     def set_model_pathnames(self):
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -226,6 +226,18 @@ class Access(Model):
             o2i_dst = os.path.join(cice5.restart_path, 'o2i.nc')
             shutil.copy2(o2i_src, o2i_dst)
 
+    def get_restart_datetime(self, restart_path):
+        """Given a restart path, parse the restart files and
+        return a cftime datetime (for date-based restart pruning)"""
+        for model in self.expt.models:
+            if model.model_type == 'mom':
+                mom_restart_path = os.path.join(restart_path, model.name)
+                return model.get_restart_datetime(mom_restart_path)
+
+        raise NotImplementedError(
+            'Cannot find mom sub-model: access-om2 date-based restart pruning '
+            'requires the mom sub-model to determine restart dates')
+
     def set_model_pathnames(self):
         pass
 
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -23,6 +23,7 @@ import f90nml
 # Local
 from payu.fsops import make_symlink
 from payu.models.model import Model
+from payu.models.mom import get_restart_datetime_using_mom_submodel
 import payu.calendar as cal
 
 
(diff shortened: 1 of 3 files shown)


jo-basevi:
Thanks for opening this PR and running some tests!

I think the testing is  probably sufficient. There's integration-like tests for pruning logic using a couple years of restarts and using `access-om2` model's `get_restart_datetime` method (which is similar logic wise) in `payu/test/test_prune_restarts.py`. There's tests for checking the datetime in `ocean_solo.res` files are parsed  in `test/models/test_mom_mixin.py`. 

Testing has been a bit tricky with setting up Experiment classes which is needed to initialise Model classes. Current tests or functions probably need a rethink to make it more "unit"-ish and include mocking, but that's definitely outside of the scope of this PR!

An potential option without creating new tests is to refactor so the `access-om2` and `access` drivers to use the same function, e.g. could add a generic helper function to the `mom.py` file or similar?

```python
def get_restart_datetime_using_mom_submodel(model, restart_path):
    """Given a model object and a restart path, check for
    a MOM sub-model and use it's restart files to find a cftime datetime 
   (for date-based restart pruning)"""
    for sub_model in model.expt.models:
        if sub_model.model_type == 'mom' :
            mom_restart_path = os.path.join(restart_path, sub_model.name)
            return sub_model.get_restart_datetime(mom_restart_path)

    raise NotImplementedError(
        f'Cannot find mom sub-model: {model.model_type} date-based '
        'restart pruning requires the mom sub-model to '
        'determine restart dates'
    )
```
Then in access/access-om2 drivers:
```python
def get_restart_datetime(self, restart_path):
        """Given a restart path, parse the restart files and
        return a cftime datetime (for date-based restart pruning)"""
       # Use mom submodel to determine restart datetimes
        get_restart_datetime_using_mom_submodel(model=self, restart_path=restart_path)
```

Having a helper function outside of a `Model` driver class might make testing easier in future too, as can just create a mocks of the model class. (I'm now wondering if the mom mixin should instead just be helper functions too...)

That said, it's not a big issue and I'm sure the code you have added works, so I am happy to hit approve on this PR!

jo-basevi (dismissed):

blimlim:
Thanks for reviewing this! I think thats a great idea with the helper function, and it helps avoid copying the exact same code across the different models. I've had a go at copying that in and it still is working with the manual tests of ESM. Would you be happy to have a quick look through those changes? No rush though!

jo-basevi (approved):
Thanks! Re-approving the changes

Issue #463, opened 2024-07-17
Syncing and NetCDF conversion for ESM1.5


blimlim:
The NetCDF conversion script for ESM1.5 will be run as an `archive` stage userscript, where it is submitted to the queue as a PBS job to avoid holding up the simulation. If syncing is enabled in `config.yaml`, there's no guarantee that the conversion job will be finished by the time the syncing job is run. E.g. testing several month long simulations with syncing enabled, some month's NetCDF files were not all copied over to the syncing directory:

```
<sync-dir>/output000/atmosphere/NetCDF:
aiihca.paa1jan.nc  aiihca.pea1jan.nc  aiihca.pga1jan.nc

<sync-dir>/output005/atmosphere/NetCDF:
aiihca.pea1jun.nc
```

From what I can understand, the collation step gets around a similar issue by calling the sync step from within the collation job:

https://github.com/payu-org/payu/blob/2391485ac4af756f45ae6d4720cd263b30940ae5/payu/subcommands/collate_cmd.py#L110-L112

and but a similar approach might not be possible with the external conversion userscript. Just flagging this now so that we can discuss ideas to handle this.

jo-basevi:
### Postscript?
An option could be to run the `archive` user-script as a `postscript`. Postscripts are always resubmitted via `qsub` and are run after collation or archive, if collation is not enabled. E.g. in `config.yaml`:
```yaml
postscript: my_script.sh
```

The command is run in payu here:
https://github.com/payu-org/payu/blob/e9bd1f4c4adc223e818c71f8d45c36de98025dc4/payu/experiment.py#L834-L837

If `postscript` is enabled, syncing latest output is disabled as payu has no idea when the postscript finishes, so only previous outputs are synced  (e.g. `outputsN` where `N < current_run`). 

Passing specific environment vars such as current output dir could be done, using `-v` to export the environment variable to the pbs job, e.g.
```
postscript: -v PAYU_CURRENT_OUTPUT_DIR my_script.sh
```

### Sync Userscript?

So the sync job has a `userscript` option which runs before any `rsync` commands are run. It means that nothing will be synced if the userscript fails to run. The sync job by default runs on `copyq` but queue and resources can be configured, see [payu postprocessing](https://payu.readthedocs.io/en/latest/config.html#postprocessing) documentation. However this may not work for this esm1.5 script as it might need to run every time- not just when sync is configured?

### Adding Dependency Logic to Payu?
[...]

PR #467, opened 2024-07-26
Add environment variables and shell checks for postscripts

blimlim:
This pull request covers initial work for #463.  

Allowing both syncing and NetCDF conversion for ESM simulations will be easier if the NetCDF conversion is run as a payu postscript, rather than as a userscript. This pull request adds in shell requirement checks and environment variable exports to the postscript submission, which are needed for the NetCDF conversion script.


jo-basevi (approved):
Thanks for adding that in! The changes look good to me

PR #477, opened 2024-08-05
Replace deprecated `distutils` dependency

### LINKED ISSUE #476: Pip install payu fails with python3.12
As brought up in this forum post: https://forum.access-hive.org.au/t/deprecation-of-distutils-from-python/2264, payu uses `distutils.sysconfig` however `distutils` is no longer in Python 3.12. This can be updated to import the `sysconfig` standard library module. 

```
  File "/g/data/PROJECT/USER/payu/model-config-tests/venv/bin/payu", line 5, in <module>
    from payu.cli import parse
  File "/g/data/PROJECT/USER/payu/model-config-tests/venv/lib/python3.12/site-packages/payu/cli.py", line 11, in <module>
    from distutils import sysconfig
ModuleNotFoundError: No module named 'distutils'
```

jo-basevi:
A `distutils.sysconfig` import was used in `payu/cli.py`. `distutils` is now deprecated and has been removed in Python 3.12. The PR replaces `distutils` with the `sysconfig` standard library (as recommended here: https://peps.python.org/pep-0632/#migration-advice).

I've also added Python 3.12 to CI tests. Keeping the Python 3.9 tests at this stage as Python 3.9 is still currently used in the `vk83`'s payu environments on NCI. 

Closes #476

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,7 +62,7 @@ jobs:
     # Run the job for different versions of python
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -8,7 +8,7 @@
 """
 
 import argparse
-from distutils import sysconfig
+import sysconfig
 import importlib
 import os
 import pkgutil


.github/workflows/CI.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     # Run the job for different versions of python
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
aidanheerdegen, line 65: This is fine for the moment, because we're still using 3.9 in deployment, but when we drop that we should also remove this from the CI so we don't just keep adding versions. Three seems like a decent span to test against.

aidanheerdegen (approved):
LGTM. Thanks.

.github/workflows/CI.yml
jo-basevi: Ok, will do that once this issue updating to 3.10 is closed (https://github.com/ACCESS-NRI/payu-condaenv/issues/32)

PR #482, opened 2024-08-13
Update check for load user modulefiles to allow  exact match

### LINKED ISSUE #481: Payu's check for module uniqueness fails incorrectly
If in config.yaml, I specify:

```
modules:
        - openmpi/4.1.5
```

and then do a payu run, using payu 1.1.4, the following error is generated:

```
  File "/g/data/vk83/apps/payu/1.1.4/bin/payu-run", line 10, in <module>
    sys.exit(runscript())
  File "/g/data/vk83/apps/payu/1.1.4/lib/python3.10/site-packages/payu/subcommands/run_cmd.py", line 133, in runscript
    expt.setup()
  File "/g/data/vk83/apps/payu/1.1.4/lib/python3.10/site-packages/payu/experiment.py", line 431, in setup
    self.setup_modules()
  File "/g/data/vk83/apps/payu/1.1.4/lib/python3.10/site-packages/payu/experiment.py", line 237, in setup_modules
    loaded_mods, paths = envmod.setup_user_modules(user_modules,
  File "/g/data/vk83/apps/payu/1.1.4/lib/python3.10/site-packages/payu/envmod.py", line 173, in setup_user_modules
    raise ValueError(
ValueError: There are multiple modules available for openmpi/4.1.5:
/apps/Modules/modulefiles:
openmpi/4.1.5
openmpi/4.1.5-debug
```

This behavior is incorrect, as `openmpi/4.1.5` is the whole name of the module and prevents the user specifying the openmpi version.

The check at line 163 in [envmod](https://github.com/payu-org/payu/blob/83001a2b45ef786e178894f77dd9fbc0878e73c1/payu/envmod.py#L162) might need modifying with an additional check if the module name specified is an exact match to the available module, or converting to a warning.

jo-basevi:
Allow checks for "modules: load:" modulefiles to allow when there's an exact match in available modules. This is to fix bug #481 where module avail for `openmpi/4.1.5` includes `openmpi/4.1.5` and `openmpi/4.1.5-debug`.

I've also added tests for parsing module avail output.

Closes #481.

@aidanheerdegen last week there was an issue with multiple modules found in `/g/data/vk83/modules` and `/g/data/vk83/modules/access-models`, even though they were the same modulefile. I noticed in the modules docs there's a `module paths` command that lists paths of available modulefiles matching a pattern so for the above, it would still evaluate to the same path. There's a couple of differences to `module avail`:
-  Running the command `module paths` using `modulecmd` (`$MODULESHOME//bin/modulecmd`) generates a bunch of code. So for `bash`, it generates a bunch of echo statements, rather than output going to `stderr` which is the case for `module avail`. So to get what values are printed, have to run the generated code, and then capture the output.
- Specifying `openmpi/` with a `/` just matches the default module version (I think), so only one path is returned.

If it's unlikely to have a case where `module/version` can be accessed from two different modulepaths and are different, e.g. 

```
/path/to/modulefiles:
test-module/1.0.0
/another/module/path:
test-module/1.0.0
```

The check could be changed from `modules_avail.count(modulefile) != 1` to `modulefile not in modules_avail` in this PR. Maybe this be moved to separate to avoid blocking this PR?

### DIFF
--- a/test/test_envmod.py
+++ b/test/test_envmod.py
@@ -20,6 +20,20 @@ def test_check_modulefile_without_version(mock_module_cmd):
     mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
 test-module/1.0.0
 test-module/2.0.0
+test-module/3.0.1
+"""
+
+    # Expect an error raised
+    with pytest.raises(ValueError) as exc_info:
+        check_modulefile('test-module')
+        exc_info.value.startswith(
+            "There are multiple modules available for test-module"
+        )
+
+    # Mock module avail command use debug in name
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0
+test-module/1.0.0-debug
 """
 
     # Expect an error raised
@@ -35,6 +49,27 @@ def test_check_modulefile_exact_match(mock_module_cmd):
     # Mock module avail command
     mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
 test-module/1.0.0
+test-module/1.0.0-debug
+"""
+
+    # Test runs without an error
+    check_modulefile('test-module/1.0.0')
+
+
+@patch('payu.envmod.run_module_cmd')
+def test_check_modulefile_exact_match_with_symbolic_version(mock_module_cmd):
+    # Mock module avail command
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0(default)
+test-module/1.0.0-debug
+"""
+
+    # Test runs without an error
+    check_modulefile('test-module/1.0.0')
+
+    # Rerun test with another symbolic version/alias other than default
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0(some_symbolic_name_or_alias)
 test-module/1.0.0-debug
 """
 
--- a/test/test_envmod.py
+++ b/test/test_envmod.py
@@ -1,0 +1,70 @@
+import pytest
+from unittest.mock import patch
+
+from payu.envmod import check_modulefile
+
+
+@patch('payu.envmod.run_module_cmd')
+def test_check_modulefile_unique(mock_module_cmd):
+    # Mock module avail command
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0
+"""
+    # Test runs without an error
+    check_modulefile('test-module/1.0.0')
+
+
+@patch('payu.envmod.run_module_cmd')
+def test_check_modulefile_without_version(mock_module_cmd):
+    # Mock module avail command
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0
+test-module/2.0.0
+"""
+
+    # Expect an error raised
+    with pytest.raises(ValueError) as exc_info:
+        check_modulefile('test-module')
+        exc_info.value.startswith(
+            "There are multiple modules available for test-module"
+        )
+
+
+@patch('payu.envmod.run_module_cmd')
+def test_check_modulefile_exact_match(mock_module_cmd):
+    # Mock module avail command
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0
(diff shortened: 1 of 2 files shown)


test/test_envmod.py
```diff
@@ -0,0 +1,70 @@
...
+            "There are multiple modules available for test-module"
+        )
+
+
+@patch('payu.envmod.run_module_cmd')
+def test_check_modulefile_exact_match(mock_module_cmd):
+    # Mock module avail command
+    mock_module_cmd.return_value.stderr = """/path/to/modulefiles:
+test-module/1.0.0
+test-module/1.0.0-debug
```
aidanheerdegen, line 38: We're not testing that this runs without error if the version is specified. The previous test uses different input.

Just saying logically we haven't tested the negation of this.

test/test_envmod.py
aidanheerdegen, line 1: How do we deal with `default` versions, e.g. 
```
openmpi/4.1.3
openmpi/4.1.3-debug
openmpi/4.1.4(default)
openmpi/4.1.4-debug
openmpi/4.1.5
openmpi/4.1.5-debug
```
Would `check_modulefile('openpi/4.1.1')` pass? 

Note `module` adds the `(default)` modifier when that is the designated default version if no version is specified.

aidanheerdegen (approved):
Some comments ... but happy to merge if they're ignorable.

aidanheerdegen:
> last week there was an issue with multiple modules found in /g/data/vk83/modules and /g/data/vk83/modules/access-models, even though they were the same modulefile

That was an implementation bug. We've removed the `access-models` subdirectory to fix it. So it isn't a use case we need to care about or support. Does that help?

test/test_envmod.py
jo-basevi: Good point, it currently did not deal with `(default)`.

The module docs for the `4.3.0` version which is on NCI (https://modules.readthedocs.io/en/v4.3.0/module.html) has the following notes for `module avail`:

"Symbolic version-names and aliases found in the search are displayed in the result of this sub-command. Symbolic version-names are displayed next to the modulefile they are assigned to within parenthesis. Aliases are listed in the MODULEPATH section where they have been defined. To distinguish aliases from modulefiles a @ symbol is added within parenthesis next to their name. Aliases defined through a global or user specific module RC file are listed under the global/user modulerc section."

I couldn't find any other version or aliases other than `(default)` on quick search through the general NCI modules, but I've added in a line to remove pararenthesis  and contents from the end of string if it exists and some more tests, in bfecc12.

test/test_envmod.py
jo-basevi: Sorry I am a bit confused on what you mean. Do you mean testing there's an error when no version is specified, when there is following output?
```
test-module/1.0.0
test-module/1.0.0-debug
```
If so, I've added to the previous test.

jo-basevi:
Thanks for the review! 

> That was an implementation bug. We've removed the access-models subdirectory to fix it. So it isn't a use case we need to care about or support. Does that help?

Yes, that does help thanks. Ok so don't have to worry about using `module paths` or changing the implementation in that case.

test/test_envmod.py
aidanheerdegen: the `hh5` conda environments have some interesting ones (including multiple aliases)
```
conda/analysis3-24.01(analysis:analysis3:default) 
conda/analysis3-24.04(analysis3-unstable)
```
We explicitly *don't* currently support these environments, but just FYI.

aidanheerdegen (approved):
LGTM.

test/test_envmod.py
aidanheerdegen: Sorry for not being clear. Yes that is what I was saying. There was a previous test but it wasn't exactly the same as this, so logically it wasn't clear that the test showing it ran without error did so because the version was specified, as it didn't have a matching fail state when the version wasn't specified.

Does that make sense?

aidanheerdegen: It was a bit nit-picky.

jo-basevi: Yes that makes sense thanks. It was good to add a test to make the expected behaviour more clear anyway

test/test_envmod.py
jo-basevi: Thanks, that is good to know. At least can now module load those environments (if it's ever needed)

Issue #470, opened 2024-07-29
Drivers lacking documentation


Whyborn:
The respective drivers lack documentation, requiring assumed knowledge to configure specific models. Require users to contribute an example experiment for the commonly used drivers to demonstrate features specific to that driver?

aidanheerdegen:
Agreed. It would be good to have a first cab off the rank to serve as an exemplar before attempting this for other drivers/models.

Any volunteers? I could have a crack at MOM5 in the first instance, but I really don't know when I could get around to it.

PR #483, opened 2024-08-13
Enable seperate ice_history.nml & cice_in.nml settings

anton-seaice:
To enable simpler changes to the icefields namelist in cice, this change adds support for an `ice_history.nml` file which is spliced into the normal cice runtime settings (`cice_in.nml`) by payu before the model is run.

This follow the discussion in https://forum.access-hive.org.au/t/access-esm1-5-release-work-plan/1721/24

The contents of ice_history.nml can be anything which also would fit in cice_in.nml , but the intention is to move &icefields_nml to the ice_history file

Do we need to update docs ? Or tests ? There doesn't appear to be any of either at this point

### DIFF
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -158,7 +158,7 @@ class Cice(Model):
     def setup(self):
         super(Cice, self).setup()
 
-        # If there is a seperate ice_history.nml, 
+        # If there is a seperate ice_history.nml,
         # update the cice namelist with its contents
         history_nml_fpath = os.path.join(self.control_path,
                                          self.history_nml_fname)
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -254,10 +254,6 @@ class Cice(Model):
         assert(total_runtime % setup_nml['dt'] == 0)
         setup_nml['istep0'] = int(total_runtime / setup_nml['dt'])
 
-        if self.model_type != 'cice':  # model_type==cice5
-            # Force creation of a dump (restart) file at end of run
-            setup_nml['dump_last'] = True
-
         nml_path = os.path.join(self.work_path, self.ice_nml_fname)
         self.ice_in.write(nml_path, force=True)
 
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -247,6 +247,7 @@ class Cice(Model):
         # Force creation of a dump (restart) file at end of run
         setup_nml['dump_last'] = True
 
+        # Write the prepared cice namelist into the work directory
         nml_path = os.path.join(self.work_path, self.ice_nml_fname)
         self.ice_in.write(nml_path, force=True)
 
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -45,7 +45,7 @@ class Cice(Model):
 
         self.ice_nml_fname = 'cice_in.nml'
 
-        self.history_nml_fname = 'ice_history.nml' #only used by payu
+        self.history_nml_fname = 'ice_history.nml'  # only used by payu
 
         self.set_timestep = self.set_local_timestep
 
@@ -158,7 +158,8 @@ class Cice(Model):
     def setup(self):
         super(Cice, self).setup()
 
-        #  If there is a seperate ice_history.nml, update the cice namelist with its contents
+        # If there is a seperate ice_history.nml, 
+        # update the cice namelist with its contents
         history_nml_fpath = os.path.join(self.control_path,
                                          self.history_nml_fname)
         if os.path.isfile(history_nml_fpath):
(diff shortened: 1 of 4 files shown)


anton-seaice:
@aidanheerdegen @blimlim - This is ready for review.

To test, make a `ice/ice_history.nml` file and put some valid namelist values in it (move something from `ice/cice_in.nml`) and the payu run.

aidanheerdegen:
> Do we need to update docs ? Or tests ? There doesn't appear to be any of either at this point

That would be great.

We have an issue about the lack of driver docs

https://github.com/payu-org/payu/issues/470

And there isn't an example to leverage off at this point I'm afraid.

But just adding some tests for this functionality in a driver test suite would be fantastic. There are some examples you could base that on here

https://github.com/payu-org/payu/blob/master/test/models/

anton-seaice:
The expected behaviour for this change is :

- ice_history.nml overrides and adds to any values also in cice_in.nml
- there is no change if ice_history.nml doesn't exist
- ice_history.nml is not moved to output* folders, as its content is in cice_in.nml

anton-seaice:
> But just adding some tests for this functionality in a driver test suite would be fantastic. There are some examples you could base that on here

I think that would require splitting the new lines in a function ? e.g. patch_nml(self) 

It feels a bit like overkill but maybe its sensible ?

payu/models/cice.py
```diff
@@ -244,8 +254,9 @@ def setup(self):
         assert(total_runtime % setup_nml['dt'] == 0)
         setup_nml['istep0'] = int(total_runtime / setup_nml['dt'])
 
-        # Force creation of a dump (restart) file at end of run
-        setup_nml['dump_last'] = True
+        if self.model_type != 'cice':  # model_type==cice5
+            # Force creation of a dump (restart) file at end of run
+            setup_nml['dump_last'] = True
```
blimlim, line 259: Would another option be to do this from within the `cice5.py` driver? I'm not sure what would be better, as that would require a separate read/write of the namelist.

blimlim (dismissed):
Awesome, really glad that this is working! 

Changes look good to me. I've just added a comment for an optional change, but am not sure that it's necessarily better. 

Running on gadi with the history parameters separated into a separate namelist worked for me.

payu/models/cice.py
```diff
@@ -45,6 +45,8 @@ def __init__(self, expt, name, config):
 
         self.ice_nml_fname = 'cice_in.nml'
 
+        self.history_nml_fname = 'ice_history.nml'  # only used by payu
```
aidanheerdegen, line 48: We should probably add this to `self.optional_config_files` so that it is automatically added to the `git` repo. It will also result in the config file being copied to the `work` directory, and archived to `outputXXX`. Arguably this step isn't necessary, as it has already been patched into the main namelist file, but I don't see much harm in it.

payu/models/cice.py
aidanheerdegen: I'm not convinced we should do this in the driver. Seems like overreach and is impossible for a user to override if done at this point.

We can add something like this to QA checks for NRI released configurations. I'd even prefer making this the default in the model because it can then be overridden if necessary.

aidanheerdegen (changes requested):
I have an opinion or two ...

payu/models/cice.py
aidanheerdegen: P.S. I realise that code change would be in `cice5`, and therefore not directly related to this PR. I am also not advocating for that code change immediately, but would support it happening _eventually_

anton-seaice: The `cice5` driver runs this `setup` function from the `cice` Class entirely. This change to the namelist happens whilst the cice setup function is making other changes to the cice namelist. Its hard to do this in the cice5 driver because it would require either re-opening the cice nml and changing it, and writing that. Or refactoring this `setup` function, such that the cice5 setup function can do additional stuff between the namelist being opened and closed. Or adding an optional argument to the cice setup function with a dictionary of additional changes.

 Sounds like a new issue aidan :)

payu/models/cice.py
anton-seaice: Its possible for other parts of the payu driver to overwrite settings in `ice_history.nml` (e.g. you can put all of `cice_in.nml` in it, including say restart times ) and as `cice_in.nml` is the one loaded by the model its potentially confusing?

Happy to go either way I think.

payu/models/cice.py
anton-seaice: By new issue - I assume you were talking about setting `dump_last` at all in payu, rather than the new if statement.

anton-seaice:
I added a test, and moved the 'dump_last' to the cice5 driver. I didn't add `ice_history.nml` to optional_config_files because I think its potentially confusing (cice_in.nml is the source of truth).

payu/models/cice.py
aidanheerdegen: Fair enough. Let's leave it out and leave it up to config QA testing to take care of these things.

payu/models/cice.py
aidanheerdegen: Correct. As you said, a different issue entirely.

aidanheerdegen (approved):
LGTM

Issue #466, opened 2024-07-24
Year 400 crashes in ESM1.5 simulations


blimlim:
This relates to discussion in #457.

### Background:
Researchers have had problems with long ESM1.5 paleo simulationscrashing in calendar year 400. See [here](https://forum.access-hive.org.au/t/payu-model-exited-with-error-code-1-aborting-stuck-at-a-leap-year/1038) and [here](https://forum.access-hive.org.au/t/payu-potentially-miscalculating-a-leap-year/986/4). In the examples, the coupler and sea ice model appear to think there are only 365 days in the year while the ocean and atmosphere use the correct 366 days, leading to the crash, and in the first example, the ice model thinks its at year 300.

The disagreement between the ice/namcouple and the other components doesn't occur in every simulation that reaches year 400 though. E.g. attempting to reproduce the error by branching from the CSIRO pre-industrial run at year 400, and setting it to run for 3 months, payu gives the `namcouple` file the correct leap year run length of 91 days.

After working through some of Himadri's simulations, it looks like the calendar mismatch comes from payu's  start date calculation for the cice submodel. It pulls in information from both the control directory and the restart directory, meaning that if you copy a restart directory across different experiments, you could end up with inconsistent start dates. I still find it a bit confusing, so I hope the following explanation makes some sense.

### How payu sets the cice and namcouple run lengths/start dates:
The cice start date and run length are (mostly) calculated in the [`access.py` driver](https://github.com/payu-org/payu/blob/master/payu/models/access.py). It uses the `<CONTROL-DIRECTORY>/ice/input_ice.nml` namelist, which for example in our pre-industrial configuration looks like 

```fortran
<CONTROL-DIRECTORY>/ice/input_ice.nml
---------------------------------------------
&coupling
 *** caltype=1 ***
 jobnum=2
 inidate=01010101
 *** init_date=00010101 ***
 runtime0=3155673600
 runtime=86400
...
```
in addition to `<RESTART-DIRECTORY>/ice/input_ice.nml`, which for the pre-industrial configuration looks like 
```fortran
<RESTART-DIRECTORY>/ice/input_ice.nml
---------------------------------------------
&coupling
 *** runtime0=3155673600 ***
 *** runtime=0 ***
```
I've highlighted the variables that `access.py` uses in for the calculation with the `*` symbols. The other variables are ignored.

[...]

Issue #305, opened 2021-06-02
ACCESS driver does not support years > 9999


aidanheerdegen:
The `cice` model when used in the`access` driver does not support years > 9999 because it relies on python `datetime.datetime` which has a maximum year of 9999

This limitation was first discovered in `accessom2`

https://github.com/payu-org/payu/issues/303

As that driver does not require this calendar manipulation a fix was put in place to remove that logic from `accessom2` runs.

The `access` driver is used by `access-esm`, so this limit will apply to any of those runs until it is fixed.

PR #484, opened 2024-08-13
Replace CICE start date calculations 

blimlim:
This PR covers the first part of #466. It tries to simplify the setting of CICE's start date in ACCESS-ESM1.5 simulations, in order to avoid inconsistency bugs and to make the restart timing more transparent to the user. 

The previous calculations for setting the cice start date used:
``` 
control_dir/ice/input_ice.nml
--------------------------------
&coupling
init_date=00010101 (control_init_date)
...
```
and 
``` 
restart_dir/ice/input_ice.nml
--------------------------------
&coupling
runtime0=3155673600 (res_runtime0)
runtime=0 (res_runtime)
/
``` 

in order to set the start date of a new simulation, payu would calculate a start date as `inidate = control_init_date + res_runtime0 + res_runtime` (with required unit conversions).

It would then populate the work directory copy of `input_ice.nml` with values for the model to use for the new simulation:
``` 
work_dir/ice/input_ice.nml
--------------------------------
&coupling
inidate = control_init_date + res_runtime0 + res_runtime
init_date = control_init_date
runtime0= res_runtime0 + res_runtime
runtime= run duration calculated from config.yaml
/
```

---
This PR replaces  `restart_dir/ice/input_ice.nml` with a new file `restart_dir/ice/restart_date.nml` containing:
```
restart_dir/ice/restart_date.nml
---------------------------------
&coupling
init_date=10101 (res_init_date)
inidate=1010101 (res_inidate)
/
```
During setup, payu then fills the work directory copy of `input_ice.nml` with:

``` 
work_dir/ice/input_ice.nml
--------------------------------
&coupling
inidate = res_inidate
init_date = res_init_date
runtime0 = res_inidate - res_init_date (in seconds)
runtime = run duration calculated from config.yaml
/
```

My goal here was to ensure that cice receives the exact same values of `init_date`, `inidate`, `runtime`, `runtime0` with the new calculations that it would have received with the old calculations, which is why the updates use the `init_date` and calculate the number of seconds between it and the `inidate`. It would be simpler to just set both `init_date` and `inidate` to the new start date, and set `runtime0=0`, however this would give different information to `cice`.

For some reason I thought it would be good to read the `init_date` from the new restart file (I forget why...), though this could be moved back to reading it from the control directory if everyone thinks it is cleaner. 

Backwards compatibility will be broken with this PR, and if merged, we will have to add `restart_date.nml` files to the `vk83` restart directories.

Future work would making a separate function for reading the start date, which could then be used for consistency checks between the different component's start dates, though I'll defer that to a future PR/issue. 

---
For the review, any suggestions on structure, clarity, and also correctness of the calculations would be very welcome!

### DIFF
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -22,6 +22,69 @@ def int_to_date(date):
 def date_to_int(date):
     return date.year * 10**4 + date.month * 10**2 + date.day
 
+# TODO: The caltype logic could probably be simplified if we switched
+# to using just a string as the caltype input. Might require reworking of other 
+# functions for consistency.
+def seconds_between_dates(start_date, end_date, caltype_int):
+    """
+    Calculate the number of seconds between two datetime objects
+    with a specified calender type by using cftime datetime objects
+    as intermiaries.
+
+    Parameters
+    ----------
+    start_date: datetime.date 
+    end_date: datetime.date
+    caltype: Integer, either GREGORIAN or NOLEAP
+
+    Returns
+    -------
+    seconds: Number of seconds between start_date and end_date.
+    """
+    # Get the cftime string corresponding to the caltype integer
+    
+    #TODO: I think it's confusing that GREGORIAN means proleptic gregorian
+    if caltype_int == GREGORIAN:
+        calendar_str = "proleptic_gregorian"
+    elif caltype_int == NOLEAP:
+        calendar_str = "noleap"
+    else:
+        raise ValueError(f"Unrecognized caltype integer {caltype_int}")
+
+    delta = datetime_to_cftime(end_date) - datetime_to_cftime(start_date)
+
+    return delta.total_seconds()
+        
+
+
+# TODO: Is there an inbuilt way to do this?
+def datetime_to_cftime(date, calendar):
+    """
+    Convert a datetime.datetime object to a cftime.datetime object which 
+    has the same year, month, day, hour, minute, second values. 
+
+    Parameters
+    ----------
+    date: datetime.date object
+    calendar: string specifying a valid cftime calendar type
+
+    Returns
+    -------
+    date_cf: cftime.datetime object.
+    """
+    date_cf = cftime.datetime(
+        year = date.year,
+        month = date.month,
+        day = date.day,
+        hour = date.hour,
+        minute = date.minute,
+        second = date.second,
+        calendar = calendar
+    )
+
+    return date_cf
+
+
 
 def runtime_from_date(start_date, years, months, days, seconds, caltype):
     """
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -90,7 +90,7 @@ def calculate_leapdays(init_date, final_date):
     return datetime.timedelta(days=leap_days)
 
 
-# TODO: The caltype logic could probably be simplified if we switched
+# TODO: The caltype logic could be simplified if we switched
 # to using just a string as the caltype input. Might require reworking of other 
 # functions for consistency.
 def seconds_between_dates(start_date, end_date, caltype_int):
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -22,69 +22,6 @@ def int_to_date(date):
 def date_to_int(date):
     return date.year * 10**4 + date.month * 10**2 + date.day
 
-# TODO: The caltype logic could probably be simplified if we switched
-# to using just a string as the caltype input. Might require reworking of other 
-# functions for consistency.
-def seconds_between_dates(start_date, end_date, caltype_int):
-    """
-    Calculate the number of seconds between two datetime objects
-    with a specified calender type by using cftime datetime objects
-    as intermiaries.
-
-    Parameters
-    ----------
-    start_date: datetime.date 
-    end_date: datetime.date
-    caltype: Integer, either GREGORIAN or NOLEAP
-
-    Returns
-    -------
-    seconds: Number of seconds between start_date and end_date.
-    """
-    # Get the cftime string corresponding to the caltype integer
-    
-    #TODO: I think it's confusing that GREGORIAN means proleptic gregorian
-    if caltype_int == GREGORIAN:
-        calendar_str = "proleptic_gregorian"
-    elif caltype_int == NOLEAP:
-        calendar_str = "noleap"
-    else:
-        raise ValueError(f"Unrecognized caltype integer {caltype_int}")
-
-    delta = datetime_to_cftime(end_date) - datetime_to_cftime(start_date)
-
-    return delta.total_seconds()
-        
-
-
-# TODO: Is there an inbuilt way to do this?
-def datetime_to_cftime(date, calendar):
-    """
-    Convert a datetime.datetime object to a cftime.datetime object which 
-    has the same year, month, day, hour, minute, second values. 
-
-    Parameters
-    ----------
-    date: datetime.date object
-    calendar: string specifying a valid cftime calendar type
-
-    Returns
-    -------
-    date_cf: cftime.datetime object.
-    """
-    date_cf = cftime.datetime(
-        year = date.year,
-        month = date.month,
-        day = date.day,
-        hour = date.hour,
-        minute = date.minute,
-        second = date.second,
-        calendar = calendar
-    )
-
-    return date_cf
-
-
 
 def runtime_from_date(start_date, years, months, days, seconds, caltype):
     """
@@ -151,6 +88,68 @@ def calculate_leapdays(init_date, final_date):
     # TODO: Internal date correction (e.g. init_date is 1-March or later)
 
     return datetime.timedelta(days=leap_days)
+
+
+# TODO: The caltype logic could probably be simplified if we switched
+# to using just a string as the caltype input. Might require reworking of other 
+# functions for consistency.
+def seconds_between_dates(start_date, end_date, caltype_int):
+    """
+    Calculate the number of seconds between two datetime objects
+    with a specified calender type by using cftime datetime objects
+    as intermiaries.
+
+    Parameters
+    ----------
+    start_date: datetime.date 
+    end_date: datetime.date
+    caltype: Integer, either GREGORIAN or NOLEAP
+
+    Returns
+    -------
+    seconds: Number of seconds between start_date and end_date.
+    """
+    # Get the cftime string corresponding to the caltype integer
+    
+    #TODO: I think it's confusing that GREGORIAN means proleptic gregorian
+    if caltype_int == GREGORIAN:
+        calendar_str = "proleptic_gregorian"
+    elif caltype_int == NOLEAP:
+        calendar_str = "noleap"
+    else:
+        raise ValueError(f"Unrecognized caltype integer {caltype_int}")
+
+    delta = datetime_to_cftime(end_date, calendar_str) - datetime_to_cftime(start_date, calendar_str)
+
+    return delta.total_seconds()
+        
+
+# TODO: Is there an inbuilt way to do this?
+def datetime_to_cftime(date, calendar):
+    """
+    Convert a datetime.datetime object to a cftime.datetime object which 
+    has the same year, month, day, hour, minute, second values. 
+
+    Parameters
+    ----------
+    date: datetime.date object
+    calendar: string specifying a valid cftime calendar type
+
+    Returns
+    -------
+    date_cf: cftime.datetime object.
+    """
+    date_cf = cftime.datetime(
+        year = date.year,
+        month = date.month,
+        day = date.day,
+        hour = date.hour,
+        minute = date.minute,
+        second = date.second,
+        calendar = calendar
+    )
+
+    return date_cf
 
 
 def add_year_start_offset_to_datetime(initial_dt, n):
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -111,7 +111,7 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     """
     # Get the cftime string corresponding to the caltype integer
     
-    #TODO: I think it's confusing that GREGORIAN means proleptic gregorian
+    #TODO: Is it confusing that GREGORIAN means proleptic gregorian?
     if caltype_int == GREGORIAN:
         calendar_str = "proleptic_gregorian"
     elif caltype_int == NOLEAP:
@@ -124,7 +124,6 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     return delta.total_seconds()
         
 
-# TODO: Is there an inbuilt way to do this?
 def datetime_to_cftime(date, calendar):
     """
     Convert a datetime.datetime object to a cftime.datetime object which 
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -119,12 +119,12 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     else:
         raise ValueError(f"Unrecognized caltype integer {caltype_int}")
 
-    delta = datetime_to_cftime(end_date, calendar_str) - datetime_to_cftime(start_date, calendar_str)
+    delta = date_to_cftime(end_date, calendar_str) - date_to_cftime(start_date, calendar_str)
 
     return delta.total_seconds()
         
 
-def datetime_to_cftime(date, calendar):
+def date_to_cftime(date, calendar):
     """
     Convert a datetime.datetime object to a cftime.datetime object which 
     has the same year, month, day, hour, minute, second values. 
@@ -142,9 +142,9 @@ def datetime_to_cftime(date, calendar):
         year = date.year,
         month = date.month,
         day = date.day,
-        hour = date.hour,
-        minute = date.minute,
-        second = date.second,
+        hour = 0,
+        minute = 0,
+        second = 0,
         calendar = calendar
     )
 
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -122,7 +122,7 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     delta = (date_to_cftime(end_date, calendar_str)
              - date_to_cftime(start_date, calendar_str))
 
-    return delta.total_seconds()
+    return int(delta.total_seconds())
 
 
 def date_to_cftime(date, calendar):
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -101,7 +101,7 @@ def seconds_between_dates(start_date, end_date, caltype_int):
 
     Parameters
     ----------
-    start_date: datetime.date 
+    start_date: datetime.date
     end_date: datetime.date
     caltype: Integer, either GREGORIAN or NOLEAP
 
@@ -127,8 +127,8 @@ def seconds_between_dates(start_date, end_date, caltype_int):
 
 def date_to_cftime(date, calendar):
     """
-    Convert a datetime.datetime object to a cftime.datetime object which 
-    has the same year, month, day, hour, minute, second values. 
+    Convert a datetime.datetime object to a cftime.datetime object which
+    has the same year, month, day, hour, minute, second values.
 
     Parameters
     ----------
--- a/payu/calendar.py
+++ b/payu/calendar.py
@@ -91,7 +91,7 @@ def calculate_leapdays(init_date, final_date):
 
 
 # TODO: The caltype logic could be simplified if we switched
-# to using just a string as the caltype input. Might require reworking of other 
+# to using just a string as the caltype input. Might require reworking of other
 # functions for consistency.
 def seconds_between_dates(start_date, end_date, caltype_int):
     """
@@ -110,8 +110,8 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     seconds: Number of seconds between start_date and end_date.
     """
     # Get the cftime string corresponding to the caltype integer
-    
-    #TODO: Is it confusing that GREGORIAN means proleptic gregorian?
+
+    # TODO: Is it confusing that GREGORIAN means proleptic gregorian?
     if caltype_int == GREGORIAN:
         calendar_str = "proleptic_gregorian"
     elif caltype_int == NOLEAP:
@@ -119,10 +119,11 @@ def seconds_between_dates(start_date, end_date, caltype_int):
     else:
         raise ValueError(f"Unrecognized caltype integer {caltype_int}")
 
-    delta = date_to_cftime(end_date, calendar_str) - date_to_cftime(start_date, calendar_str)
+    delta = (date_to_cftime(end_date, calendar_str)
+             - date_to_cftime(start_date, calendar_str))
 
     return delta.total_seconds()
-        
+
 
 def date_to_cftime(date, calendar):
     """
@@ -139,13 +140,13 @@ def date_to_cftime(date, calendar):
     date_cf: cftime.datetime object.
     """
     date_cf = cftime.datetime(
-        year = date.year,
-        month = date.month,
-        day = date.day,
-        hour = 0,
-        minute = 0,
-        second = 0,
-        calendar = calendar
+        year=date.year,
+        month=date.month,
+        day=date.day,
+        hour=0,
+        minute=0,
+        second=0,
+        calendar=calendar
     )
 
     return date_cf
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -95,21 +95,16 @@ class Access(Model):
                 cpl_fpath = os.path.join(model.work_path, cpl_fname)
                 cpl_nml = f90nml.read(cpl_fpath)
 
-            
-
                 # Which calendar are we using, noleap or Gregorian.
                 caltype = cpl_nml[cpl_group]['caltype']
 
-                # Get time info about the beginning of this run. We're
-                # interested in:
-                #   1. start date of run
-                #   2. total runtime of all previous runs.
+                # Get timing information for the new run.
                 if model.prior_restart_path and not self.expt.repeat_run:
-                    
-                    # Read the start date from the restart date namelist
+                    start_date_nml_name = "start_date.nml"
+                    # Read the start date from the restart date namelist.
                     start_date_fpath = os.path.join(
                         model.prior_restart_path,
-                        "start_date.nml"
+                        start_date_nml_name
                     )
 
                     try:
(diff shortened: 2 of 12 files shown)


blimlim:
Oh my bad. I'll fix these PEP issues and ping everyone when done

payu/calendar.py
```diff
@@ -90,6 +90,68 @@ def calculate_leapdays(init_date, final_date):
...
+        raise ValueError(f"Unrecognized caltype integer {caltype_int}")
+
+    delta = (date_to_cftime(end_date, calendar_str)
+             - date_to_cftime(start_date, calendar_str))
+
+    return delta.total_seconds()
+
+
+def date_to_cftime(date, calendar):
+    """
```
blimlim, line 129: Let me know if you are aware of any built in ways to do this – I couldn't find any from a quick search.

payu/models/access.py
```diff
@@ -52,13 +52,24 @@ def __init__(self, expt, name, config):
...
+                model.runtime0_key = 'truntime0'
+                model.start_date_nml_name = "restart_date.nml"
+
     def setup(self):
         if not self.top_level_model:
             return
 
-        cpl_keys = {'cice': ('input_ice.nml', 'coupling', 'runtime0'),
-                    'matm': ('input_atm.nml', 'coupling', 'truntime0')}
-
```
blimlim, line 61: I moved these from here to the initialisation section, making them model attributes so that they would be accessible when writing the new restart date during `archive`.

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
             if model.model_type in ('cice', 'matm'):
```
blimlim, line 101: The previous calculations were identical for both `cice` and `matm`. It was simplest to make the same changes for both rather than keep all the old code for `matm` and add the new code separately for `cice`. This means that anyone who wants to run with `matm` would need to add the new restart file. Would this be ok to impose?

blimlim (commented):
Just added comments with background on a couple of the choices I made.

payu/calendar.py
```diff
@@ -90,6 +90,68 @@ def calculate_leapdays(init_date, final_date):
...
+    -------
+    seconds: Number of seconds between start_date and end_date.
+    """
+    # Get the cftime string corresponding to the caltype integer
+
+    # TODO: Is it confusing that GREGORIAN means proleptic gregorian?
+    if caltype_int == GREGORIAN:
+        calendar_str = "proleptic_gregorian"
+    elif caltype_int == NOLEAP:
+        calendar_str = "noleap"
```
aidanheerdegen, line 118: If this is just a case change, then I'd just do a blanket `.lowercase()` at the beginning to ensure any calendar is correct.

payu/calendar.py
aidanheerdegen: Judging from this `xarray` function you an just subtract one `CFTime` object from another 

https://github.com/pydata/xarray/blob/main/xarray/core/resample_cftime.py#L443-L480

And yeah the code backs that up

https://github.com/Unidata/cftime/blob/master/src/cftime/_cftime.pyx#L1063-L1064

and subtraction just returns a vanilla `timedelta`

https://github.com/Unidata/cftime/blob/master/src/cftime/_cftime.pyx#L1596

payu/models/access.py
aidanheerdegen: I'm of the opinion we should drop support for `matm`.

I have made an issue to this effect

https://github.com/payu-org/payu/issues/486

But for the time being, I wouldn't worry about breaking `matm`. Or we can removed it and rebase this PR and it goes away.

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
...
-                                                           total_runtime,
-                                                           caltype)
+                        start_date_nml = f90nml.read(start_date_fpath)[
+                            model.cpl_group]
+                    except FileNotFoundError:
+                        print(
+                            "Missing restart date file for model "
+                            f"{model.model_type}",
+                            file=sys.stderr
+                        )
```
aidanheerdegen, line 128: Will this report the path that is supposed to exist? I think it is always good to tell the user what the expected behaviour was, which in this case that there be a file of a certain name in a particular location.

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
...
+                    # run_start_date must be after initialisation date
+                    if run_start_date < init_date:
+                        msg = (
+                            "Restart date 'inidate` in "
+                            f"{model.start_date_nml_name} must not be "
+                            "before initialisation date `init_date. "
+                            "Values provided: \n"
+                            f"inidate = {start_date_nml['inidate']}\n"
+                            f"init_date = {start_date_nml['init_date']}"
+                        )
```
aidanheerdegen, line 148: Yeah this is great info to feedback to the user.

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
...
+                    # initialisation and new run start date,
+                    # to use for the runtime0 field.
+                    previous_runtime = cal.seconds_between_dates(
+                        init_date,
+                        run_start_date,
+                        caltype
+                    )
 
                 else:
-                    total_runtime = 0
```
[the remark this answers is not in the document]
aidanheerdegen: Good idea, much clearer.

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
...
-                cpl_fpath = os.path.join(model.work_path, cpl_fname)
+                # cpl_nml is the coupling namelist copied from the control to
+                # work directory.
+                cpl_fpath = os.path.join(model.work_path, model.cpl_fname)
                 cpl_nml = f90nml.read(cpl_fpath)
 
                 # Which calendar are we using, noleap or Gregorian.
-                caltype = cpl_nml[cpl_group]['caltype']
-                init_date = cal.int_to_date(cpl_nml[cpl_group]['init_date'])
+                caltype = cpl_nml[model.cpl_group]['caltype']
```
aidanheerdegen, line 110:
```suggestion
                # Use variable to point directly to coupling namelist group
                cpl_group = cpl_nml[model.cpl_group]
                caltype = cpl_group['caltype']
```
If you want to can make a variable to point directly to the coupling namelist group, which would clean up the later sections which also set variables here. It's a pattern that used elsewhere in the code.

Just a suggestion, up to you if you think it improves or degrades readability.

test/test_calendar.py
```diff
@@ -150,3 +154,42 @@ def test_parse_date_offset_no_offset_magnitude():
...
+                366 * SEC_PER_DAY
+            ),
+            (
+                datetime.datetime(year=1, month=1, day=1),
+                datetime.datetime(year=1001, month=1, day=1),
+                GREGORIAN,
+                (1000 * 365 + 242) * SEC_PER_DAY
+            ),
+        ]
+)
```
aidanheerdegen, line 193: Great to add tests, thanks. Just a note that all of the tests use `month=1` and `day=1`, so probably worth changing a couple to something else.

payu/calendar.py
```diff
@@ -90,6 +90,68 @@ def calculate_leapdays(init_date, final_date):
...
+    end_date: datetime.date
+    caltype: Integer, either GREGORIAN or NOLEAP
+
+    Returns
+    -------
+    seconds: Number of seconds between start_date and end_date.
+    """
+    # Get the cftime string corresponding to the caltype integer
+
+    # TODO: Is it confusing that GREGORIAN means proleptic gregorian?
```
aidanheerdegen, line 114: Well it's confusing that we use `gregorian` when we mean `proleptic_gregorian` and so have to write little work-arounds like this.

aidanheerdegen (commented):
I'd really really really recommend adding some model driver tests, even just a minimal set that covers the expected behaviour of this code.

There are some examples you can base tests on here

https://github.com/payu-org/payu/blob/master/test/models/

I'd guess the one @jo-basevi wrote would be a good start

https://github.com/payu-org/payu/blob/master/test/models/test_mom_mixin.py

payu/models/access.py
```diff
@@ -89,61 +100,68 @@ def setup(self):
...
+
+                    # Experiment initialisation date
+                    init_date = cal.int_to_date(
+                        start_date_nml["init_date"]
+                    )
+
+                    # Start date of new run
+                    run_start_date = cal.int_to_date(start_date_nml["inidate"])
+
+                    # run_start_date must be after initialisation date
```
anton-seaice, line 139: Did you test this? Its possible this would just work ?

anton-seaice (commented):
[input_ice](https://github.com/ACCESS-NRI/access-esm1.5-configs/blame/dev-historical%2Bconcentrations/ice/input_ice.nml) has lots of fields unrelated to restarts in it, so i'm not sure why we are adding another file? Whats the purpose in keeping the timing information on the completed run seperately?

I assume we are changing how runtime0 is calculated
from
`runtime0= res_runtime0 + res_runtime`
to
`runtime0 = res_inidate - res_init_date (in seconds)`
so that the user can change `res_inidate` if they desire (this is the date the current run starts at)?.

> For some reason I thought it would be good to read the init_date from the new restart file (I forget why...), though this could be moved back to reading it from the control directory if everyone thinks it is cleaner.

This is something to do with the warm-start script ... if the config comes from one location and the restart comes from a different location, I don't know which one is sensible to get the restart date in? I guess it depends on the warm start script and also which location its set in other models ?? I'm not really sure why this isn't the only change needed actually ?

payu/calendar.py
blimlim: Very glad that `cftime` does this. I'm making use of this in the `seconds_between_dates` function:
```
  delta = (date_to_cftime(end_date, calendar_str)
             - date_to_cftime(start_date, calendar_str))
```
however am manually converting from `datetime.date` to `cftime.date` to do this calculation. This is because the `start_date` and `end_date`  need to be `datetime.date` objects for other calculations such as `runtime_from_date`.

payu/calendar.py
blimlim: This one's actually converting from the [integer constants]( https://github.com/ACCESS-NRI/payu/blob/8f18f245787cafa940a9c397b5173479a40b6273/payu/calendar.py#L7) used in the cice namelists to the calendar strings used by `cftime`.

payu/models/access.py
blimlim: Great idea! I think it makes it easier to read, and added this change in to the latest commit.

test/test_calendar.py
blimlim: Good idea. I've added in tests using different months and days, and also added a test running between year 1 and 9999, which are the minimum and maximum years for `datetime.date`s.

payu/models/access.py
blimlim: I tried running with 
```    
inidate = 1010101
init_date = 10101
```
and then during `payu setup` forced `runtime0=0`, to test whether it is sufficient to set just the `inidate`.

Some strange things end up happening, e.g. in `iceout085`:
```
(calendar)  idate =      1010101
(calendar)  idate =      2010101
(calendar)  idate =      2010101
```
the year jumped by 100 years, which also impacted the output file names:
```
iceh.0201-01.nc
```
Values like `runtime0` are [showing up](https://github.com/ACCESS-NRI/cice4/blob/e7549ebd2044690a432cc67c1317c81cb194b750/source/ice_restart.F90#L330) in the cice code, and so figured that a safe approach for the new calculations would be to ensure that cice is still provided with the same values of `inidate`, `init_date`, and `runtime0` as before.

blimlim:
> [input_ice](https://github.com/ACCESS-NRI/access-esm1.5-configs/blame/dev-historical%2Bconcentrations/ice/input_ice.nml) has lots of fields unrelated to restarts in it, so i'm not sure why we are adding another file? Whats the purpose in keeping the timing information on the completed run seperately?

Previously payu had been copying the whole `input_ice.nml`  into the restart directory, but only the `runtime0` and `runtime` fields from this copy were required. Meanwhile the `inidate` and `init_date` were being ignored, along with all the other timing related fields. The main idea behind the separate `restart_date.nml` file is to try and simplify this, so that the new file holds all the information needed for the run timing, and none of the unnecessary extra fields. Let me know if this approach sounds ok!

Related to this – with the changes, the `input_ice.nml` was not meant to be copied to the restart directory anymore as it isn't needed. I'd forgotten to implement this change, but have added it in to the latest commit.

> to
> `runtime0 = res_inidate - res_init_date (in seconds)`
> so that the user can change `res_inidate` if they desire (this is the date the current run starts at)?.

Unfortunately the restart date is also held by the binary `iced...` restart files, and it needs to be consistent with the `inidate`. This means that changing `res_inidate` won't be enough by itself to change to a completely new restart date. It will make it easier though to create a simulation from a CSIRO restart which currently requires the `runtime0+runtime` calculation to be performed if done manually. Another reason for the change is to try and make it clearer which date cice will be starting at (e.g. so people don't need to calculate `control_init_date + res_runtime0 + res_runtime').

> This is something to do with the warm-start script
That's a good point. If these changes go through, the warm start scripts would need to be updated down the line for them to work with the changed calculations.

> I'm not really sure why this isn't the only change needed actually ?

I think it would work to read the `init_date` from the restart directory instead of the control directory, and have this as the only change, though it might be  a bit less transparent which date the model will restart at. I.e. this would require `runtime_0 + runtime + init_date` calculation. Happy to hear from everyone if either approach is preferrable!

payu/calendar.py
aidanheerdegen: Ah duh! (Me). Sorry.
```suggestion
    elif caltype_int == NOLEAP:
        calendar_str = "noleap"
```
```suggestion
    elif caltype_int == NOLEAP:
        calendar_str = "noleap"
```
```suggestion
    elif caltype_int == NOLEAP:
        calendar_str = "noleap"
```
```suggestion
    elif caltype_int == NOLEAP:
        calendar_str = "noleap"
```

aidanheerdegen: Ah duh! (me). Sorry.

blimlim:
> I'd really really really recommend adding some model driver tests, even just a minimal set that covers the expected behaviour of this code.

Good idea. I've had a go at adding some tests, first of the [individual calendar functions](https://github.com/ACCESS-NRI/payu/blob/8f18f245787cafa940a9c397b5173479a40b6273/test/test_calendar.py#L219-L261) used in the calculations, and then a bigger test of the actual driver:
https://github.com/ACCESS-NRI/payu/blob/8f18f245787cafa940a9c397b5173479a40b6273/test/models/test_access.py#L178-L242

The driver test cycles over the `access.setup()` and `access.archive()` steps (currently 1000 times – perhaps excessive) with a year long runtime, and checks that the final end date matches what's expected. It's admittedly a bit slow (~22s for the 1000 cycles), but would be keen to hear if this sounds like a reasonable type of test to ensure that the calculations are working correctly across the different tricky years that can come up?

payu/models/access.py
anton-seaice: Thanks :)

test/test_calendar.py
```diff
@@ -150,3 +156,107 @@ def test_parse_date_offset_no_offset_magnitude():
...
+                datetime.datetime(year=300, month=1, day=1),
+                datetime.datetime(year=301, month=1, day=1),
+                GREGORIAN,
+                365 * SEC_PER_DAY
+            ),
+            (
+                datetime.datetime(year=400, month=1, day=1),
+                datetime.datetime(year=401, month=1, day=1),
+                GREGORIAN,
+                366 * SEC_PER_DAY
```
anton-seaice, line 186:
```suggestion
                datetime.datetime(year=400, month=1, day=1),
                datetime.datetime(year=401, month=12, day=31),
                GREGORIAN,
                365 * SEC_PER_DAY
```
This tests that month and day increments are working also

test/test_calendar.py
anton-seaice: Thanks - new test looks good :)

test/test_calendar.py
blimlim: Thanks great idea - added this one in!

payu/models/access.py
```diff
@@ -52,13 +52,24 @@ def __init__(self, expt, name, config):
             if model.model_type == 'cice5':
                 model.access_restarts.append(['u_star.nc', 'sicemass.nc'])
 
+            if model.model_type == 'cice':
+                # Structure of model coupling namelist
+                model.cpl_fname = 'input_ice.nml'
+                model.cpl_group = 'coupling'
+                model.runtime0_key = 'runtime0'
+                model.start_date_nml_name = "restart_date.nml"
```
anton-seaice, line 60: Should keys for "init_date", "inidate" and "runtime" be here rather than hardcoded at line 182 ?

Have we documented what each field means somewhere?

payu/models/access.py
```diff
@@ -208,13 +233,53 @@ def archive(self):
...
+                # namelist.
+
+                # Calculate the end date using information from the work
+                # directory coupling namelist.
+                work_cpl_fpath = os.path.join(model.work_path, model.cpl_fname)
+                work_cpl_nml = f90nml.read(work_cpl_fpath)
+                work_cpl_grp = work_cpl_nml[model.cpl_group]
+
+                # Timing information on the completed run.
+                run_init_date_int = work_cpl_grp["init_date"]
```
anton-seaice, line 260: Is it clearer to call this `exp_init_date...` ? Because its start date for the whole experiment ?

test/test_calendar.py
```diff
@@ -150,3 +156,107 @@ def test_parse_date_offset_no_offset_magnitude():
...
+def test_seconds_between_dates(start_date, end_date, caltype_int, expected):
+    assert seconds_between_dates(start_date, end_date, caltype_int) == expected
+
+
+@pytest.mark.parametrize(
+        "date_int, expected",
+        [
+            (10101, datetime.date(1, 1, 1)),
+            (100321, datetime.date(10, 3, 21)),
+            (38211130, datetime.date(3821, 11, 30))
```
anton-seaice, line 217:
```suggestion
            (99991231, datetime.date(9999, 12, 31))
```

test/test_calendar.py
```diff
@@ -150,3 +156,107 @@ def test_parse_date_offset_no_offset_magnitude():
...
+        ]
+)
+def test_int_to_date(date_int, expected):
+    """
+    Check that integers typically read in from namelists
+    are correctly converted to datetime.date objects.
+    """
+    converted_date = int_to_date(date_int)
+    assert converted_date == expected
+
```
anton-seaice, line 227: For completeness, this should also test the negative cases:

e.g. what if the year is 0 or year 10000, or the date requested doesn't exist (i.e. 999999)

test/test_calendar.py
```diff
@@ -150,3 +156,107 @@ def test_parse_date_offset_no_offset_magnitude():
...
+    Test that the number of seconds calculated for run lengths is correct.
+    """
+    runtime = runtime_from_date(start_date,
+                                years,
+                                months,
+                                days,
+                                seconds,
+                                caltype)
+
+    assert runtime == expected
```
anton-seaice, line 262: Again - we should test the negative cases ... year -1, month 13, etc

test/test_calendar.py
```diff
@@ -150,3 +156,107 @@ def test_parse_date_offset_no_offset_magnitude():
...
+            (datetime.date(40, 2, 8), 0, 1, 0, 0, NOLEAP, 28*SEC_PER_DAY),
+            (datetime.date(153, 8, 21), 2, 7, 510, 5321,
+             GREGORIAN, 1453*SEC_PER_DAY + 5321),
+            # Extreme cases, unlikely to ever happen
+            (datetime.date(1, 1, 1), 9998, 0, 0, 0,
+             NOLEAP, 9998 * 365 * SEC_PER_DAY),
+            (datetime.date(1, 1, 1), 9998, 0, 0, 0,
+             GREGORIAN, (9998 * 365 + 2424) * SEC_PER_DAY)
+        ]
+)
```
anton-seaice, line 243: I think this would be clearer if you just have the limits in all permutations:

For start_date: the minimum (e.g. datetime.date(1,1,1)) and the maximum (datetime.date(9999,12,31)) 
For run time: the minimum (1 second) and the maximums, (9999,0,0,0) and (9998,12,31,86400)
For calendar: GREGORIAN and NOLEAP

Its about 12 cases.

payu/models/access.py
```diff
@@ -157,20 +176,26 @@ def setup(self):
...
 
-                # Now write out new run start date and total runtime.
-                cpl_nml[cpl_group]['inidate'] = cal.date_to_int(run_start_date)
-                cpl_nml[cpl_group][runtime0_key] = total_runtime
-                cpl_nml[cpl_group]['runtime'] = int(run_runtime)
+                # Now write out new run start date and total runtime into the
+                # work directory namelist.
+                cpl_group["init_date"] = cal.date_to_int(init_date)
+                cpl_group['inidate'] = cal.date_to_int(run_start_date)
+                cpl_group[model.runtime0_key] = previous_runtime
```
anton-seaice, line 185: When I test this, my `input_ice.nml` gave `    runtime0 = 3155673600.0`

The extra `.0` probably doesn't matter , but is it easy to tidy up?

anton-seaice:
I ran this and it works with the dates look correct for the first couple of months of picontrol. I didn't test trying to change the start date or what happens with a leap year.

anton-seaice:
The tests run and pass as-is, so if you want to defer the suggested changes to the tests to after release, that would be ok

payu/models/access.py
blimlim: That's a good idea – makes it more consistent. I've added in these extra keys, and included comments briefly specifying what their meaning:

https://github.com/ACCESS-NRI/payu/blob/c19133c83109e768236486b0879d09e6f05c3be2/payu/models/access.py#L60-L67

payu/models/access.py
blimlim: Yep that's much clearer! I've added this in.

test/test_calendar.py
blimlim: Good idea, added this in

test/test_calendar.py
blimlim: Good question, I've added a separate test to check that the function fails on invalid dates:

https://github.com/ACCESS-NRI/payu/blob/c19133c83109e768236486b0879d09e6f05c3be2/test/test_calendar.py#L229-L239

test/test_calendar.py
blimlim: This is a good point – It turns out that this function currently doesn't guard against bad runtimes being specified in the `config.yaml`. If it's given a negative input it outputs a negative number of seconds for the runtime.

Since this is something that touches a couple other models, do you think it would be ok to defer adding guard rails to a separate issue?

payu/models/access.py
blimlim: Good catch, I've adjusted the `seconds_between_dates` function to output integers:
```
return int(delta.total_seconds())
```
which should be safe as it's inputs are `datetime.date` objects, and so there shouldn't be any fractions of seconds appearing anywhere.

test/test_calendar.py
blimlim: Good idea to test around the edges here. I think it's worth retaining the cases like `(datetime.date(101, 1, 1), 1, 0, 0, 0, GREGORIAN, 365*SEC_PER_DAY)`, as these match what people will generally use for ESM1.5, e.g. 1 year runtimes, starting at the beginning of the year. 

In terms of the max cases, start dates like `datetime.date(9999,12,31)` cause issues, as the `runtime_from_date` function internally creates an end date of `start date + runtime`, which then goes outside the limits of `datetime.date`. I've added in a few different limiting cases though – let me know if you think they cover what we need.

blimlim:
Thanks for the reviews! Lots of good ideas which help make the changes clearer and also improve the tests. I've added in the different suggestions, and also added an additional test of the `access.py` driver.

https://github.com/ACCESS-NRI/payu/blob/c19133c83109e768236486b0879d09e6f05c3be2/test/models/test_access.py#L261-L269

The previous test which cycled over the `setup` and `archive` stages would not catch cases where calculations in both steps were simultaneously wrong (e.g. if they somehow both used the wrong calendar). The new test focuses on the setup date, checking that the correct runtime is written to the namelists when starting at a range of leap/non-leap years. I think there's some redundancy with this and the more focused tests of `calendar.py` but think it's still helpful to have to make sure that everything comes together correctly in the driver.

payu/models/access.py
anton-seaice: Nice

payu/models/access.py
anton-seaice: In the text for the FileNoteFoundError if will give the expected path

test/models/test_access.py
```diff
@@ -0,0 +1,340 @@
...
+
+@pytest.mark.parametrize(
+        "start_date_int, caltype, expected_runtime",
+        [(1010101, GREGORIAN, 365*SEC_PER_DAY),
+         (1010101, NOLEAP, 365*SEC_PER_DAY),
+         (1040101, GREGORIAN, 366*SEC_PER_DAY),
+         (1040101, NOLEAP, 365*SEC_PER_DAY),
+         (3000101, GREGORIAN, 365*SEC_PER_DAY),
+         (3000101, NOLEAP, 365*SEC_PER_DAY),
+         (4000101, GREGORIAN, 366*SEC_PER_DAY),
```
anton-seaice, line 258:
```suggestion
         (4000101, GREGORIAN, 366*SEC_PER_DAY), #not a leap year!
```

test/models/test_access.py
```diff
@@ -0,0 +1,340 @@
...
+    fake_cice_in,
+    restart_dir,
+    initial_start_date_file,
+    start_date_int,
+    caltype,
+    expected_runtime
+):
+    """
+    The large setup/archive cycling test won't pick up situations 
+    where the calculations dyring setup and archive are simultaneously
```
anton-seaice, line 273:
```suggestion
    where the calculations during setup and archive are simultaneously
```

test/test_calendar.py
```diff
@@ -150,3 +156,137 @@ def test_parse_date_offset_no_offset_magnitude():
...
+            # Leap year
+            (datetime.date(4, 1, 1), 1, 0, 0, 0, GREGORIAN, 366*SEC_PER_DAY),
+            (datetime.date(4, 1, 1), 1, 0, 0, 0, NOLEAP, 365*SEC_PER_DAY),
+            # Non-leap year due to 100 year rule
+            (datetime.date(100, 1, 1), 1, 0, 0, 0, GREGORIAN, 365*SEC_PER_DAY),
+            (datetime.date(100, 1, 1), 1, 0, 0, 0, NOLEAP, 365*SEC_PER_DAY),
+            # Leap year due to 400 year rule
+            (datetime.date(400, 1, 1), 1, 0, 0, 0, GREGORIAN, 366*SEC_PER_DAY),
+            (datetime.date(500, 1, 1), 1, 0, 0, 0, NOLEAP, 365*SEC_PER_DAY),
+            # Febraury in leap years
```
anton-seaice, line 257:
```suggestion
            # February in leap years
```

test/test_calendar.py
anton-seaice: Im just going to go ahead and assume no paleo folk do a climate run longer than 9999 years !

test/test_calendar.py
```diff
@@ -150,3 +156,137 @@ def test_parse_date_offset_no_offset_magnitude():
...
+            (datetime.date(40, 2, 8), 0, 1, 0, 0, GREGORIAN, 29*SEC_PER_DAY),
+            (datetime.date(40, 2, 8), 0, 1, 0, 0, NOLEAP, 28*SEC_PER_DAY),
+            # Misc
+            (datetime.date(153, 8, 21), 2, 7, 510, 5321,
+             GREGORIAN, 1453*SEC_PER_DAY + 5321),
+            # Extreme cases, unlikely to ever happen
```
anton-seaice, line 263:
```suggestion
            # Max & min limits
```

test/test_calendar.py
```diff
@@ -150,3 +156,137 @@ def test_parse_date_offset_no_offset_magnitude():
...
+            (datetime.date(100, 1, 1), 1, 0, 0, 0, NOLEAP, 365*SEC_PER_DAY),
+            # Leap year due to 400 year rule
+            (datetime.date(400, 1, 1), 1, 0, 0, 0, GREGORIAN, 366*SEC_PER_DAY),
+            (datetime.date(500, 1, 1), 1, 0, 0, 0, NOLEAP, 365*SEC_PER_DAY),
+            # Febraury in leap years
+            (datetime.date(40, 2, 8), 0, 1, 0, 0, GREGORIAN, 29*SEC_PER_DAY),
+            (datetime.date(40, 2, 8), 0, 1, 0, 0, NOLEAP, 28*SEC_PER_DAY),
+            # Misc
+            (datetime.date(153, 8, 21), 2, 7, 510, 5321,
+             GREGORIAN, 1453*SEC_PER_DAY + 5321),
```
anton-seaice, line 262:
```suggestion
            (datetime.date(1, 1, 1), 0, 0, 0, 86400,
             GREGORIAN, 86400),
```

test/test_calendar.py
anton-seaice: This is much clearer, thankyou

test/test_calendar.py
```diff
@@ -150,3 +156,137 @@ def test_parse_date_offset_no_offset_magnitude():
...
+                datetime.datetime(year=23, month=3, day=15),
+                GREGORIAN,
+                (10 * 365 + 238) * SEC_PER_DAY
+            ),
+            (
+                datetime.datetime(year=1, month=1, day=1),
+                datetime.datetime(year=1001, month=1, day=1),
+                GREGORIAN,
+                (1000 * 365 + 242) * SEC_PER_DAY
+            ),
```
anton-seaice, line 199: These two cases are covered sufficiently by the other tests (unless I missed something)
```suggestion
```

test/test_calendar.py
anton-seaice: Turns out there was already an issue for the 9999 years thing ! https://github.com/payu-org/payu/issues/305

test/test_calendar.py
anton-seaice: Ok i raised https://github.com/payu-org/payu/issues/492

payu/models/access.py
anton-seaice: nice

test/models/test_access.py
anton-seaice: hangon this is a leap year ?

test/test_calendar.py
anton-seaice: Nice

anton-seaice:
This looks good @blimlim - I just had some tidyups on the testing in the suggestions :)

test/models/test_access.py
blimlim: Yeah this one should be a leap year!

test/models/test_access.py
blimlim: Ah thanks for spotting that!

test/test_calendar.py
blimlim: Good catch!

test/test_calendar.py
blimlim: I've kept in the first test which was based off Aidan's suggestion to test using some mid-month values, and removed the second one – I agree I think that one's covered by the other cases!

blimlim:
Awesome thanks for those suggestions – very helpful for making the code clearer and improving the tests! I've added those in and let me know if there are any other suggestions.

anton-seaice (approved):
Good work Spencer !

blimlim:
@aidanheerdegen are you happy for this one to be merged, or did you have any other suggestions?

PR #488, opened 2024-08-16
Add warnings for signing runlog commits

### LINKED ISSUE #449: Support for signing git commits
As brought up in this forum post - https://forum.access-hive.org.au/t/payu-runlog-and-git-commit-signing/2178, there can be errors when global sign commits is enabled on NCI, as `payu run` PBS job uses system git with an older openssh version than the library installed in `hh5` and `vk83` payu environments. 

This error can be worked around by using passphrase-less signing keys with a public/private key pair on NCI, and configuring git config to use an `ssh-keygen` executable with version >= 8.2p1. For example, configuring `git config --global gpg.ssh.program /g/data/hh5/public/apps/miniconda3/envs/analysis3/bin/ssh-keygen` when using the `hh5`'s conda environment, or `git config --global gpg.ssh.program /g/data/vk83/apps/payu/1.1.3/bin/ssh-keygen` when using `vk83`'s payu environments. 

To support commit signing, so it can work with signing keys that have been forwarded from the host machine, and keys that have a passphrase, it would be ideal if payu could commit runlog on a login node. This would require experiment setup to run before the pbs job submission to generate the manifests. Currently, the commits are added only after the model run command has been executed- this is good for ensuring the model run command actually ran and there weren't any errors prior to it. 

Other cons to running experiment setup on the login nodes, is that in the time elapsed between running setup and run job running, something in the configuration or work directory could change. Also, running experiment on the pbs job sets up checks that the compute job is able to access the various projects needed to run.. There's probably other reasons to keeping the experiment setup as part of the run job on the compute nodes that I am not aware of. 

Another option could if be if signing a commit errors out on a pbs job, then revert to just committing an unsigned commit?

jo-basevi:
Re-attempt to sign commits without signing with a warning, when signing runlog commits fails due to git config using:
- signing key with a passphrase or the private key is hosted on local machine
- or the gpg.ssh.program is not accessible on run pbs job - path is not added to storage flags

Closes #449 

I thought about adding a help message, something like
```
If signed runlog commits are required, either:
 - manually run 'git commit --amend' to resign the latest commit
 - change git config to a  use passphrase-less and local signing key pair,
    and check configured 'gpg.ssh.program' is accessible during PBS run job (check storage flags) 
```
Which I can add in if it'll be helpful - I just don't want to make it even more confusing..

### DIFF
--- a/payu/runlog.py
+++ b/payu/runlog.py
@@ -12,6 +12,7 @@ import os
 import shlex
 import subprocess
 import sys
+import warnings
 
 # Third party
 import requests
@@ -95,9 +96,18 @@ class Runlog(object):
         print(cmd)
         try:
             subprocess.check_call(shlex.split(cmd), stdout=f_null,
-                                  cwd=self.expt.control_path)
+                                  stderr=f_null, cwd=self.expt.control_path)
         except subprocess.CalledProcessError:
-            print('TODO: Check if commit is unchanged')
+            # Attempt commit without signing commits
+            cmd = f'git -c "commit.gpgsign=false" commit -am "{commit_msg}"'
+            print(cmd)
+            try:
+                subprocess.check_call(shlex.split(cmd),
+                                      stdout=f_null,
+                                      cwd=self.expt.control_path)
+                warnings.warn("Runlog commit was commited without git signing")
+            except subprocess.CalledProcessError:
+                warnings.warn("Error occured when attempting to commit runlog")
 
         # Save the commit hash
         self.expt.run_id = commit_hash(self.expt.control_path)


payu/runlog.py
```diff
@@ -95,9 +96,18 @@ def commit(self):
         print(cmd)
         try:
             subprocess.check_call(shlex.split(cmd), stdout=f_null,
-                                  cwd=self.expt.control_path)
+                                  stderr=f_null, cwd=self.expt.control_path)
         except subprocess.CalledProcessError:
-            print('TODO: Check if commit is unchanged')
+            # Attempt commit without signing commits
+            cmd = f'git -c "commit.gpgsign=false" commit -am "{commit_msg}"'
```
CodeGat, line 102: There is also a `--no-gpg-sign` flag for `git commit`, if you didn't want to modify config.

CodeGat (approved):

CodeGat (approved):
LGTM

payu/runlog.py
jo-basevi: Thanks! Yeah that makes the command a little more easier to read. I've added it in 55f594a and have tested that it works the same on a payu run job that errors with signing commits .

Issue #367, opened 2023-09-27
Assert is an anti-pattern in production code


aidanheerdegen:
payu is using `assert` in a number of locations throughout the production code

https://github.com/search?q=repo%3Apayu-org%2Fpayu%20assert&type=code

This is not good practice:

https://www.inspiredpython.com/article/assertions-and-exceptions-are-not-the-same

Note that it is fine in to use `assert` in tests.

PR #496, opened 2024-08-29
add checks of io layout to cmeps driver

anton-seaice:
This PR implements checks of the processors selected in the nuopc.runconfig file for the access-om3 payu driver.  This ensures that the processors requested for normal model operations and parallel IO are within range of the CPUs set in config.yaml and each model "realm" uses processors for IO that are within the range of processors for that realm.

It adds tests for the min/max bounds of each parameter.

Contributes to https://github.com/COSIMA/access-om3/issues/109

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -184,7 +184,7 @@ class CesmCmeps(Model):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
-    def _setup_checks(self) :
+    def _setup_checks(self):
         # check pelayout fits within requested cpucount
         cpucount = int(
             self.expt.config.get('ncpus', multiprocessing.cpu_count())
@@ -195,47 +195,51 @@ class CesmCmeps(Model):
             nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
             rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
             pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
-            npes = (ntasks-1)*nthreads*pestride #count to the last process, starting at 0
+            npes = (ntasks-1)*nthreads*pestride  #count to the last process, starting at 0
             if (rootpe + npes) > cpucount:
                 raise ValueError(
                     f"Insufficient cpus for the {realm} pelayout in {NUOPC_CONFIG}"
                 )
 
             # check iolayout
-            if realm == "cpl": 
+            if realm == "cpl":
                 comp = "MED"  # med and cpl names are both used in runconfig
             else :
                 comp = realm.upper()
             if comp in self.runconfig.get_component_list():
                 io_section = f"{comp}_modelio"
                 nc_type = self.runconfig.get(io_section, "pio_typename")
-                if nc_type == "netcdf4c":
-                    raise ValueError(
-                        f"netcdf4c in {io_section} of {NUOPC_CONFIG} is deprecated, "
-                        "use netcdf4p"
-                    )
+                ioroot = int(self.runconfig.get(io_section, "pio_root"))
 
-                # if nc_type is netcdf, only one pe is needed
-                ioroot = (self.runconfig.get(io_section, "pio_root"))
-                if ioroot is None:
-                    raise ValueError(f"{ioroot} not valid for ioroot for the {io_section}")
-                if int(ioroot) > npes:
+                if ioroot > npes:
                     raise ValueError(
                         f"Insufficient cpus for the {io_section} ioroot pe in "
                         f"{NUOPC_CONFIG}"
                     )
-                # To-do: add coverage for nc_type=pnetcdf, and pio_async == .true.
-                if all([
-                    nc_type == "netcdf4p", 
-                    self.runconfig.get(io_section, "pio_async_interface") == ".false."
-                    ]) :
-                        niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
-                        iostride = int(self.runconfig.get(io_section, "pio_stride"))
-                        if (int(ioroot) + (niotasks-1)*iostride) > npes:
-                            raise ValueError(
-                                f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
-                                "requesting out of range cpus"
-                            )
+
+                match nc_type:
+                    case "netcdf" :
+                        break
+                    case "netcdf4p" | "pnetcdf" :
+                        if self.runconfig.get(io_section, "pio_async_interface") == ".false." :
+                            niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
+                            iostride = int(self.runconfig.get(io_section, "pio_stride"))
+                            if (int(ioroot) + (niotasks-1)*iostride) > npes:
+                                raise ValueError(
+                                    f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
+                                    "requesting out of range cpus"
+                                )
+                        # To-do: add coverage for pio_async == .true.
+                    case "netcdf4c":
+                        raise ValueError(
+                            f"netcdf4c in {io_section} of {NUOPC_CONFIG} is deprecated, "
+                            "use netcdf4p"
+                        )
+                    case _ : 
+                        raise ValueError(
+                            f"The iotype for {io_section} in {NUOPC_CONFIG} is "
+                            'invalid, valid options are "netcdf", "pnetcdf" and "netcdf4p"'
+                        )
         return True
 
     def archive(self):
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -198,10 +198,10 @@ class CesmCmeps(Model):
             if nthreads < 1:
                 raise ValueError(f"The number of {realm}_nthreads ({nthreads}) in "
                                  f"{NUOPC_CONFIG} must be at least 1.")
-                                 
+                           
             if nthreads > 1:
                 npes = nthreads*ntasks*pestride
-                # this is taken from 
+                # this is taken from
                 # https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
                 # the correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
             else:
@@ -242,7 +242,17 @@ class CesmCmeps(Model):
                         case "netcdf4p" | "pnetcdf":
                             niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
                             iostride = int(self.runconfig.get(io_section, "pio_stride"))
-                            if (ioroot + (niotasks-1)*iostride) >= npes:
+                            if (niotasks<=0) : 
+                                warn(f"The pio_numiotasks for {io_section} in {NUOPC_CONFIG} is "
+                                    "not set, using model default")
+                            if (iostride<=0) : 
+                                warn(f"The pio_stride for {io_section} in {NUOPC_CONFIG} is "
+                                    "not set, using model default")
+                            if (all(
+                                    niotasks>0, 
+                                    iostride>0,
+                                    ioroot + (niotasks-1)*iostride) >= npes
+                                ):
                                 raise ValueError(
                                     f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
                                     "requesting out of range cpus"
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -163,7 +163,7 @@ class CesmCmeps(Model):
             ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
             rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
             pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
-            #rootpe is zero-based
+            # rootpe is zero-based
             if cpucount < (rootpe + ntasks*pestride):
                 raise ValueError(
                     f"Insufficient cpus for the {realm} pelayout in nuopc.runconfig"
@@ -172,26 +172,30 @@ class CesmCmeps(Model):
         # check iolayout
         for realm in self.realms:
             # med and cpl names are both used in runconfig
-            if realm == "cpl": realm="MED"
+            if realm == "cpl": 
+                realm = "MED"
             io_section = f"{realm.upper()}_modelio"
             nc_type = self.runconfig.get(io_section, "pio_typename")
             if nc_type == "netcdf4c":
                 raise ValueError(
-                    f"netcdf4c in {io_section} of nuopc.runconfig is deprecated, use netcdf4p"
+                    f"netcdf4c in {io_section} of nuopc.runconfig is deprecated, "
+                      "use netcdf4p"
                 )
             else:
                 # if nc_type is netcdf, only one pe is needed
-                ioroot=int(self.runconfig.get(io_section, "pio_root"))
+                ioroot = int(self.runconfig.get(io_section, "pio_root"))
                 if cpucount < int(ioroot):
                     raise ValueError(
-                        f"Insufficient cpus for the {io_section} ioroot pe in nuopc.runconfig"
+                        f"Insufficient cpus for the {io_section} ioroot pe in "
+                        "nuopc.runconfig"
                     )
                 if nc_type == "netcdf4p":
-                    niotasks=int(self.runconfig.get(io_section, "pio_numiotasks"))
-                    iostride=int(self.runconfig.get(io_section, "pio_stride"))
+                    niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
+                    iostride = int(self.runconfig.get(io_section, "pio_stride"))
                     if cpucount <= (ioroot + niotasks*iostride):
                         raise ValueError(
-                            f"The iolayout for {io_section} in nuopc.runconfig is requesting out of range cpus"
+                            f"The iolayout for {io_section} in nuopc.runconfig is "
+                            "requesting out of range cpus"
                         )
 
         # Ensure that restarts will be written at the end of each run
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -198,7 +198,8 @@ class CesmCmeps(Model):
 
             if nthreads > 1:
                 npes = nthreads*ntasks*pestride
-                # this is taken from https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
+                # this is taken from 
+                # https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
                 # this correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
             else:
                 npes = (ntasks-1)*pestride + 1
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -13,7 +13,7 @@ from __future__ import print_function
 import os
 import re
 import shutil
-import multiprocessing
+from warnings import warn
 
 from payu.fsops import mkdir_p, make_symlink
 from payu.models.model import Model
@@ -187,9 +187,7 @@ class CesmCmeps(Model):
 
     def _setup_checks(self):
         # check pelayout fits within requested cpucount
-        cpucount = int(
-            self.expt.config.get('ncpus', multiprocessing.cpu_count())
-        )
+        cpucount = int(self.expt.config.get('ncpus'))
         all_realms = self.realms
         for realm in all_realms:
             ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
@@ -215,10 +213,11 @@ class CesmCmeps(Model):
                 )
 
             # check iolayout
-            if realm == "cpl":
+            if realm == "cpl" or realm == "med":
                 comp = "MED"  # med and cpl names are both used in runconfig
             else:
                 comp = realm.upper()
+            
             if comp in self.runconfig.get_component_list():
                 io_section = f"{comp}_modelio"
                 nc_type = self.runconfig.get(io_section, "pio_typename")
@@ -230,11 +229,17 @@ class CesmCmeps(Model):
                         f"in {NUOPC_CONFIG}."
                     )
 
-                match nc_type:
-                    case "netcdf":
-                        break
-                    case "netcdf4p" | "pnetcdf":
-                        if self.runconfig.get(io_section, "pio_async_interface") == ".false.":
+                pio_async = self.runconfig.get(io_section, "pio_async_interface")
+                if pio_async == ".true.":
+                    warn(
+                        "Payu does not do consistency checks for asynchronous pio, as "
+                        f"set in {io_section} of {NUOPC_CONFIG}. Consider adding them"
+                    )
+                elif pio_async == ".false.":
+                    match nc_type:
+                        case "netcdf":
+                            break
+                        case "netcdf4p" | "pnetcdf":
                             niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
                             iostride = int(self.runconfig.get(io_section, "pio_stride"))
                             if (ioroot + (niotasks-1)*iostride) >= npes:
@@ -242,17 +247,16 @@ class CesmCmeps(Model):
                                     f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
                                     "requesting out of range cpus"
                                 )
-                        # To-do: add coverage for pio_async == .true.
-                    case "netcdf4c":
-                        raise ValueError(
-                            f"netcdf4c in {io_section} of {NUOPC_CONFIG} is deprecated, "
-                            "use netcdf4p"
-                        )
-                    case _:
-                        raise ValueError(
-                            f"The iotype for {io_section} in {NUOPC_CONFIG} is "
-                            'invalid, valid options: "netcdf", "pnetcdf", "netcdf4p"'
-                        )
+                        case "netcdf4c":
+                            raise ValueError(
+                                f"netcdf4c in {io_section} of {NUOPC_CONFIG} is deprecated, "
+                                "use netcdf4p"
+                            )
+                        case _:
+                            raise ValueError(
+                                f"The iotype for {io_section} in {NUOPC_CONFIG} is "
+                                'invalid, valid options: "netcdf", "pnetcdf", "netcdf4p"'
+                            )
         return True
 
     def archive(self):
@@ -378,7 +382,7 @@ class Runconfig:
         else:
             return value
 
-    def get_component_list(self, value=None):
+    def get_component_list(self):
         """
         Get the `component_list`
         """
@@ -391,7 +395,7 @@ class Runconfig:
             components_str = m.group(1).strip()
             return components_str.split()
         else:
-            return value
+            return None
 
     def set(self, section, variable, new_value):
         """
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -21,6 +21,7 @@ from payu.models.fms import fms_collate
 from payu.models.mom6 import mom6_add_parameter_files
 
 NUOPC_CONFIG = "nuopc.runconfig"
+NUOPC_RUNSEQ = "nuopc.runseq"
 
 # Add as needed
 component_info = {
@@ -76,7 +77,7 @@ class CesmCmeps(Model):
             "drv_in",
             "fd.yaml",
             NUOPC_CONFIG,
-            "nuopc.runseq"
+            NUOPC_RUNSEQ
         ]
 
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -198,7 +198,6 @@ class CesmCmeps(Model):
             if nthreads < 1:
                 raise ValueError(f"The number of {realm}_nthreads ({nthreads}) in "
                                  f"{NUOPC_CONFIG} must be at least 1.")
-            
             if nthreads > 1:
                 npes = nthreads*ntasks*pestride
                 # this is taken from
@@ -244,10 +243,10 @@ class CesmCmeps(Model):
                             iostride = int(self.runconfig.get(io_section, "pio_stride"))
                             if (niotasks <= 0):
                                 warn(f"The pio_numiotasks for {io_section} in {NUOPC_CONFIG} is "
-                                    "not set, using model default")
+                                     "not set, using model default")
                             if (iostride <= 0):
                                 warn(f"The pio_stride for {io_section} in {NUOPC_CONFIG} is "
-                                    "not set, using model default")
+                                     "not set, using model default")
                             if (all([
                                     niotasks > 0, 
                                     iostride > 0,
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -195,7 +195,7 @@ class CesmCmeps(Model):
             nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
             rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
             pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
-            npes = (ntasks-1)*nthreads*pestride  # count to the last process, starting at 0
+            npes = ntasks*nthreads*pestride  # count to the last process, starting at 0
             if (rootpe + npes) > cpucount:
                 raise ValueError(
                     f"Insufficient cpus for the {realm} pelayout in {NUOPC_CONFIG}"
@@ -224,7 +224,7 @@ class CesmCmeps(Model):
                         if self.runconfig.get(io_section, "pio_async_interface") == ".false.":
                             niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
                             iostride = int(self.runconfig.get(io_section, "pio_stride"))
-                            if (int(ioroot) + (niotasks-1)*iostride) > npes:
+                            if (ioroot + niotasks*iostride) > npes:
                                 raise ValueError(
                                     f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
                                     "requesting out of range cpus"
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -20,6 +20,8 @@ from payu.models.model import Model
 from payu.models.fms import fms_collate
 from payu.models.mom6 import mom6_add_parameter_files
 
+NUOPC_CONFIG = "nuopc.runconfig"
+
 # Add as needed
 component_info = {
     "mom": {
@@ -73,7 +75,7 @@ class CesmCmeps(Model):
         self.config_files = [
             "drv_in",
             "fd.yaml",
-            "nuopc.runconfig",
+            NUOPC_CONFIG,
             "nuopc.runseq"
         ]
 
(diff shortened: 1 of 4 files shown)


anton-seaice:
Apologies to the wrong minghangli !! @minghangli-uni, can you look at this please?

payu/models/cesm_cmeps.py
```diff
@@ -163,8 +161,30 @@ def setup(self):
...
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+            #rootpe is zero-based
+            assert cpucount >= (rootpe + ntasks*pestride), f"Insufficient cpus for the {realm} pelayout in nuopc.runconfig"
+
+        # check iolayout
+        for realm in self.realms:
+            #med and cpl names are both used in runconfig
+            if realm == "cpl" : realm="MED"
+            io_section = f"{realm.upper()}_modelio"
```
anton-seaice, line 173: @dougiesquire - Do you know, in nuopc.runconfig, are the `_modelio` names case-sensitive?

i.e. will it always be `ICE_modelio` or is `ice_modelio` valid ?

payu/models/cesm_cmeps.py
```diff
@@ -163,8 +161,30 @@ def setup(self):
             )
         for realm in all_realms:
             ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
-            assert cpucount >= ntasks, "Insufficient cpus for the pelayout in nuopc.runconfig"
-        
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+            #rootpe is zero-based
+            assert cpucount >= (rootpe + ntasks*pestride), f"Insufficient cpus for the {realm} pelayout in nuopc.runconfig"
```
anton-seaice, line 167: @jo-basevi - Should I use a runtime error here instead of an asset ??

minghangli-uni:
Thanks @anton-seaice. I just did `payu setup` with a "wrong" pio setting, but it doesnt complain.

payu/models/cesm_cmeps.py
dougiesquire: I don't know sorry

minghangli-uni:
I've done a few tests on `pio_root`, `pio_numiotasks` and `pio_stride`. One `ICE_modelio` example below is based on a most recent merge of the [0.25 deg ryf configuration](https://github.com/ACCESS-NRI/access-om3-configs/tree/025deg_jra55do_ryf). The `ncpus` for `cice` is 96.

```
ICE_modelio::
     diro = ./log
     logfile = ice.log
     pio_async_interface = .false.
     pio_netcdf_format = nothing
     pio_numiotasks = 3
     pio_rearranger = 1
     pio_root = 100000
     pio_stride = 48
     pio_typename = netcdf4p
::
```

What I've found is,
1. `pio_root` does not appear to impact the I/O tasks as expected, which seems to be a bug. For example, setting `pio_root` to a very large value, such as 100000, does not affect the run.
2. Both `pio_numiotasks` and `pio_stride` function as intended, and their effects on the I/O tasks are properly reflected.

payu/models/cesm_cmeps.py
dougiesquire: Yes - see https://github.com/payu-org/payu/issues/367

payu/models/cesm_cmeps.py
```diff
@@ -156,15 +154,50 @@ def setup(self):
...
             ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
-            assert cpucount >= ntasks, "Insufficient cpus for the pelayout in nuopc.runconfig"
-        
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+            # rootpe is zero-based
+            if cpucount < (rootpe + ntasks*pestride):
+                raise ValueError(
+                    f"Insufficient cpus for the {realm} pelayout in nuopc.runconfig"
+                )
```
minghangli-uni, line 170:
```suggestion
            assert cpucount >= rootpe + ntasks * \
                pestride, "Insufficient cpus for the pelayout in nuopc.runconfig"
```

payu/models/cesm_cmeps.py
```diff
@@ -156,15 +154,50 @@ def setup(self):
...
+
+        # check iolayout
+        for realm in self.realms:
```
minghangli-uni, line 173:
```suggestion

```

This has to be removed, since the following check using ncpus for each component, instead of the total ncpus.

dougiesquire (commented):
Thanks for adding this @anton-seaice. Now that we're starting to accumulate a few checks in the driver, I think it would be worth putting these in a new `_qa_check` (or something) method that's called from `setup` for clarity.

payu/models/cesm_cmeps.py
```diff
@@ -156,15 +154,50 @@ def setup(self):
...
+                      "use netcdf4p"
+                )
+            else:
+                # if nc_type is netcdf, only one pe is needed
+                ioroot = int(self.runconfig.get(io_section, "pio_root"))
+                if cpucount < int(ioroot):
+                    raise ValueError(
+                        f"Insufficient cpus for the {io_section} ioroot pe in "
+                        "nuopc.runconfig"
+                    )
```
minghangli-uni, line 191: Since `pio_root` was found to be non-functional in [#496 (comment)](https://github.com/payu-org/payu/pull/496/link), can it be removed? If `pio_root` does function, I am not sure why we need this additional check?

```suggestion

```

payu/models/cesm_cmeps.py
```diff
@@ -156,15 +154,50 @@ def setup(self):
...
+                if nc_type == "netcdf4p":
+                    niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
+                    iostride = int(self.runconfig.get(io_section, "pio_stride"))
+                    if cpucount <= (ioroot + niotasks*iostride):
+                        raise ValueError(
+                            f"The iolayout for {io_section} in nuopc.runconfig is "
+                            "requesting out of range cpus"
+                        )
```
minghangli-uni, line 199:
```suggestion
                assert ntasks * pestride >= niotasks * \
                    iostride, (
                        f"The iolayout for {io_section} in nuopc.runconfig is requesting "
                        f"out of range cpus of {realm}_component ({ntasks*pestride})"
                    )
```

Instead of checking the total `ncpus`, here should be on the number of cpus for each component - `ntasks * pestride`

minghangli-uni (changes requested):
Thanks @anton-seaice for your work! I've left some comments. Can you please take a look at those?

payu/models/cesm_cmeps.py
anton-seaice: I am not sure what you are trying to change here ?

dougiesquire: [Assert is an anti-pattern in production code](https://github.com/payu-org/payu/issues/367)

anton-seaice:
> I've done a few tests on `pio_root`, `pio_numiotasks` and `pio_stride`. One `ICE_modelio` example below is based on a most recent merge of the [0.25 deg ryf configuration](https://github.com/ACCESS-NRI/access-om3-configs/tree/025deg_jra55do_ryf). The `ncpus` for `cice` is 96.
> 
> ```
> ICE_modelio::
>      diro = ./log
>      logfile = ice.log
>      pio_async_interface = .false.
>      pio_netcdf_format = nothing
>      pio_numiotasks = 3
>      pio_rearranger = 1
>      pio_root = 100000
>      pio_stride = 48
>      pio_typename = netcdf4p
> ::
> ```
> 
> What I've found is,
> 
>     1. `pio_root` does not appear to impact the I/O tasks as expected, which seems to be a bug. For example, setting `pio_root` to a very large value, such as 100000, does not affect the run.

`pio_root` in this example would be (almost silently) changed to 0 in the pio shared code. It's not a behaviour I love, I think it should just cause an error. Payu would reject this after this PR.

payu/models/cesm_cmeps.py
minghangli-uni: I was thinking this is consistent with what `payu` did. Sorry I didnt check https://github.com/payu-org/payu/issues/367

payu/models/cesm_cmeps.py
anton-seaice: My preference is we force the user to set pio_root correctly, rather than let the model change the value

payu/models/cesm_cmeps.py
jo-basevi: Yes, I think there is still a few asserts in payu code in general, but the goal is move towards exceptions instead

payu/models/cesm_cmeps.py
anton-seaice: I sort of agree. I think we should be comparing to {realm}_ntasks rather than cpucoun

payu/models/cesm_cmeps.py
anton-seaice: :)

payu/models/cesm_cmeps.py
anton-seaice: :) - thanks

payu/models/cesm_cmeps.py
dougiesquire: I assume the reversal of the `<` is a typo?

payu/models/cesm_cmeps.py
anton-seaice: I refactored the code such that this comment doesn't apply ( I think)

anton-seaice:
@minghangli-uni - Are you able to review this please? I updated the initial comment with the scope

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
+    def _setup_checks(self):
+        # check pelayout fits within requested cpucount
+        cpucount = int(
+            self.expt.config.get('ncpus', multiprocessing.cpu_count())
+        )
+        all_realms = self.realms
```
minghangli-uni, line 192:
```suggestion
        all_realms = self.realms + ["glc", "lnd"]
```

Are `glc` and `lnd` necessary?

minghangli-uni (changes requested):
Thanks @anton-seaice for implementing this check. I have some comments and suggested changes that need to be addressed.

I havent checked the tests yet.

payu/models/cesm_cmeps.py
anton-seaice: No we can remove the `glc` and `lnd` sections from PELayout in our configs.

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
...
+        for realm in all_realms:
+            ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
+            nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+
+            if (pestride<nthreads):
+                raise ValueError(
+                    f"{realm}_pestride needs to be equal or greater than {realm}_nthreads"
+                )
```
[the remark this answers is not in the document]
anton-seaice: But we can have threads across multiple cores, can't we ?

payu/models/cesm_cmeps.py
```diff
@@ -22,6 +20,8 @@
 from payu.models.fms import fms_collate
 from payu.models.mom6 import mom6_add_parameter_files
 
+NUOPC_CONFIG = "nuopc.runconfig"
```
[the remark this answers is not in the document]
anton-seaice: No but its neater / easier to change in the future

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
...
+            if comp in self.runconfig.get_component_list():
+                io_section = f"{comp}_modelio"
+                nc_type = self.runconfig.get(io_section, "pio_typename")
+                ioroot = int(self.runconfig.get(io_section, "pio_root"))
+
+                if ioroot >= npes:
+                    raise ValueError(
+                        f"Insufficient cpus for the {io_section} ioroot pe in "
+                        f"{NUOPC_CONFIG}"
+                    )
```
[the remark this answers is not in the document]
anton-seaice: {
    "PELAYOUT_attributes": dict(
        moc_ntasks=1,
        moc_nthreads=1,
        moc_pestride=1,
        moc_rootpe=0
    ),
    "MOC_modelio": dict(
        pio_numiotasks=1,
        pio_root=1,
        pio_stride=1,
        )
}

ie pio_root is asking for a PE which is not available

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
...
+            if (pestride<nthreads):
+                raise ValueError(
+                    f"{realm}_pestride needs to be equal or greater than {realm}_nthreads"
+                )
+
+            npes = (ntasks-1)*pestride + nthreads  # count to the last process, starting at 0
+
+            # can pestride be less than nthreads ??
+
+            if (rootpe + npes) > cpucount:
```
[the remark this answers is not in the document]
anton-seaice: cpucount is from config.yaml for all realms.

ie if cpucount is 240, npes might be 196 and rootpe = 48 for one realm , so you need to sum them to check its within the range of cpucount

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: Threads can be distributed across cores of a node but each core will handle only one thread at a time. Hence there will be only one thread per core without the ability for each core to deal with multi threads at the same time.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: I've updated your `npes` on line 204.

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
...
+            nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+
+            if (pestride<nthreads):
+                raise ValueError(
+                    f"{realm}_pestride needs to be equal or greater than {realm}_nthreads"
+                )
+
+            npes = (ntasks-1)*pestride + nthreads  # count to the last process, starting at 0
```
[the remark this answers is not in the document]
anton-seaice: I did the original calculate of npes to be the number of pes for each realm (so we can check the iolayout too)

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Yes so we need this check to ensure pestride >= nthreads so that its not trying to put more than one thread on a core?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,73 @@ def setup(self):
...
 
+    def _setup_checks(self):
+        # check pelayout fits within requested cpucount
+        cpucount = int(
+            self.expt.config.get('ncpus', multiprocessing.cpu_count())
+        )
+        all_realms = self.realms
+        for realm in all_realms:
+            ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
+            nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
```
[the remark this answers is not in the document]
anton-seaice: We should include this for future proofing

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: I dont think so, since threading is not enabled from CMEPS. And why does pestride need to be equal or greater than nthreads?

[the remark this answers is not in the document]
anton-seaice: If pestride =1 and nthreads =1 , ntasks =2 :

task0 - thread0 -> PE0
task0 - thread1 -> PE1
task1 - thread0 -> PE1
task1 -thread1 -> PE 2

-> which is problematic because theres two threads on PE1 ?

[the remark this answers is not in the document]
anton-seaice: i.e. pestride >= nthreads to enforce there is one thread per pe

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: Thank you. But the error message is misleading. How about,
```suggestion
                if ioroot >= npes:
                    raise ValueError(
                       f"{io_section} ioroot pe exceeds available PEs ({npes - 1}) "
                       f"for {comp} in {NUOPC_CONFIG}."
                       f"Ensure {io_section} ioroot pe is within the range."
                    )
```

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: Ha, I see. So `npes` should be,
```suggestion
            npes = (ntasks-1)*pestride  # count to the last process, starting at 0
```

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: And this should be,
```suggestion
            if (rootpe + npes) >= cpucount:
```

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: I’m not sure. It is not currently in use and its removal would streamline the code.

If, in the future, Gadi provides cpus with hyperthreading or if we need nthreads, we can certainly revisit its inclusion then. For now, I believe removing it will help keep our codebase more focused on our immediate needs.

Or we could leave a comment, such as `#not used`, to indicate thread is not in use directly in `nuopc.runconfig`, which is similar to what we have done for other variables.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Assuming i've understood correctly, threads would use extra PEs 

i.e. with ntasks = 4, pestride=2 and nthreads=2 , the last PE will be number 7 (not number 6 per the suggested change).

with ntasks = 4, pestride=2 and nthreads=1, the last PE would be number 6

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: It can't be 0, so its not really "not used".

I would definitely like to investigate threading in cice, and think the option might get used.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Ok thanks :) I tweaked the suggestion

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: I guess i should rename `npes` to `last_pe` ?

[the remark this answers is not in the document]
anton-seaice: i.e. with ntasks =2, pstride = 4 , nthreads = 2 , the PEs used are 0,1,4,5 so npes is 6

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: I've tried -1 and 0, both work.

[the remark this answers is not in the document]
anton-seaice: Oh sorry , yes - but I guess that just means the model has overwritten the setting ?

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: if rootpe is 5, and npes is 6, cpucount is 11 and valid / should be excepted

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: I was also thinking that pestride is the spacing between PEs allocated to each task, but not each thread. This was my initial understanding as well. But CMEPS considers it differently.

In CMEPS, the pestride affects the distribution of PEs in a way that incorporates threading. When `ESMF_AWARE_THREADING` is enabled, the petlist is generated based on the number of tasks, threads, and the stride, as shown [here](https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1005-L1010):

```
#ifdef ESMF_AWARE_THREADING
       cnt = 1
       do ntask = rootpe, rootpe+nthrds*ntasks*stride-1, stride
          petlist(cnt) = ntask
          cnt = cnt + 1
       enddo
```
So, CMEPS will output the PEs as 0, 4, 8, 12 with rootpe = 0.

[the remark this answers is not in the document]
anton-seaice: thankyou ! I will fix this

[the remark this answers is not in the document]
minghangli-uni: But I really doubt we will use the hyperthreading.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: why is npes 6?

Lets use the example of we currently have for 1deg ryf configuration.

```
  ocn_ntasks = 216
  ocn_nthreads = 0
  ocn_pestride = 1
  ocn_rootpe = 24
```

rootpe is 24, npes = (ntasks-1)*pestride, which is 215, rootpe+npes=239. This shall be less than the total requested cpus, which is 240.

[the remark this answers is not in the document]
anton-seaice: hmm - I think we should set nthreads to 1 in that case ? (because thats the real value)

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: There are also `nuopc.runconfig` and `nuopc.runseq` somewhere else. Would it be good to open a separete PR to address this issue?

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: We dont need to set `nthreads` explicitly, as it defaults to 1 unless `ESMF_AWARE_THREADING` is enabled.

Even though `ESMF_AWARE_THREADING` is on, the relation will still hold true, cuz the cal `rootpe+nthrds*ntasks*stride-1`, which minus 1.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: Yes, it is overwritten by `maxthreads`, which [defaults to be 1](https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L930).

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Are they in other files ?

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Ok i put a fix in for this

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: Ok the fix is in

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
anton-seaice: I added an if nthread>1 case which captures this

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
minghangli-uni: Are you going to add another one for `nuopc.runseq`?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,72 @@ def setup(self):
...
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
```
minghangli-uni, line 196: is `rootpe` obselete?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,72 @@ def setup(self):
...
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+
+            if nthreads > 1:
+                npes = nthreads*ntasks*pestride
```
minghangli-uni, line 200: Why isnt npes now the last index of the pelist? This is confusing to me.

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +184,72 @@ def setup(self):
...
+                # this is taken from 
+                # https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
+                # this correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
+            else:
+                npes = (ntasks-1)*pestride + 1
```
minghangli-uni, line 205: same here. what is the definition of npes?

payu/models/cesm_cmeps.py
anton-seaice: No we need it

payu/models/cesm_cmeps.py
anton-seaice: pelist starts counting at 0 , npes starts counting at 1

ie for one task, pe is 0

minghangli-uni: Ah I see. Thanks!

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,72 @@ def setup(self):
...
+        cpucount = int(
+            self.expt.config.get('ncpus', multiprocessing.cpu_count())
+        )
+        all_realms = self.realms
+        for realm in all_realms:
+            ntasks = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_ntasks"))
+            nthreads = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_nthreads"))
+            rootpe = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_rootpe"))
+            pestride = int(self.runconfig.get("PELAYOUT_attributes", f"{realm}_pestride"))
+
```
minghangli-uni, line 199: Does this improvement work better? In case nthreads is less than 1. 
```suggestion

            if nthreads < 1:
                raise ValueError(f"The number of {realm}_nthreads ({nthreads}) in "
                                 f"{NUOPC_CONFIG} must be at least 1.")
                                 
```

anton-seaice: It depends if we want to support setting it automatically (i.e. `-99`) ?

anton-seaice:
@aidanheerdegen - this is ready to go. Can you or one of the release team want to review ?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+            if nthreads < 1:
+                raise ValueError(f"The number of {realm}_nthreads ({nthreads}) in "
+                                 f"{NUOPC_CONFIG} must be at least 1.")
+                                 
+            if nthreads > 1:
+                npes = nthreads*ntasks*pestride
+                # this is taken from 
+                # https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
+                # this correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
```
anton-seaice, line 208:
```suggestion
                # the correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
```

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
+    def _setup_checks(self):
+        # check pelayout fits within requested cpucount
+        cpucount = int(
+            self.expt.config.get('ncpus', multiprocessing.cpu_count())
```
aidanheerdegen, line 191: Multiprocessing is not designed to be used on multiple nodes

https://docs.nersc.gov/development/languages/python/parallel-python/#multiprocessing

Consequently I don't think `multiprocessing.cpu_count()` will return useful information and you should just insist that the user set this value and error out otherwise.

payu/models/cesm_cmeps.py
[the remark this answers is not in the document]
aidanheerdegen: As `cpucount` is for all realms why do this check within the loop over realms?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+            if (rootpe + npes) > cpucount:
+                raise ValueError(
+                    f"Insufficient cpus for the {realm} pelayout in {NUOPC_CONFIG}"
+                )
+
+            # check iolayout
+            if realm == "cpl":
+                comp = "MED"  # med and cpl names are both used in runconfig
+            else:
+                comp = realm.upper()
```
aidanheerdegen, line 221:
```suggestion
            if realm == "cpl" or realm == "med":
                comp = "MED"  # med and cpl names are both used in runconfig
```

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+                        # To-do: add coverage for pio_async == .true.
+                    case "netcdf4c":
+                        raise ValueError(
+                            f"netcdf4c in {io_section} of {NUOPC_CONFIG} is deprecated, "
+                            "use netcdf4p"
+                        )
+                    case _:
+                        raise ValueError(
+                            f"The iotype for {io_section} in {NUOPC_CONFIG} is "
+                            'invalid, valid options are "netcdf", "pnetcdf" and "netcdf4p"'
```
aidanheerdegen, line 254:
```suggestion
                            'invalid, valid options: "netcdf", "pnetcdf", "netcdf4p"'
```
or change "and" to "or", as the latter is more suggestive of choosing one option.

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+                    case "netcdf4p" | "pnetcdf":
+                        if self.runconfig.get(io_section, "pio_async_interface") == ".false.":
+                            niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
+                            iostride = int(self.runconfig.get(io_section, "pio_stride"))
+                            if (ioroot + (niotasks-1)*iostride) >= npes:
+                                raise ValueError(
+                                    f"The iolayout for {io_section} in {NUOPC_CONFIG} is "
+                                    "requesting out of range cpus"
+                                )
+                        # To-do: add coverage for pio_async == .true.
```
aidanheerdegen, line 245: If `pio_async == .true.` isn't supported what can we tell the user? Warn them it hasn't been checked for consistency?

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+                        f"in {NUOPC_CONFIG}."
+                    )
+
+                match nc_type:
+                    case "netcdf":
+                        break
+                    case "netcdf4p" | "pnetcdf":
+                        if self.runconfig.get(io_section, "pio_async_interface") == ".false.":
+                            niotasks = int(self.runconfig.get(io_section, "pio_numiotasks"))
+                            iostride = int(self.runconfig.get(io_section, "pio_stride"))
```
aidanheerdegen, line 239: Is `pio_stride` always defined? If not then this should have a default value.

payu/models/cesm_cmeps.py
```diff
@@ -313,6 +378,21 @@ def get(self, section, variable, value=None):
...
+        m = re.search(
+            r'component_list:\s*(.*)',
+            self.contents
+        )
+
+        if m is not None:
+            components_str = m.group(1).strip()
+            return components_str.split()
+        else:
+            return value
```
aidanheerdegen, line 394: Why does this return an input variable? It isn't used this way in the code.

payu/models/cesm_cmeps.py
```diff
@@ -190,6 +185,76 @@ def setup(self):
...
+                raise ValueError(f"The number of {realm}_nthreads ({nthreads}) in "
+                                 f"{NUOPC_CONFIG} must be at least 1.")
+                                 
+            if nthreads > 1:
+                npes = nthreads*ntasks*pestride
+                # this is taken from 
+                # https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1007
+                # the correct calculation might be (ntasks-1)*pestride*nthreads + nthreads
+            else:
+                npes = (ntasks-1)*pestride + 1
```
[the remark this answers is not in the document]
minghangli-uni: In CMEPS, the pestride affects the distribution of PEs in a way that incorporates threading. For `nthreads > 1`, the petlist is generated based on the number of tasks, threads, and the stride, as shown [here](https://github.com/ESCOMP/CMEPS/blob/5b7d76978e2fdc661ec2de4ba9834b985decadc6/cesm/driver/esm.F90#L1005-L1015):

```
#ifdef ESMF_AWARE_THREADING
       cnt = 1
       do ntask = rootpe, rootpe+nthrds*ntasks*stride-1, stride
          petlist(cnt) = ntask
          cnt = cnt + 1
       enddo
```

This is for single thread, 
```
       do ntask = 1, size(petlist)
          petlist(ntask) = rootpe + (ntask-1)*stride
       enddo
```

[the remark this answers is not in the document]
aidanheerdegen: Thanks @minghangli-uni.

payu/models/cesm_cmeps.py
anton-seaice: I just followed the example of the `get()` function

payu/models/cesm_cmeps.py
anton-seaice: That doesn't sound like a very useful warning ? ...

"Warning: payu hasn't done something you probably weren't expecting it to do ... ?"

aidanheerdegen: I think from a users perspective `payu` is doing a number of consistency checks but for this specific case it doesn't. I think it is useful to tell users this so they don't assume if they have an error this part is fine because `payu` hasn't complained.

payu/models/cesm_cmeps.py
aidanheerdegen: Oh right. Unless you think it is likely this functionality will be used at some point in the near future I'd recommend removing it. It confuses the intent, and if you need proof, I was confused.

Extraneous code is still code that must be tested and maintained.

So I'd say remove the input variable and just return `None`. Or you could do ultra minimalist:
```python
    def get_component_list(self):
        """
        Get the `component_list`
        """
        m = re.search(
            r'component_list:\s*(.*)',
            self.contents
        )
        return None if m is None else m.group(1).strip().split()
```

Or probably better, check you're getting what you expect even if not `None`
```python
    def get_component_list(self):
        """
        Get the `component_list`
        """
        m = re.search(
            r'component_list:\s*(.*)',
            self.contents
        )

        try:
            return m.group(1).strip().split()
        except SomeExceptionsThatAreOkToIgnore:
            return None
```

aidanheerdegen:
Before I forget when this does get merged please either squash when merging, or rebase beforehand to clean up the commit history.

payu/models/cesm_cmeps.py
anton-seaice: ok done

payu/models/cesm_cmeps.py
anton-seaice: I put a check in and handling for pio_stride<=0 and pio_numiotasks<0

payu/models/cesm_cmeps.py
anton-seaice: Ok this has a warning now

payu/models/cesm_cmeps.py
anton-seaice: Ok done

payu/models/cesm_cmeps.py
anton-seaice: Ok done

aidanheerdegen (approved):
LGTM. I have pinged @minghangli-uni because there are some unresolved conversations that he should check have been resolved and mark them as such. Then good to merge I think.

minghangli-uni (approved):
LGTM! Thanks @anton-seaice

PR #504, opened 2024-09-08
Default shortpath project bug

### LINKED ISSUE #502: `shortpath` differences with default project
There's a difference between `shortpath` default on the login node and pbs payu run submission when `project` is defined in `config.yaml`.  With the logic with creating new `UUID`s when there isn't a pre-existing archive, this can result in restarts not being picked up in payu as brought up in this forum post: https://forum.access-hive.org.au/t/payu-archives-and-restarts/3626

To replicate the error on my NCI account, I had a `config.yaml` with the following settings:

```yaml
project: nf33
# shortpath: /scratch/nf33
```

Then running payu commands on the login node (e.g. `payu setup`/`payu checkout -b`/`payu clone`), the `shortpath` defaults to `/scratch/tm70` because `tm70` is my default project. This creates an archive at `/scratch/nf33/jb4202/mom6/archive/mom6-double-grye-base-c248b389`. Then on `payu run`, the`shortpath` defaults to `/scratch/nf33` and there isn't an archive under `/scratch/nf33/jb4202/mom6/archive/mom6-double-grye-base-c248b389` so it creates a new archive with a new UUID, e.g. `/scratch/nf33/jb4202/mom6/work/mom6-double-grye-base-42bdfc18`. Then once the run has finished, running `payu setup`, it can't find `/scratch/tm70/jb4202/mom6/work/mom6-double-grye-base-42bdfc18` so it creates a new UUID and archive, and so on..

So basically payu needs to use the `project` in `config.yaml` in it's default project for `shortpath`. `shortpath` is currently being set here: 
https://github.com/payu-org/payu/blob/2826621a83bcb13315d3a003a946c5ca8069b1cd/payu/laboratory.py#L69-L74

I think sometime like the following might work:

```
default_project = config.get('project', os.environ.get('PROJECT', ''))
```

In the meantime to avoid this bug, setting `shortpath` to the `config.yaml` to match the project would work. For example:

```yaml
project: nf33
shortpath: /scratch/nf33
```

jo-basevi:
Closes #502 

If `project` is defined in `config.yaml`, use this project for default shortpaths. This is to prevent a mismatch of default shortpath when running commands on a login-node vs on a PBS `payu run` submission.

I also refactored the tests for laboratory basepaths to make each test a bit more isolated. I also added a couple more test cases.

### DIFF
--- a/payu/laboratory.py
+++ b/payu/laboratory.py
@@ -67,7 +67,7 @@ class Laboratory(object):
         # Default path settings
 
         # Append project name if present (NCI-specific)
-        default_project = os.environ.get('PROJECT', '')
+        default_project = config.get('project', os.environ.get('PROJECT', ''))
         default_short_path = os.path.join(self.base, default_project)
         default_user = pwd.getpwuid(os.getuid()).pw_name
 
--- a/test/test_paths.py
+++ b/test/test_paths.py
@@ -2,108 +2,124 @@ import os
 from pathlib import Path
 import shutil
 
-import pdb
 import pytest
-import yaml
-
-import payu
 
 from payu.laboratory import Laboratory
-from payu.schedulers.pbs import find_mounts
 
-
-from .common import cd, make_random_file, get_manifests
-from .common import tmpdir, ctrldir, labdir, workdir
-from .common import config, sweep_work, payu_init, payu_setup
+from .common import cd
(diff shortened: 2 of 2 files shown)


aidanheerdegen (approved):
LGTM. 

Nice work cleaning up the tests. Thanks.

PR #507, opened 2024-09-11
Don't require `cice_in.nml` namelist in CICE5 restarts

### LINKED ISSUE #499: Branching OM2 from existing restart cannot find `cice_in.nml`
Branching an OM2 experiment from a previous simulation's restart produces an error at the payu setup stage. Following the instructions [here](https://forum.access-hive.org.au/t/running-model-experiments-with-payu-and-git/2285), i.e.

```
gh repo clone ACCESS-Community-Hub/access-om2-1deg_jra55_ryf-example
cd access-om2-1deg_jra55_ryf-example
payu checkout -r /g/data/nf33/public/training-day-2024/payu-training/experiments/20240827-release-preindustrial+concentrations-run-0225dcf2/restart020 -b perturb1 0f2e2bb
```
followed by 
```
payu setup
```
leads to the following error:
```
...
Setting up atmosphere
Setting up ocean
Setting up ice
payu: error: Cannot find prior namelist cice_in.nml
```

This error is raised in the `cice.py` driver:
https://github.com/payu-org/payu/blob/2826621a83bcb13315d3a003a946c5ca8069b1cd/payu/models/cice.py#L210-L219

`cice.py` first searches the restart directory and then the previous output directory for a `cice_in.nml` file. OM2 simulations don't make a copy of `cice_in.nml` in the restart directories, and hence when branching from a separate restart, payu won't find this file.

---

A quick fix might be to edit the `cice5.py` driver so that `cice_in.nml` is always copied to the restart directory. The `access.py` driver [already does this for ESM1.5 ](https://github.com/payu-org/payu/blob/2826621a83bcb13315d3a003a946c5ca8069b1cd/payu/models/access.py#L253C15-L264), so we might be able to  do this step for both OM2 and ESM1.5 directly from `cice.py`.

However the output directory/restart directory version of `cice_in.nml` might not be necessary for OM2. Payu uses it to calculate a run length [here](https://github.com/payu-org/payu/blob/2826621a83bcb13315d3a003a946c5ca8069b1cd/payu/models/cice.py#L249-L258), however I think this uses a different timestep to [the one actually used by the model](https://github.com/ACCESS-NRI/access-om2-configs/blob/0a29b451744dcbc82a90a8b663ce5f7f0d3f2bc2/accessom2.nml#L11). E.g. I'd run a 5 month simulation a while ago and the ice output directory contains:
```bash
cd output000/ice
grep npt *
cice_in.nml:    npt = 35040.0 (from the payu cice.py calculation)
ice_diag.d:  npt                       =     2416 (Equals 151 days assuming 1.5h timestep as specified in accessom2.nml)
```

If the information from the output directory `cice_in.nml` isn't used, I'm wondering if we should stop reading it at all for OM2 simulations? We'd just need to think about the best way to handle ESM1.5 which does read it from the restart directory (but perhaps also might not need to?).

blimlim:
Attempts to close #499 

This pull request modifies the `cice.py` driver, by pulling the timing calculations out from `setup()` into a separate `calc_runtime()` method. This method is then overridden in the `cice5.py` driver,  allowing the `cice_in.nml` namelist file to be absent from the restart directory, since OM2 does not use the information from these calculations. This in turn will allow OM2 to be run after using `payu clone -r` with a specified OM2 restart directory.

Most of the code changes are in the tests, which require a bit of intricate set up. Given that we're skipping `payu setup` and directly running `model.setup()` (in order to avoid more complicated pre-test setup), it's difficult to know how well we've matched the conditions for setting up a real experiment. E.g. When setting up the tests, inadvertent changes to the directories that existed at the start of the test led to payu using different work restart directories during different parts of the setup, and different restart files were moved to different places in a way that was tricky to understand. And so while the tests are working, they do feel a bit precarious. If you have any suggestions that would be very welcome!

The main relevabt tests are the `test_restart_clone` tests for both [CICE 4 ](https://github.com/payu-org/payu/blob/10f5122a16798d9ca12e4453809a9b4399b91b4f/test/models/test_cice.py#L296-L303)and [CICE 5](https://github.com/payu-org/payu/blob/10f5122a16798d9ca12e4453809a9b4399b91b4f/test/models/test_cice5.py#L238-L247). These do the following:

1. Create a restart directory with fake restart files required by CICE4/5.
2. Clone an experiment and point to the created restart directory.
3. Set up CICE within the cloned control directory to see if it works. 

The CICE 5 version of `test_restart_clone` passes with the new cice driver changes, and would have failed previously.

One thing I'm unsure about: Do you think it's necessary to include the cloning steps in the tests. I think the main issue was payu requiring restart files that cice5 didn't use. This came up when trying to clone, but I don't think the issue came from the cloning step itself. Perhaps it's still useful to be testing the whole clone/setup process though.

Any feedback/suggestions would be very welcome! I'd be keen to get some other ideas before trying to finalise this, and so have marked this PR as a draft.

### DIFF
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -325,8 +325,9 @@ class Cice(Model):
             if os.path.isfile(test_path):
                 input_path = test_path
                 break
-        raise RuntimeError(
-            f"Cannot find previous restart file, expected {fpath} to exist"
-        )
+        if input_path is None:
+            raise RuntimeError(
+                f"Cannot find previous restart file, expected {fpath} to exist"
+            )
 
         make_symlink(input_path, input_work_path)
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -185,8 +185,7 @@ class Cice(Model):
                 iced_restart_file = sorted(iced_restart_files)[-1]
 
             if iced_restart_file is None:
-                print('payu: error: No restart file available.')
-                sys.exit(errno.ENOENT)
+                raise RuntimeError('payu: error: No restart file available.')
 
             res_ptr_path = os.path.join(self.work_init_path,
                                         'ice.restart_file')
@@ -326,6 +325,8 @@ class Cice(Model):
             if os.path.isfile(test_path):
                 input_path = test_path
                 break
-        assert input_path
+        raise RuntimeError(
+            f"Cannot find previous restart file, expected {fpath} to exist"
+        )
 
         make_symlink(input_path, input_work_path)
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -202,14 +202,32 @@ class Cice(Model):
             if setup_nml['restart']:
                 self.link_restart(setup_nml['pointer_file'])
 
-        self.calc_timestep_runtime(setup_nml)
+        self.calc_runtime(setup_nml)
 
         # Write any changes to the work directory copy of the cice
         # namelist
         nml_path = os.path.join(self.work_path, self.ice_nml_fname)
         self.ice_in.write(nml_path, force=True)
 
-    def calc_timestep_runtime(self, setup_nml):
+    def calc_runtime(self, setup_nml):
+        """
+        Calculate 1: the previous number of timesteps simulated, and 2:
+        the number of timesteps to simulate in the next run.
+
+        Note 1: This method is overridden in the cice5 driver, as cice5
+        in ACCESS OM2 does not require the calculated runtime information.
+        Instead it performs the calculations within the model using the
+        restart files and OM2 namelist.
+
+        Note 2: For ESM1.5, the actual model start date and run time are
+        controlled via the separate input_ice.nml namelist, with relevant
+        calculations in the access driver.
+
+        Parameters
+        ----------
+        setup_nml: Open 'setup_nml' section of the cice_in.nml namelist.
+        Modifies in place.
+        """
         init_date = datetime.date(year=setup_nml['year_init'], month=1, day=1)
         if setup_nml['days_per_year'] == 365:
             caltype = cal.NOLEAP
@@ -219,13 +237,18 @@ class Cice(Model):
         if self.prior_restart_path:
 
             prior_nml_path = os.path.join(self.prior_restart_path,
-                                        self.ice_nml_fname)
+                                          self.ice_nml_fname)
+
+            # TODO: Are there any models which leave cice_in.nml in
+            # the output rather than restart directory? If not we can remove
+            # this.
+
             # With later versions this file exists in the prior restart path,
             # but this was not always the case, so check, and if not there use
             # prior output path
             if not os.path.exists(prior_nml_path) and self.prior_output_path:
                 prior_nml_path = os.path.join(self.prior_output_path,
-                                                self.ice_nml_fname)
+                                              self.ice_nml_fname)
 
             # If we cannot find a prior namelist, then we cannot determine
             # the start time and must abort the run.
@@ -241,12 +264,13 @@ class Cice(Model):
             prior_runtime_seconds = prior_runtime * prior_setup_nml['dt']
 
         else:
-            # Initialise runtime
+            # If no prior restart directory exists, set the prior runtime to 0
             prior_runtime_seconds = 0
 
-        # Set runtime for this run.
+        # Calculate runtime for this run.
         if self.expt.runtime:
-            run_start_date = cal.date_plus_seconds(init_date, prior_runtime_seconds,
+            run_start_date = cal.date_plus_seconds(init_date,
+                                                   prior_runtime_seconds,
                                                    caltype)
             run_runtime = cal.runtime_from_date(
                 run_start_date,
@@ -258,17 +282,12 @@ class Cice(Model):
             )
         else:
             run_runtime = setup_nml['npt']*setup_nml['dt']
-        
+
         # Add the prior runtime and new runtime to the working copy of the
         # CICE namelist.
         setup_nml['npt'] = run_runtime / setup_nml['dt']
         assert (prior_runtime_seconds % setup_nml['dt'] == 0)
         setup_nml['istep0'] = int(prior_runtime_seconds / setup_nml['dt'])
-
-        
-
-
-
 
     def set_local_timestep(self, t_step):
         dt = self.ice_in['setup_nml']['dt']
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -167,12 +167,6 @@ class Cice(Model):
             self.ice_in.patch(history_nml)
 
         setup_nml = self.ice_in['setup_nml']
-        init_date = datetime.date(year=setup_nml['year_init'], month=1, day=1)
-
-        if setup_nml['days_per_year'] == 365:
-            caltype = cal.NOLEAP
-        else:
-            caltype = cal.GREGORIAN
 
         if self.prior_restart_path:
             # Generate ice.restart_file
@@ -200,15 +194,38 @@ class Cice(Model):
             setup_nml['runtype'] = 'continue'
             setup_nml['restart'] = True
 
-            prior_nml_path = os.path.join(self.prior_restart_path,
-                                          self.ice_nml_fname)
+        else:
+            # Locate and link any restart files (if required)
+            if not setup_nml['ice_ic'] in ('none', 'default'):
+                self.link_restart(setup_nml['ice_ic'])
 
+            if setup_nml['restart']:
+                self.link_restart(setup_nml['pointer_file'])
+
+        self.calc_timestep_runtime(setup_nml)
+
+        # Write any changes to the work directory copy of the cice
+        # namelist
+        nml_path = os.path.join(self.work_path, self.ice_nml_fname)
+        self.ice_in.write(nml_path, force=True)
+
+    def calc_timestep_runtime(self, setup_nml):
+        init_date = datetime.date(year=setup_nml['year_init'], month=1, day=1)
+        if setup_nml['days_per_year'] == 365:
+            caltype = cal.NOLEAP
+        else:
+            caltype = cal.GREGORIAN
+
+        if self.prior_restart_path:
+
+            prior_nml_path = os.path.join(self.prior_restart_path,
+                                        self.ice_nml_fname)
             # With later versions this file exists in the prior restart path,
             # but this was not always the case, so check, and if not there use
             # prior output path
             if not os.path.exists(prior_nml_path) and self.prior_output_path:
                 prior_nml_path = os.path.join(self.prior_output_path,
-                                              self.ice_nml_fname)
+                                                self.ice_nml_fname)
 
             # If we cannot find a prior namelist, then we cannot determine
             # the start time and must abort the run.
@@ -220,22 +237,16 @@ class Cice(Model):
             prior_setup_nml = f90nml.read(prior_nml_path)['setup_nml']
 
             # The total time in seconds since the beginning of the experiment
-            total_runtime = prior_setup_nml['istep0'] + prior_setup_nml['npt']
-            total_runtime = total_runtime * prior_setup_nml['dt']
+            prior_runtime = prior_setup_nml['istep0'] + prior_setup_nml['npt']
+            prior_runtime_seconds = prior_runtime * prior_setup_nml['dt']
+
         else:
-            # Locate and link any restart files (if required)
-            if not setup_nml['ice_ic'] in ('none', 'default'):
-                self.link_restart(setup_nml['ice_ic'])
-
-            if setup_nml['restart']:
-                self.link_restart(setup_nml['pointer_file'])
-
             # Initialise runtime
-            total_runtime = 0
+            prior_runtime_seconds = 0
 
         # Set runtime for this run.
         if self.expt.runtime:
-            run_start_date = cal.date_plus_seconds(init_date, total_runtime,
+            run_start_date = cal.date_plus_seconds(init_date, prior_runtime_seconds,
                                                    caltype)
             run_runtime = cal.runtime_from_date(
                 run_start_date,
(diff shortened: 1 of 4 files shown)


jo-basevi:
Thanks for putting in these changes! 

> One thing I'm unsure about: Do you think it's necessary to include the cloning steps in the tests. I think the main issue was payu requiring restart files that cice5 didn't use. This came up when trying to clone, but I don't think the issue came from the cloning step itself. Perhaps it's still useful to be testing the whole clone/setup process though.

I think cloning wouldn't be necessary in these tests, as when a restart is added through cloning, payu just sets the `restart` option to `config.yaml`, e.g.
```yaml
restart: <fullpath/to/restart>
```
This can be replicated in tests by configuring `restart` in the test `config.yaml` files as part of test setup.  I think the restart path for cloning is currently tested for in `test_branch.py`.

payu/models/cice.py
```diff
@@ -201,9 +194,55 @@ def setup(self):
...
+                self.link_restart(setup_nml['pointer_file'])
+
+        self.calc_runtime(setup_nml)
+
+        # Write any changes to the work directory copy of the cice
+        # namelist
+        nml_path = os.path.join(self.work_path, self.ice_nml_fname)
+        self.ice_in.write(nml_path, force=True)
+
+    def calc_runtime(self, setup_nml):
```
anton-seaice, line 212:
```suggestion
    def _calc_runtime(self, setup_nml):
```

preceding underscore to indicate its a "private" function (even though python doesn't do private)

payu/models/cice.py
```diff
@@ -201,9 +194,55 @@ def setup(self):
...
+        """
+        Calculate 1: the previous number of timesteps simulated, and 2:
+        the number of timesteps to simulate in the next run.
+
+        Note 1: This method is overridden in the cice5 driver, as cice5
+        in ACCESS OM2 does not require the calculated runtime information.
+        Instead it performs the calculations within the model using the
+        restart files and OM2 namelist.
```
anton-seaice, line 220:
```suggestion
        Note 1: This method is overridden in the cice5 driver.
```

I think this is enoguh detail?

anton-seaice (commented):
I will keep reviewing tomorrow :)

test/models/test_cice.py
```diff
@@ -121,24 +151,202 @@ def test_setup(ice_history):
...
-    with cd(expt_workdir):
-        os.remove(cice_nml["setup_nml"]["pointer_file"])
-        os.rmdir(cice_nml["setup_nml"]["restart_dir"])
+
+def test_no_restart_ptr(cice_config_files):
+    """
+    Test that payu raises an error if no prior restart path is specified,
+    restart is `true` in cice_in.nml, and the restart pointer is missing.
+    """
+    cice_in_patch = {"setup_nml": {"restart": True}}
```
anton-seaice, line 175:
```suggestion
    cice_in_patch = {"setup_nml": {"restart": ".true."}}
```

unless f90nml is doing this substition automatically

test/models/test_cice.py
```diff
@@ -121,24 +151,202 @@ def test_setup(ice_history):
...
     # Check dump_last doesn't exist
     with pytest.raises(KeyError, match="dump_last"):
         input_nml["setup_nml"]["dump_last"]
 
-    # cleanup
-    with cd(expt_workdir):
-        os.remove(cice_nml["setup_nml"]["pointer_file"])
-        os.rmdir(cice_nml["setup_nml"]["restart_dir"])
+
+def test_no_restart_ptr(cice_config_files):
```
anton-seaice, line 170: Maybe you are right this should be a seperate issue, because ideally we would cover all the variations in the comment below right? 

e.g.

prior restart path specified/ not specified
restart from nml is true/false
restart ptr missing/not missing

which is possible 6 permutations

anton-seaice (commented):
Is it possible to set-up the cice5 tests to import more code from `test_cice.py` instead of copying it ? There may be enough differences its not worth doing ?

test/models/test_cice.py
```diff
@@ -5,32 +5,41 @@
...
     "setup_nml": {
         "history_dir": "./HISTORY/",
         "restart_dir": "./RESTART/",
-        "year_init": 9999,
-        "days_per_year": 360,
+        "year_init": DEFAULT_YEAR_INIT,
+        "days_per_year": 365,
         "ice_ic": "default",
-        "restart": ".false.",
+        "restart": False,
```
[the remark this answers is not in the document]
blimlim: Oh yeah, this was something annoying with `f90nml` which was causing problems with the test namelists:
```
>>> import f90nml
>>> mynml = {"section": {"falsevar1":".false.", "falsevar2":False}}
>>> f90nml.write(mynml, "mynml.nml")
>>> readnml = f90nml.read("mynml.nml")
>>> if readnml["section"]["falsevar1"]:
...    print("falsevar1 not False")
falsevar1 not False
>>> if readnml["section"]["falsevar2"]:
...    print("falsevar2 not False")
... 

```

It looks like it encloses ".true." and ".false." with quote marks in the written files, making it read them back in as strings instead of booleans.

test/models/test_cice.py
blimlim: Same issue here with the f90nml substitution

test/models/test_cice.py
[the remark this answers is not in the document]
anton-seaice: looks like f90nml is handling this, so what you have is correct :)

https://f90nml.readthedocs.io/en/latest/usage.html#f90nml.parser.Parser.strict_logical

jo-basevi (commented):
The documentation of the code was easy to follow, so great for me who does not have much knowledge of these model drivers. I only have some fairly minor comments so far to add to this review

payu/models/cice.py
```diff
@@ -249,13 +283,11 @@ def setup(self):
...
-        # Now write out new run start date and total runtime.
+        # Add the prior runtime and new runtime to the working copy of the
+        # CICE namelist.
         setup_nml['npt'] = run_runtime / setup_nml['dt']
-        assert(total_runtime % setup_nml['dt'] == 0)
-        setup_nml['istep0'] = int(total_runtime / setup_nml['dt'])
-
-        nml_path = os.path.join(self.work_path, self.ice_nml_fname)
-        self.ice_in.write(nml_path, force=True)
+        assert (prior_runtime_seconds % setup_nml['dt'] == 0)
```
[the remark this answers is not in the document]
blimlim: Good question, I'm not sure what its checking for either. I think you're right that the only way it could fail is if `dt` was changed, but then it wouldn't even reliably fail depending on the new value.

Do you think it's worth guarding against changes to `dt` at all? E.g. would it ever be valid for someone to try and set up ESM with a different ocean/ice timestep?

payu/models/cice.py
blimlim: Yeah I think so! I'll commit this along with the other suggestions.

payu/models/cice.py
blimlim: Good idea! I'll commit this outside of the github editor to add in the changes to the method calls.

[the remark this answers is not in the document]
blimlim: Yeah I think this suggestion is clearer than passing it as an argument

payu/models/cice.py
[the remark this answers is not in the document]
anton-seaice: Actually I think it is valid to change the timestep, it might get used if someone has messed with some forcing / boundary condition and then had issues with the model being unstable. Reducing the timestep can help get past instabilities.

payu/models/cice.py
```diff
@@ -185,8 +179,7 @@ def setup(self):
                 iced_restart_file = sorted(iced_restart_files)[-1]
 
             if iced_restart_file is None:
-                print('payu: error: No restart file available.')
-                sys.exit(errno.ENOENT)
+                raise RuntimeError('payu: error: No restart file available.')
```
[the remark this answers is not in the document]
blimlim: `FileNotFoundError` sounds good! I've added it in f5ab3b4.

test/models/test_cice.py
```diff
@@ -121,24 +151,202 @@ def test_setup(ice_history):
...
+        cice_in_default.patch(cice_in_patch)
+        cice_in_default.write(CICE_NML_NAME, force=True)
+
+    with cd(ctrldir):
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
+        expt = payu.experiment.Experiment(lab, reproduce=False)
+        model = expt.models[0]
+
+        # Function to test
+        with pytest.raises(RuntimeError):
```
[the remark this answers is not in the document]
blimlim: Good idea! I've added a `match` argument in f5ab3b4

payu/models/cice.py
blimlim: Fixed up in f5ab3b4

blimlim:
> I think cloning wouldn't be necessary in these tests, as when a restart is added through cloning, payu just sets the restart option to config.yaml

Awesome thank you! I've removed the cloning steps from the tests, so that they now just test setting up the cice models with and without restart paths specified in the `config.yaml`.

> Is it possible to set-up the cice5 tests to import more code from test_cice.py instead of copying it ? There may be enough differences its not worth doing ?

Good question, I think the tests are very similar between the two cice versions. Now that the tests have been updated, I'll test this out in a separate branch to see how sharing tests between the two might look.

payu/models/cice.py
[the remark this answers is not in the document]
anton-seaice: Needs a different PR 😅. Well just to check if it would work in reality

blimlim:
Ok hmm I'm finding it a bit challenging to share the test code between the cice 4 and 5 tests.  I've temporarily uploaded a (not quite working)  attempt here: 377cd5f

My idea was to build a base test class, which separate `TestCice4Setup` and `TestCice5Setup` classes added specific details to. However the two cases were doing a few different things with their assertions, and my attempts at handling this got a bit messy. The bigger problem I'm having though is correctly handling all the fixtures and constants for the two cases. at the moment the tests are failing as they're mixing correct and incorrect ones. I think it should be possible to come up with something better, but it might require a bit of a rework of the pre-test setup. 

Let me know if you think the (attempted) changes do look any cleaner though, and if it might be worth trying to get them working.

payu/models/cice.py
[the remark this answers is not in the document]
blimlim: Good point! So maybe we just keep the existing behaviour for now and work out whether to change it in a separate issue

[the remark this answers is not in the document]
anton-seaice: Yeah I think so, it would be nice to get rid of the `assert` but also it probably doesn't break anything so lets leave it there

anton-seaice:
Happy to take your advice on whether moving to more generalised tests add benefit or confusion :)

blimlim:
I found it a bit difficult (and in the end wasn't successful) in getting the imported tests to use the correct versions of constants and fixtures, and so I'm thinking for the moment it might be easiest to keep the tests separate.

I've just added a couple of small fixes: Re-adding in the `assert (prior_runtime_seconds % setup_nml['dt'] == 0)`, testing that run time calculations are correct for a few different values (previously I was just testing one), and some smaller cleaning up, and so I think it should be ready for final review :)

payu/models/cice.py
```diff
@@ -201,9 +195,50 @@ def setup(self):
...
+
+        if self.prior_restart_path:
+
             prior_nml_path = os.path.join(self.prior_restart_path,
                                           self.ice_nml_fname)
 
+            # TODO: Are there any models which leave cice_in.nml in
+            # the output rather than restart directory? If not we can remove
+            # this.
+
```
anton-seaice, line 241:
```suggestion
```

payu/models/cice5.py
```diff
@@ -71,3 +71,10 @@ def set_access_timestep(self, t_step):
         # TODO: Figure out some way to move this to the ACCESS driver
         # Re-read ice timestep and move this over there
         self.set_local_timestep(t_step)
+
+    def _calc_runtime(self):
+        """
+        Overrides the cice driver method, as CICE5 in OM2 does not use
```
anton-seaice, line 77:
```suggestion
        Overrides the cice driver method, as CICE5 can store the timing information in restart files does not use
```

payu/models/cice.py
```diff
@@ -221,22 +256,17 @@ def setup(self):
...
             # The total time in seconds since the beginning of the experiment
-            total_runtime = prior_setup_nml['istep0'] + prior_setup_nml['npt']
-            total_runtime = total_runtime * prior_setup_nml['dt']
-        else:
-            # Locate and link any restart files (if required)
-            if not setup_nml['ice_ic'] in ('none', 'default'):
-                self.link_restart(setup_nml['ice_ic'])
-
-            if setup_nml['restart']:
-                self.link_restart(setup_nml['pointer_file'])
```
anton-seaice, line 232: Why are these removed ?

blimlim: Ah yep, it's not very clear in the diff but these have just been moved up to L198-204 to make them run outside of the new `_calc_runtime()` method.

anton-seaice: Oh great :)

anton-seaice (commented):
I pip installed this branch. Then I cloned ACCESS-OM2 

`payu clone -b om2_1deg_ryf -B release-1deg_jra55_ryf https://github.com/ACCESS-NRI/access-om2-configs om2`

and changed the run length to one month. And did a payu run. - Success

Then I did payu sweep && run again

I also did payu clone again 

`payu clone om2 om2-clone-n`

payu sweep && run. - Success (This started again from 1900-01-01 - I assume that is correct.)

I also did a restart clone

`payu clone -r om2/archive/restart000 om2 om2-clone-r`

Which gave me 

UserWarning: Given restart path om2/archive/restart000 does not exist. Skipping setting 'restart' in config file ??

`payu clone -r /g/data/tm70/as2285/payu/om2/archive/restart000 om2 om2-clone-r`

worked and did payu run - Success (run started from 1900-Feb-01)

anton-seaice (approved):
Thanks @blimlim - this looks good.

Confirm with @aidanheerdegen before merging

jo-basevi (approved):
The changes look good to me

aidanheerdegen (approved):
Looks great. Thanks for all the work implementing and testing @blimlim, and reviewing @anton-seaice and @jo-basevi.

blimlim:
Thanks for the reviews and help everyone!

PR #511, opened 2024-09-16
Add minimum payu version check 

### LINKED ISSUE #147: Require payu version
Since the issue with some buggy versions, how about adding a minimum version requirement?

I can see this is also useful when doing things like changing the collate config layout, so it is no longer compatible with earlier versions. So have a config flag that sets the minimum version required?

This is most useful for @nichannah and @aekiss when creating configs for COSIMA that will be widely shared. It allows them to use the latest features and be confident that they will be supported by the payu version someone is running, or tell them to upgrade.

jo-basevi:
Add a `payu_minimum_version` option to `config.yaml`. This runs a check at the start of the `Experimet.setup` comparing the current payu version and raises a `RuntimeError` if version is earlier than the configured minimum version.

I added a dependency on the `packaging` lib to parse and compare versions, which seems to the recommended library for parsing version strings (https://stackoverflow.com/a/11887885, https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools). I could do some regex parsing, but this library can handle already handles versions like `1.0`, `v0.1`, `1.0.0+4.gabc1234`  etc. So hopefully it'll be less buggy and will be able to support future version formats..

I added the check to the start of the setup method to the experiment class which seemed the logical place for it at the moment. So unfortunately it's adding more logic to an already long file. I thought about adding a `setup.py` file for setup helper methods, but I am not sure whether that'll make it better having one file for one small function?

Closes #147

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -23,6 +23,7 @@ from pathlib import Path
 
 # Extensions
 import yaml
+from packaging import version
 
 # Local
 from payu import envmod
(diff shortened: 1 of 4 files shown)


payu/experiment.py
```diff
@@ -387,7 +388,31 @@ def build_model(self):
...
+            # Skip version check
+            return
+
+        minimum_version = str(self.config['payu_minimum_version'])
+
+        # Get the current version of the package
+        current_version = payu.__version__
+
+        # Compare versions
+        if version.parse(current_version) < version.parse(minimum_version):
```
aidanheerdegen, line 406: Maybe  do`version.parse(minimum_version)` when `minimum_version` is set above and wrap in a `try/except` block and tell the user they've stuffed up their version  specification before throwing an inscrutable python exception.

aidanheerdegen (changes requested):
Just a single change, but feel free to push back if you disagree.

payu/experiment.py
jo-basevi: Good point, I've moved the parsing earlier with hopefully a more helpful error message in a6bed42. I've added a couple tests with invalid versions.

aidanheerdegen (approved):
Looks great! Thanks Jo. merge if you're happy to do so.

Issue #405, opened 2024-01-22
Missing origin repo throws error


aidanheerdegen:
Using the new version (1.1) version of `payu` with an existing configuration repo threw the following error with `payu checkout` because the `origin` repo no longer exists:
```
Traceback (most recent call last):
  File "/g/data/vk83/apps/payu/1.1/bin/payu", line 10, in <module>
    sys.exit(parse())
  File "/home/502/aph502/code/python/payu/payu/cli.py", line 41, in parse
    run_cmd(**args)
  File "/home/502/aph502/code/python/payu/payu/subcommands/checkout_cmd.py", line 34, in runcmd
    checkout_branch(is_new_branch=new_branch,
  File "/home/502/aph502/code/python/payu/payu/branch.py", line 145, in checkout_branch
    repo.checkout_branch(branch_name, is_new_branch, start_point)
  File "/home/502/aph502/code/python/payu/payu/git_utils.py", line 147, in checkout_branch
    remote_branches = self.remote_branches_dict()
  File "/home/502/aph502/code/python/payu/payu/git_utils.py", line 127, in remote_branches_dict
    remote.fetch()
  File "/g/data/vk83/apps/payu/1.1/lib/python3.9/site-packages/git/remote.py", line 1014, in fetch
    res = self._get_fetch_info_from_stderr(proc, progress, kill_after_timeout=kill_after_timeout)
  File "/g/data/vk83/apps/payu/1.1/lib/python3.9/site-packages/git/remote.py", line 853, in _get_fetch_info_from_stderr
    proc.wait(stderr=stderr_text)
  File "/g/data/vk83/apps/payu/1.1/lib/python3.9/site-packages/git/cmd.py", line 600, in wait
    raise GitCommandError(remove_password_if_present(self.args), status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git fetch -v -- origin
  stderr: 'fatal: '/home/502/aph502/1deg_jra55_ryf' does not appear to be a git repository
fatal: Could not read from remote repository.'
```

Command run:
```
payu checkout timing-test
```

Output of `git remote -v`
```
origin  /home/502/aph502/1deg_jra55_ryf (fetch)
origin  /home/502/aph502/1deg_jra55_ryf (push)
```
But this directory (repo) no longer exists:
```
ls: cannot access '/home/502/aph502/1deg_jra55_ryf': No such file or directory
```

[...]

PR #512, opened 2024-09-16
 Fix missing origin repository bug

jo-basevi:
In #405, `payu checkout` raises an error when it is no longer able to fetch branches from the remote repository. This happens when cloning from a remote repository using a path, and then that path gets deleted at a later point of time.

I changed it to catch the exception, and to display a warning to the user instead that payu was not able to fetch from the remote repository.

### DIFF
--- a/payu/git_utils.py
+++ b/payu/git_utils.py
@@ -129,7 +129,10 @@ class GitRepository:
             except git.exc.GitCommandError:
                 warnings.warn(
                     f"Failed to fetch from remote repository: {remote.name} " +
-                    f"(url: {remote.url})",
+                    f"(url: {remote.url}). Payu is not able to determine " +
+                    "remote branch names, which are used in payu checkout " +
+                    "to check if a branch already exists, or when creating " +
+                    "a new branch from a remote branch.",
                     PayuGitWarning
                 )
                 continue
--- a/payu/git_utils.py
+++ b/payu/git_utils.py
@@ -124,7 +124,16 @@ class GitRepository:
         objects"""
         branch_names_dict = {}
         for remote in self.repo.remotes:
-            remote.fetch()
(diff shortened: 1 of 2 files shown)


jo-basevi:
Closes #405

payu/git_utils.py
```diff
@@ -124,7 +124,16 @@ def remote_branches_dict(self) -> Dict[str, git.Head]:
         objects"""
         branch_names_dict = {}
         for remote in self.repo.remotes:
-            remote.fetch()
+            try:
+                remote.fetch()
+            except git.exc.GitCommandError:
+                warnings.warn(
+                    f"Failed to fetch from remote repository: {remote.name} " +
```
aidanheerdegen, line 131: I know it might make it a bit verbose, but is it worth explaining why `payu` was trying to do this so the user knows they an either safely ignore it, or if fix it if they want that functionality?

aidanheerdegen (approved):
LGTM, just a question as per usual.

aidanheerdegen (dismissed):
LGTM. Thanks!

jo-basevi:
Sorry I had small spacing error! Though it did get me thinking whether it should still use that last saved changes from the remote repository (whether to remove the `continue` on the loop through remotes).. But if it can't fetch the latest updates it probably should not be able to checkout a remote branch? So maybe fine to leave as is

aidanheerdegen (approved):
LGTM.

aidanheerdegen:
> Though it did get me thinking whether it should still use that last saved changes from the remote repository (whether to remove the `continue` on the loop through remotes).. But if it can't fetch the latest updates it probably should not be able to checkout a remote branch? So maybe fine to leave as is

Yeah I think that is the cleanest approach.

PR #515, opened 2024-09-18
Add support for cloning configuration using a commit hash

### LINKED ISSUE #503: Support for cloning configurations from a git commit hash
I am wondering if Payu supports cloning a specific configuration directly from a git commit hash rather than from a branch? This feature would be helpful for reproducing experiments from precise commits in the repo's history.

jo-basevi:
Added `--start-point/-s` argument to `payu clone` to specify what commit/tag/branch name to start new branches from. 

So there is still `-B/--branch` flag which initially checks out a branch during the `git clone` command. This was initially done (I think?) to avoid checking out a default branch first. For example, the [access-om2-configs repository](https://github.com/ACCESS-NRI/access-om2-configs/), the `main` branch is not a payu configuration.

The start point is used similarly to the `payu checkout`'s  positional argument, so 

 ```
payu clone -b ${NEW_BRANCH_NAME} -s ${COMMIT_OR_TAG} ${REPOSITORY} ${DIR}
```

is equivalent to running 

```
payu clone ${REPOSITORY} ${DIR}
payu checkout -b ${NEW_BRANCH_NAME} ${COMMIT_OR_TAG}
```

Closes #503

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -128,7 +128,7 @@ new branch name::
 To create a new git branch starting from a tag or commit, use ``-s/--start-point``
 flag::
 
-      payu clone -b ${NEW_BRANCH} -s {COMMIT_HASH} ${REPOSITORY} my_expt
+      payu clone -b ${NEW_BRANCH} -s {COMMIT_HASH|TAG} ${REPOSITORY} my_expt
 
 To see more configuration options for ``payu clone``, 
 run:: 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -115,15 +115,20 @@ For example::
 Where ``${REPOSITORY}`` is the git URL or path of the repository to clone from, 
 for example, https://github.com/payu-org/mom-example.git.
 
-To clone and checkout an existing git branch, use the ``--branch`` flag and 
+To clone and checkout an existing git branch, use the ``-B/--branch`` flag and
 specify the branch name::
 
       payu clone --branch ${EXISTING_BRANCH} ${REPOSITORY} my_expt
 
-To create and checkout a new git branch use ``--new-branch`` and specify a 
-new branch name:
+To create and checkout a new git branch use ``-b/--new-branch`` and specify a
+new branch name::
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: payu/subcommands/args.py [hunk 1]
```diff
@@ -191,6 +191,17 @@
...
+# Clone branch
+clone_start_point = {
+    'flags': ('--start-point', '-s'),
+    'parameters': {
+        'action':   'store',
+        'dest': 'start_point',
+        'default':  None,
+        'help': 'New branch will start from this commit or tag'
+    }
+}
```
aidanheerdegen, line 203: Is there a reason not to re-use the existing `start_point` option for this case?

docs/source/usage.rst
```diff
@@ -115,16 +115,21 @@ For example::
...
-new branch name:
+To create and checkout a new git branch use ``-b/--new-branch`` and specify a
+new branch name::
 
       payu clone --new-branch ${NEW_BRANCH} ${REPOSITORY} my_expt
 
+To create a new git branch starting from a tag or commit, use ``-s/--start-point``
+flag::
+
+      payu clone -b ${NEW_BRANCH} -s {COMMIT_HASH} ${REPOSITORY} my_expt
```
aidanheerdegen, line 131: Worth adding another example with `-s {TAG}`? Could use `-s {COMMIT_HASH|TAG}` if you think it is clear enough.

aidanheerdegen (commented):
I have a question and suggestion.

payu/subcommands/args.py
jo-basevi: Yes, for the checkout command `start_point` is an optional positional argument (to be more similar to `git checkout -b NEW_BRANCH  START_POINT`)

docs/source/usage.rst
jo-basevi: Yeah I think `{COMMIT_HASH|TAG}` is clear enough so I've added it in 8c9dcaa

aidanheerdegen (approved):
LGTM

jo-basevi:
Thanks for the review!

PR #518, opened 2024-09-20
Support symbolic links in input files

### LINKED ISSUE #385: Support symbolic links in inputs
Currently using symbolic links in `input` in `config.yaml` does not fully resolve the target of the link when making a link in the `work` directory during [setup](https://github.com/payu-org/payu/blob/master/payu/experiment.py#L378).

For examples, making a symbolic link like this in the control dir:
```
inputs -> /g/data/vk83/experiments/inputs/access-om2
```
and then referencing `inputs` in the `input` path:
```yaml
          - ./inputs/ocean/chlorophyll/global.1deg/2020.05.30/chl.nc
```
leads to this symbolic link in `work`
```
lrwxrwxrwx 1 aph502 tm70 56 Nov 22 16:43 chl.nc -> ./inputs/ocean/chlorophyll/global.1deg/2020.05.30/chl.nc
```
The manifest also has the incorrect `fullpath`
```yaml
work/ocean/INPUT/chl.nc:
  fullpath: ./inputs/ocean/chlorophyll/global.1deg/2020.05.30/chl.nc
```

Why is this important? 

Being able to utilise symlinks in the control experiment directory can remove a lot of repeated text in `config.yaml`, which makes the configuration more legible and understandable. Also putting a symlink in the control directory makes the inputs discoverable easily from within the control directory, just as `archive` and `work` are.

Note I am not suggesting `payu` makes an `input` directory symlink, simply that it support such usage.

### LINKED ISSUE #450: Support symbolic links to configuration files
For ACCESS models it would be convenient to have a number of possible pre-set diagnostic output profiles and use symbolic links to select the desired output profile.

Currently this is not supported correctly with `payu`. Configuration files are copied into the `work` directory here

https://github.com/payu-org/payu/blob/master/payu/models/model.py#L234

and

https://github.com/payu-org/payu/blob/master/payu/models/model.py#L239

According to the [documentation of the `shutil.copy` function](https://docs.python.org/3/library/shutil.html#shutil.copy) this will copy a symbolic link rather than the file it points to.

This is undesirable behaviour for `payu`: if it is a symbolic link we want a copy of the file in `work`. This can be achieved by adding the `follow_symlinks=True` option to the `copy` command.

jo-basevi:
This PR:
 - adds a step to model setup to resolve the input directory paths to absolute paths and resolve symlinks.
-  pass `follow_symlink=True` to copy commands when copying config files. However this is already a default value but I added in so it clear in the code that it's intentional.
- more tests for the above

In 0daf9115b65468e05f1db4bd8285fdc9a896c923, I had just one test that would run `payu setup` once and test that inputs, configuration files and executables that were set up as expected. In 6d975159f4102ac184ab13544f1dbbb451e2225e, I split up the tests to try(?) make it simpler to read and have one test for one thing. My concern was that it would make running tests slower as were running a `payu setup` multiple times. It ended up only adding a couple seconds to the tests..

To check how current manifests are changed due to these changes, I ran `payu setup` for `release-1deg_jra55_ryf` branch in `access-om2-configs`(https://github.com/ACCESS-NRI/access-om2-configs/tree/release-1deg_jra55_ryf). One of the full path changes are in `input.yaml` manifest are:

```
 work/INPUT/JRA55_MOM1_conserve2nd.nc:
-  fullpath: /g/data/vk83/experiments/inputs/access-om2/remapping_weights/JRA55/global.1deg/2020.05.30/JRA55_MOM1_conserve2nd.nc
+  fullpath: /g/data/vk83/configurations/inputs/access-om2/remapping_weights/JRA55/global.1deg/2020.05.30/JRA55_MOM1_conserve2nd.nc
   hashes:
-    binhash: abe66e4b33cbb75d25c010ea5f96f053
+    binhash: 3ba86d1ae2c7641a29dce1179c55ad64
     md5: 00be8ec0f1126055b65dc68178ce4eb5
```

(as `/g/data/vk83/experiments/` is a symlink for `/g/data/vk83/configurations`)

Closes #385
Closes #450

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -144,7 +144,7 @@ configuration.
 ``input``
    Listing of the directories containing model input fields, linked to the
    experiment during setup. This can either be the name of a directory in the
-   laboratory's ``input`` directory::
+   control directory or in laboratory's ``input`` directory::
 
       input: core_inputs
 
--- a/payu/models/model.py
+++ b/payu/models/model.py
@@ -3,7 +3,7 @@
 :copyright: Copyright 2011 Marshall Ward, see AUTHORS for details
 :license: Apache License, Version 2.0, see LICENSE for details
 """
-import errno
+
 import os
 import shutil
 import shlex
@@ -129,14 +129,20 @@ class Model(object):
         self.input_paths = []
         for input_dir in input_dirs:
 
-            # First test for absolute path
+            # First test if path exists
             if os.path.exists(input_dir):
+                # Resolve symlinks and relative paths
+                input_dir = os.path.realpath(input_dir)
+
                 self.input_paths.append(input_dir)
             else:
                 # Test for path relative to /${lab_path}/input/${model_name}
                 assert self.input_basepath
                 rel_path = os.path.join(self.input_basepath, input_dir)
                 if os.path.exists(rel_path):
+                    # Resolve symlinks
+                    rel_path = os.path.realpath(rel_path)
+
                     self.input_paths.append(rel_path)
                 else:
                     sys.exit('payu: error: Input directory {0} not found; '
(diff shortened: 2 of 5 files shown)


aidanheerdegen (approved):
LGTM. Thanks!

PR #519, opened 2024-09-22
Update docs on model executable loaded via environment modules

### LINKED ISSUE #517: Update docs with model exe configuration options
Documentation for configuring model executables needs to be updated to reflect executables found using loaded environment modules (see `exe` in https://payu.readthedocs.io/en/latest/config.html#model)

jo-basevi:
Updated documentation that model executables can be specified by filenames, if found in paths loaded by model environment modules.

Closes #517

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -163,9 +163,12 @@ configuration.
    directory of the list takes precedence.
 
 ``exe``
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
LGTM

Issue #527, opened 2024-10-11
Compressing model log files


blimlim:
Model log files can take up a significant amount of storage. For a single year of ESM1.5 simulation, the sub-models produce:
- Atmosphere: ~50M
- Ocean: ~60M
- Ice: ~900M
which can add up over long simulations.

[Previous discussions](https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/32#issuecomment-2287723927) raised the idea of compressing the log files during the `archive` step. @aidanheerdegen mentioned a good place to start would be the CICE driver, and if successful it could be expanded to a general (perhaps optional) step across models. 

A couple of implementation questions/concerns:
- Would it be preferable to compress each log file separately, or combine all the logs for a single sub-model into a `tar.gz`?
- People may have workflows set up for extracting information from the logs (e.g. subroutine timing information) which would be affected.

aidanheerdegen:
It is possible to `grep` into a compressed `tar` archive file:

```bash
$ ls
cice_in.nml    ice_diag.d    iceout085  iceout087  iceout089  iceout091  iceout093  iceout095  input_ice.nml                                      
debug.root.03  ice_diag_out  iceout086  iceout088  iceout090  iceout092  iceout094  iceout096
$ grep 'Timer  15:  from_ocn' *
ice_diag.d:Timer  15:  from_ocn       7.83 seconds
$ tar -zcvf logs.tar.gz *
cice_in.nml
debug.root.03
ice_diag.d
ice_diag_out
iceout085
iceout086
iceout087
iceout088
iceout089
iceout090
iceout091
iceout092
iceout093
iceout094
iceout095
iceout096
input_ice.nml
$ zgrep -a 'Timer  15:  from_ocn' logs.tar.gz                                                                           
Timer  15:  from_ocn       7.83 seconds
```

So I think a compressed tar archive makes most sense.

The question of what to archive? Probably best to have a list in the model driver of patterns to match files. Simple case would be just anything that `beginswith` a list of strings. 

For `cice5` it would be
```python
log_files = ['iceout', 'ice_diag', 'debug.root']
```

Which isn't too bad.

For `mom5` (in ESM1.5)
```python
log_files = ['fort', 'logfile', 'debug.root', 'oceout']
```

For `atmosphere` (in ESM1.5)
```python
log_files = ['atm.fort', 'fort', 'nout', 'debug.root']
```
Are the various `hist` files also log files? (`xhist`, `thist`, `ihist`)

PR #532, opened 2024-10-16
Model log compression

blimlim:
This draft PR addresses #527.

It adds a `compress_log_files` method, initially just to the `cice.py` driver. This  is called during the `archive` step and compresses all the log files matching some specified regex patterns into a tarball. A `compress_logs` option in the `config.yaml` file controls whether to run the compression.

I've marked this PR as a draft, as I'd be keen to see if the implementation looks reasonable to others, and if so I think it would be good to move the changes over to the general `model` class making it available to each model.

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -276,10 +276,11 @@ section for details.
 Archiving
 ---------
 
-``archiving``
+``archive``
       On completion of a model run, payu moves model output, restart, and log
       files from the temporary work area to the experiment archive directory.
       The following settings control the steps taken during the archive step:
+
       ``enable`` (*Default:* ``True``)
             Flag to enable/disable the archive step. If ``False`` all output, restart,
             and log files will remain in the work directory, and any collation, post-processing,
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -503,7 +503,7 @@ class Experiment(object):
 
         # Check restart pruning for valid configuration values and
         # warns user if more restarts than expected would be pruned
-        if self.config.get('archive', True):
+        if self.archiving():
             self.get_restarts_to_prune()
 
     def run(self, *user_flags):
@@ -769,8 +769,25 @@ class Experiment(object):
         if run_script:
             self.run_userscript(run_script)
 
+    def archiving(self):
+        """
+        Determine whether to run archive step based on config.yaml settings.
+        Default to True when archive settings are absent.
+        """
+        archive_config = self.config.get('archive', {})
+        if isinstance(archive_config, dict):
+            return archive_config.get('enable', True)
+
+        # Backwards compatibility for configs with boolean archive setting
+        elif isinstance(archive_config, bool):
+            return archive_config
+
+        else:
+            msg = "Incorrect format for archive settings in config.yaml"
+            raise RuntimeError(msg)
+
     def archive(self, force_prune_restarts=False):
-        if not self.config.get('archive', True):
+        if not self.archiving():
             print('payu: not archiving due to config.yaml setting.')
             return
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -775,16 +775,7 @@ class Experiment(object):
         Default to True when archive settings are absent.
         """
         archive_config = self.config.get('archive', {})
-        if isinstance(archive_config, dict):
-            return archive_config.get('enable', True)
-
-        # Backwards compatibility for configs with boolean archive setting
-        elif isinstance(archive_config, bool):
-            return archive_config
-
-        else:
-            msg = "Incorrect format for archive settings in config.yaml"
-            raise RuntimeError(msg)
+        return archive_config.get('enable', True)
 
     def archive(self, force_prune_restarts=False):
         if not self.archiving():
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -367,8 +367,7 @@ class Cice(Model):
         """
         log_files = []
         for filename in os.listdir(self.work_path):
-            if any((re.match(pattern, filename)
-                    for pattern in self.logs_to_compress)):
+            if re.match("|".join(self.logs_to_compress), filename):
                 log_files.append(os.path.join(self.work_path, filename))
         return log_files
 
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -342,19 +342,10 @@ class Cice(Model):
         else:
             shutil.rmtree(self.work_input_path)
 
-        if self.compression_enabled():
-            self.compress_log_files()
-
-    def compression_enabled(self):
-        """
-        Determine whether to run log compression based on config.yaml settings.
-        Default to True when 'compress_logs' setting is absent.
-        """
         archive_config = self.expt.config.get('archive', {})
-        if isinstance(archive_config, dict):
-            return archive_config.get('compress_logs', True)
-        else:
-            return True
+        compressing_logs = archive_config.get('compress_logs', True)
+        if compressing_logs:
+            self.compress_log_files()
 
     def get_log_files(self):
         """
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -381,7 +381,10 @@ class Cice(Model):
                           mode="w:gz") as tar:
             for file in log_files:
                 tar.add(file, arcname=os.path.basename(file))
-                os.remove(file)
+
+        # Delete files after tarball is written
+        for file in log_files:
+            os.remove(file)
 
     def collate(self):
         pass
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -17,6 +17,8 @@ import os
 import sys
 import shutil
 import datetime
+import re
+import tarfile
 
 # Extensions
 import f90nml
@@ -50,6 +52,13 @@ class Cice(Model):
         self.set_timestep = self.set_local_timestep
 
         self.copy_inputs = False
+
+        # regex patterns for matching log files. When empty, no logs compressed
+        self.logs_to_compress = [r"iceout[0-9]{3}",
+                                 r"debug\.root\.[0-9]{2}",
+                                 r"ice_diag\.d",
+                                 r"ice_diag_out"]
+        self.log_tar_name = "logfiles.tar.gz"
 
     def set_model_pathnames(self):
         super(Cice, self).set_model_pathnames()
@@ -332,6 +341,25 @@ class Cice(Model):
                     os.remove(os.path.join(self.restart_path, f))
         else:
             shutil.rmtree(self.work_input_path)
+
+        if self.expt.config.get('compress_logs', False):
+            self.compress_log_files()
+
+    def get_log_files(self):
+        log_files = []
+        for filename in os.listdir(self.work_path):
+            if any((re.match(pattern, filename)
+                    for pattern in self.logs_to_compress)):
+                log_files.append(os.path.join(self.work_path, filename))
+        return log_files
+
+    def compress_log_files(self):
+        log_files = self.get_log_files()
+        with tarfile.open(name=os.path.join(self.work_path, self.log_tar_name),
+                          mode="w:gz") as tar:
+            for file in log_files:
+                tar.add(file, arcname=os.path.basename(file))
+                os.remove(file)
 
     def collate(self):
         pass
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -342,10 +342,29 @@ class Cice(Model):
         else:
             shutil.rmtree(self.work_input_path)
 
-        if self.expt.config.get('compress_logs', False):
+        if self.compression_enabled():
             self.compress_log_files()
 
+    def compression_enabled(self):
+        """
+        Determine whether to run log compression based on config.yaml settings.
+        Default to True when 'compress_logs' setting is absent.
+        """
+        archive_config = self.expt.config.get('archive', {})
+        if isinstance(archive_config, dict):
+            return archive_config.get('compress_logs', True)
+        else:
+            return True
+
     def get_log_files(self):
+        """
+        Find model log files in the work directory based on regex patterns
+        in self.logs_to_compress.
+
+        Returns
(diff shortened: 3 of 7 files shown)


test/models/test_cice.py
```diff
@@ -124,41 +125,47 @@ def empty_workdir():
...
+@pytest.fixture
+def cice_nml():
+    nml_path = os.path.join(ctrldir, CICE_NML_NAME)
+    f90nml.write(DEFAULT_CICE_NML, nml_path)
+
+    yield nml_path
+
+    # Cleanup
+    os.remove(nml_path)
+
```
blimlim, line 144: I moved the default cice namelist into a separate fixture here, as the new tests don't need to be repeated for each of the cases in the previous `cice_config_files` fixture.

test/models/test_cice.py
```diff
@@ -300,3 +307,85 @@ def test_no_restart_ptr(config, cice_config_files):
...
+    properly compressed into a tarball by cice.compress_log_files().
+    """
+    with cd(ctrldir):
+        # Initialise laboratory and experiment
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
+        expt = payu.experiment.Experiment(lab, reproduce=False)
+        model = expt.models[0]
+
+        # Function to test
+        model.compress_log_files()
```
blimlim, line 380: I originally hoped to test `model.archive()` here, which would allow us to test the cases where `compress_logs` is and isn't included in the `config.yaml`. However running the actual `cice.archive()` required several directories and files to be created that weren't really relevant to what I wanted to test, and I think they would have made the added tests fairly brittle. Hence I ended up just testing the `compress_log_files()` method directly.

If we move everything over to the `model.py` driver, I think it will be easier to run a more general test of `model.archive()` as there would be less requirements on the work directory.

test/models/test_cice.py
```diff
@@ -300,3 +307,85 @@ def test_no_restart_ptr(config, cice_config_files):
...
+    log_names = ["ice_diag_out", "ice_diag.d", "debug.root.03",
+                 "iceout086", "iceout088", "iceout090", "iceout092",
+                 "iceout094", "iceout096", "iceout085", "iceout087",
+                 "iceout089", "iceout091", "iceout093", "iceout095"]
+    log_paths = [os.path.join(expt_workdir, name) for name in log_names]
+
+    for log_file in log_paths:
+        with open(log_file, "w") as f:
+            f.close()
+
```
aidanheerdegen, line 337: We're already making good use of the [pathlib `touch()` function](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch)

https://github.com/search?q=repo%3Apayu-org%2Fpayu+touch&type=code

test/models/test_cice.py
```diff
@@ -300,3 +307,85 @@ def test_no_restart_ptr(config, cice_config_files):
...
+@pytest.fixture
+def cice4_log_files():
+    """
+    Create cice log files matching those produced during ESM1.5 simulations.
+    """
+    log_names = ["ice_diag_out", "ice_diag.d", "debug.root.03",
+                 "iceout086", "iceout088", "iceout090", "iceout092",
+                 "iceout094", "iceout096", "iceout085", "iceout087",
+                 "iceout089", "iceout091", "iceout093", "iceout095"]
+    log_paths = [os.path.join(expt_workdir, name) for name in log_names]
```
aidanheerdegen, line 332:
```python
log_names = ["ice_diag_out", "ice_diag.d", "debug.root.03"] + [f'iceout{x:03.0f}' for x in range(86,96)]
```
Just a bit more compact ... but maybe less readable?

test/models/test_cice.py
```diff
@@ -300,3 +307,85 @@ def test_no_restart_ptr(config, cice_config_files):
...
+@pytest.fixture
+def non_log_file():
+    """
+    Create a cice4 output file to be ignored by log compression.
+    Use cice_in.nml which is copied to the work directory in ESM1.5.
+    """
+    non_log_path = os.path.join(expt_workdir, CICE_NML_NAME)
+    with open(non_log_path, "w") as f:
+        f.close()
+
```
aidanheerdegen, line 357: See `.touch()` comment.

aidanheerdegen:
I'd be tempted to add `compress_log_files` as a sub-option to `archive`. Something like
```yaml
archive:
    enable: True
    compress_logs: True
```
and make `compress_logs` default to `True`.

Currently `archive` is a boolean, but we can support it being a dict as well. c.f. the `collate` option, which is converted to a `dict` from a boolean configuration option:

https://github.com/payu-org/payu/blob/7c885b855254ebb963c0a88d1f2ec306a026b71a/payu/fsops.py#L108-L112

It would also require changing the code to check if `archive` is enabled  to something like this

https://github.com/payu-org/payu/blob/7c885b855254ebb963c0a88d1f2ec306a026b71a/payu/experiment.py#L827-L828

test/models/test_cice.py
blimlim: I think that reads fine! I've added it in 0b2fcc9

test/models/test_cice.py
blimlim: Good idea, added in 0b2fcc9

test/models/test_cice.py
blimlim: Added in 0b2fcc9

blimlim:
> I'd be tempted to add compress_log_files as a sub-option to archive

I've had a go at implementing this in 0b2fcc9. See comments for further details!

payu/experiment.py
```diff
@@ -769,8 +769,25 @@ def run(self, *user_flags):
...
+            return archive_config.get('enable', True)
+
+        # Backwards compatibility for configs with boolean archive setting
+        elif isinstance(archive_config, bool):
+            return archive_config
+
+        else:
+            msg = "Incorrect format for archive settings in config.yaml"
+            raise RuntimeError(msg)
+
```
blimlim, line 788: I've added this method assuming we'll swap the `archive` boolean to a dict in `config.yaml` files, e.g:
```yaml
archive:
    enable: True
    compress_logs: True
```
The above block will retain backwards compatibility with configs using an `archive: <True/False>` boolean`, and will also default to `True` when no `archive` settings are included.

payu/models/cice.py
```diff
@@ -342,10 +342,29 @@ def archive(self, **kwargs):
...
+        """
+        Determine whether to run log compression based on config.yaml settings.
+        Default to True when 'compress_logs' setting is absent.
+        """
+        archive_config = self.expt.config.get('archive', {})
+        if isinstance(archive_config, dict):
+            return archive_config.get('compress_logs', True)
+        else:
+            return True
+
```
blimlim, line 358: Similar to the `experiment.archiving()` method. This tries to keep backwards compatibility with `archive` being given as a boolean in the `config.yaml`. The cases are:

- No `archive` setting -> default to `compression_enabled = True`
- `archive` given as a boolean -> default to `compression_enabled = True`
- `archive` given as a dictionary but missing a `compress_logs` setting -> default to `compression_enabled = True`
- `archive` given as a dictionary and `compress_logs` setting included -> use specified value.

Does this behaviour seem reasonable?

blimlim:
Currently cice4 in ESM1.5 is the main model causing problems, but there are a couple of options for generalising this, making it available to other models:
1. Just move the three methods [`cice.compression_enabled()`](https://github.com/ACCESS-NRI/payu/blob/0b2fcc957f736aac1762026f24184601596c70ac/payu/models/cice.py#L348), [`cice.get_log_files()`](https://github.com/ACCESS-NRI/payu/blob/0b2fcc957f736aac1762026f24184601596c70ac/payu/models/cice.py#L359), [`cice.compress_log_files()`](https://github.com/ACCESS-NRI/payu/blob/0b2fcc957f736aac1762026f24184601596c70ac/payu/models/cice.py#L375-L384) from the `cice.py` driver to the `model.py` driver, and add `self.logs_to_compress = []` to the general `model.__init__`.  Compression can be added to other models by adding the required regex patterns to `self.logs_to_compress = []` in the relevant model driver.

2. Leave the methods in `cice.py`, add methods with the same name to `model.py` that just `pass`. Compression can then be implemented in other models by overriding these methods in the relevant driver.

I think the methods should work generally enough that option 1 would work, and I think this would be a bit simpler/easier to follow. Keen to hear other ideas or opinions though!

blimlim:
@aidanheerdegen I've just marked this one as ready to review! I think the main outstanding question is whether we are happy with the log compression being implemented in the `cice.py` driver to start off with, or if we prefer to have it more general.

payu/models/cice.py
aidanheerdegen: Yes. Very sensible approach.

payu/models/cice.py
```diff
@@ -333,6 +342,47 @@ def archive(self, **kwargs):
...
+        """
+        Compress model log files into tarball.
+        """
+        log_files = self.get_log_files()
+        with tarfile.open(name=os.path.join(self.work_path, self.log_tar_name),
+                          mode="w:gz") as tar:
+            for file in log_files:
+                tar.add(file, arcname=os.path.basename(file))
+                os.remove(file)
+
```
aidanheerdegen, line 385:
```suggestion
        try:
            with tarfile.open(name=os.path.join(self.work_path, self.log_tar_name),
                              mode="w:gz") as tar:
                for file in log_files:
                   tar.add(file, arcname=os.path.basename(file))
        except:
            # catch some exception
        finally:
            for file in log_files: os.remove(file)
```
I think it might be better to separate the tar bit from the deletion. I know you've written tests for it, but it seems at some stage it [wasn't safe to remove the file until the tar object was written](https://stackoverflow.com/a/48292904/4727812). That may no longer be the case, or perhaps I misinterpreted the SO post.

test/models/test_cice.py
```diff
@@ -300,3 +309,82 @@ def test_no_restart_ptr(config, cice_config_files):
...
+    assert set(os.listdir(expt_workdir)) == {model.log_tar_name,
+                                             non_log_file.name}
+
+    # Check all logs present in tarball
+    log_file_names = {log_path.name for
+                      log_path in cice4_log_files}
+
+    with tarfile.open(os.path.join(expt_workdir, model.log_tar_name),
+                      "r") as tar:
+        assert set(tar.getnames()) == log_file_names
```
aidanheerdegen, line 390: It would be nice for the files to have a non-zero size and test that they're written and read back out correctly.

aidanheerdegen (commented):
I think it's fine to keep it in the cice driver to begin with as this is the only model for which it is desperately required. Then we can test it operationally and tweak it if necessary.

I have suggested a couple of changes, but push back if you think they're unnecessry.

test/models/test_cice.py
aidanheerdegen: Definitely worth doing in the future, but I'd hate to hold this up any further.

payu/models/cice.py
blimlim: That's a good point. It seemed to be working reasonably but I haven't been able to find any more info on this since it was posted, so it's probably safest to delete the files after the writing is finished. 

I've just implemented this as:
https://github.com/ACCESS-NRI/payu/blob/d4e333473702f9938e73fce43c40855ae7ce08a8/payu/models/cice.py#L380-L387

and opted to just delete the files directly after the tarfile is written, so that nothing will be deleted if any exceptions do come up when writing the tarball. Let me know if this looks ok!

test/models/test_cice.py
blimlim: Good idea, I've updated the tests to add some contents to the fake log files, and added a check that the correct content appears in the tarball:

https://github.com/ACCESS-NRI/payu/blob/d4e333473702f9938e73fce43c40855ae7ce08a8/test/models/test_cice.py#L404-L409

payu/models/cice.py
```diff
@@ -333,6 +342,50 @@ def archive(self, **kwargs):
...
+        in self.logs_to_compress.
+
+        Returns
+        -------
+        log_files: list of paths to model log files.
+        """
+        log_files = []
+        for filename in os.listdir(self.work_path):
+            if any((re.match(pattern, filename)
+                    for pattern in self.logs_to_compress)):
```
aidanheerdegen, line 371: An alternative would be to create a single regex with all your patterns separate by `OR` (`|`) and matching once for each file, e.g. https://stackoverflow.com/a/8888615

aidanheerdegen (changes requested):
A couple of small changes, others are optional.

payu/models/cice.py
blimlim: Good idea, I've added this in d110cae

docs/source/config.rst
```diff
@@ -273,6 +273,21 @@ section for details.
...
+``archiving``
+      On completion of a model run, payu moves model output, restart, and log
+      files from the temporary work area to the experiment archive directory.
+      The following settings control the steps taken during the archive step:
+      ``enable`` (*Default:* ``True``)
+            Flag to enable/disable the archive step. If ``False`` all output, restart,
+            and log files will remain in the work directory, and any collation, post-processing,
+            and syncing will not be run.
+      ``compress_logs`` (*Default:* ``True``)
+            Compress model log files into a tarball. Currently only implemented for CICE4.
```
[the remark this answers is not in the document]
blimlim: Sorry I completely blanked on your earlier suggestion to convert to a dict in`fsops.py`. Good idea though, I've added this in c111d6b and I think it helps to make everything cleaner.

docs/source/config.rst
```diff
@@ -273,6 +273,21 @@ section for details.
       POSIX filesystem.
 
 
+Archiving
+---------
+
+``archiving``
```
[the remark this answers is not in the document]
blimlim: Thanks for catching that! I've fixed this in c111d6b

blimlim:
Thanks @aidanheerdegen for the suggestions! I've added in these changes, let me know if you have any additional suggestions!

aidanheerdegen (approved):
LGTM!

PR #535, opened 2024-10-25
fix cice5 restart pointers

### LINKED ISSUE #534: CICE4 & CICE5 Restart Pointer Files
In CICE4, the restart directory is automatically added to the restart filename

```
#ifdef AusCOM
         write(nu_diag,*) 'XXX: restart_dir = ', restart_dir
         filename = trim(restart_dir)//trim(filename)
         write(nu_diag,*) 'XXX: restart file => ', filename
#endif
```

https://github.com/ACCESS-NRI/cice4/blob/e7549ebd2044690a432cc67c1317c81cb194b750/source/ice_restart.F90#L313

However in CICE5, this is not done:

```
#ifdef AusCOM
         write(nu_diag,*) 'XXX: restart_dir = ', restart_dir
         write(nu_diag,*) 'XXX: org restart file => ', filename
!ars599: 28042015 restart issue
!         filename = trim(restart_dir)//trim(filename)
         filename = trim(filename)
         write(nu_diag,*) 'XXX: restart file => ', filename
#endif
```

https://github.com/ACCESS-NRI/cice5/blob/62448127e0e431e86c56c5b388be70c50a91ea93/source/ice_restart_driver.F90#L576C1-L583C7

When using the access driver, payu assumes the first:

https://github.com/payu-org/payu/blob/74be0366a30b667daa708e3410546f2b87ad9aa9/payu/models/access.py#L50

https://github.com/payu-org/payu/blob/74be0366a30b667daa708e3410546f2b87ad9aa9/payu/models/cice.py#L153-L156

But we need to seperate the handling for CICE4 & CICE5, to allow use of CICE5 with ACCESS-ESM1.x

(We could update the model code to be consistent across versions, but I think its best not to do this for backward compatibility).

anton-seaice:
Closes #534 

For CICE4 with AUSCOM set, restart pointer file contains:

`./iced.xxxx`

for CICE5, restart pointer file contains

`./RESTART/iced.xxxx`

### DIFF
--- a/payu/models/access.py
+++ b/payu/models/access.py
@@ -47,12 +47,14 @@ class Access(Model):
                 model.copy_restarts = True
 
                 model.set_timestep = model.set_access_timestep
-                model.get_ptr_restart_dir = model.get_access_ptr_restart_dir
 
             if model.model_type == 'cice5':
                 model.access_restarts.extend(['u_star.nc', 'sicemass.nc'])
 
             if model.model_type == 'cice':
+                # The ACCESS build of CICE assumes that restart_dir is 'RESTART'
+                model.get_ptr_restart_dir = lambda : '.'
+
                 # Structure of model coupling namelist
                 model.cpl_fname = 'input_ice.nml'
                 model.cpl_group = 'coupling'
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -150,11 +150,6 @@ class Cice(Model):
     def get_ptr_restart_dir(self):
         return os.path.relpath(self.work_init_path, self.work_path)
 
-    def get_access_ptr_restart_dir(self):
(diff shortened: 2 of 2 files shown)


aidanheerdegen (approved):
LGTM and thanks for the informative issue for the context.

blimlim:
Changes look good to me! 

Slightly unrelated question, but would there be any value in splitting off a separate `cice4` driver? I feel like we keep having to make changes for cice4 in the `cice.py` driver, and then have to override them in the `cice5.py` driver, and other times try to fix something for cice5, and then override it for `cice4` in the `access.py` driver. I often get confused jumping between the three different parts trying to work out what's actually going on, so am interested to hear everyone else's opinions!

blimlim (approved):

PR #537, opened 2024-11-12
Staged cable improve resubmission and restart handling

Whyborn:
## Improvements to ```staged_cable``` driver
I made some improvements to the ```staged_cable``` driver to hopefully make it more easily understandable, and more easily shareable. Summary of changes:

* Changed the way older restarts are handled. Previously, when setting up the current work directory, we would also run through older restart directories to retrieve any "unique" restarts (any with file name not matching a current restart). I've changed it so that at the end of a stage, we grab any "unique" restart from the previous restarts and copy it into the current restart work directory. Quick example: we've finished stage 1, which created restarts ```climate_rst.nc``` and ```cable_rst.nc```, which live in ```{archive}/restart000```. We run stage 2, which created restarts ```casa_rst.nc``` and ```cable_nc.nc```. We look back at ```{archive}/restart000```, see that there's no new version of ```climate_rst.nc```, so copy that to the work restart directory. The contents of the work restart directory are then moved to ```{archive}/restart001```. This makes restarts much more easily shareable.
* Changed the way the remaining number of runs are set. Previously, on initialisation of the model, the configuration log was read and the environment variable dictating the number of runs was modified. I realised I can set ```expt.n_runs``` instead of the environment variable, which is persistent. Now the configuration log work is only done where necessary in ```setup()``` and archive```, rather than every initialisation.

Coming with this is the removal of a fair bit of now-dead code.

Edit: One thing I forgot to mention is the handling of potentially new namelists. I would like it to be possible for people adding new science (or perhaps ```cable``` configurations I don't know about yet) to be able to use the tool. Unfortunately, I can't know a priori which namelists will be required in these instances. To get around this, I could simply set

```
self.optional_config_files = glob.glob('*/nml')
```

rather than the predefined list of namelists, but it feels like a bit of an abuse of process. Thoughts?

### DIFF
--- a/payu/models/staged_cable.py
+++ b/payu/models/staged_cable.py
@@ -37,7 +37,8 @@ class StagedCable(Model):
         # but this feels like a bit of an abuse of feature.
         self.config_files = ['stage_config.yaml']
         self.optional_config_files = ['cable.nml', 'cru.nml',
-                                      'luc.nml', 'met_names.nml']
+                                      'luc.nml', 'met_names.nml',
+                                      'bios.nml']
 
     def setup(self):
         super(StagedCable, self).setup()
@@ -60,11 +61,12 @@ class StagedCable(Model):
         mkdir_p(os.path.join(self.work_path, "logs"))
 
         # Get the additional restarts from older restart dirs
-        self._get_further_restarts()
+        # self._get_further_restarts()
 
         # Make necessary adjustments to the configuration log
-        self._handle_configuration_log_setup()
+        # self._handle_configuration_log_setup()
 
+        self._set_current_stage()
     def _build_new_configuration_log(self):
         """Build a new configuration log for the first stage of the run."""
 
@@ -266,20 +268,30 @@ configuration log."""
         os.rmdir(self.work_restart_path)
 
         # Now collect any restarts from prior_restart_path only if said restart
-        # doesn't already exist
-        prior_restarts = os.listdir(self.restart_path)
-        for f in self.prior_restart_path:
-            if f not in prior_restarts:
-                shutil.copy(os.path.join(self.prior_restart_path, f),
-                        self.restart_path)
+        # doesn't already exist (except on first run)
+        if self.expt.counter > 0:
+            # self.prior_restart_path nor expt.prior_restart_path are defined here
+            # build it as in experiment.py
+            prior_restart_dir = 'restart{0:03}'.format(self.expt.counter - 1)
+            prior_restart_path = os.path.join(self.expt.archive_path,
+                                              prior_restart_dir)
+
+            current_restarts = os.listdir(self.restart_path)
+            for f in os.listdir(prior_restart_path):
+                if f not in current_restarts:
+                    shutil.copy(os.path.join(prior_restart_path, f),
+                            self.restart_path)
 
     def _archive_current_stage(self):
         """Move the current stage to the list of completed stages."""
         self.configuration_log['completed_stages'].append(
-                self.configuraton_log['current_stage'])
+                self.configuration_log['current_stage'])
 
         self.configuration_log['current_stage'] = ''
         self._save_configuration_log()
+
+        # Copy the configuration log to the restart directory for shareability
+        shutil.copy('configuration_log.yaml', self.restart_path)
 
     def collate(self):
         pass
--- a/payu/models/staged_cable.py
+++ b/payu/models/staged_cable.py
@@ -21,23 +21,6 @@ from payu.models.model import Model
 from payu.fsops import mkdir_p
 
 
-def deep_update(d_1, d_2):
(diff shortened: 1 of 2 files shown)


aidanheerdegen:
> I wanted to add Claire to the reviewers, but she doesn't appear in the list of reviewers.

I've made a team for ACCESS-NRI and this is rectified

payu/models/staged_cable.py
```diff
@@ -47,30 +30,45 @@ def __init__(self, expt, name, config):
         self.model_type = 'staged_cable'
         self.default_exec = 'cable'
 
+        # We want people to be able to use payu during testing, which
+        # often means additions of new namelists due to new science
+        # modules. I would like to set
+        # optional_config_files = glob.glob("*.nml")
+        # but this feels like a bit of an abuse of feature.
```
aidanheerdegen, line 37: In this case I don't think this would be a problem. It just creates a list of files that will be copied into the ephemeral `work` directory and tracked with `git`.

payu/models/staged_cable.py
```diff
@@ -238,52 +190,85 @@ def _apply_stage_namelists(self, stage_name):
...
+
+            # For each restart, check if newer version was created. If not,
+            # copy into the work restart path.
+            generated_restarts = os.listdir(self.work_restart_path)
+
+            for f in os.listdir(prior_restart_path):
+                if f not in generated_restarts:
+                    shutil.copy(os.path.join(prior_restart_path, f),
+                                self.work_restart_path)
+
```
aidanheerdegen, line 248: I think it should be possible to just make links, which will resolve to their original location IIRC.

aidanheerdegen (commented):
Looks fine to me, but I've not looked at the logic deeply, just had a skim and commented on things that stood out to me.

payu/models/staged_cable.py
Whyborn: What about the scenario where we want to share a restart directory with someone else, or perhaps even publicly somewhere, so that they can use it as a start point for simulations? I think it's likely enough that the original run would be somewhere that others won't have read permissions to, or even potentially on other machines.

ccarouge: I haven't looked into the code changes but I don't see how making a copy solves problems of read permissions? You need read access to be able to copy files. So links would just be as well.

Whyborn: I'm thinking of a situation where someone (e.g. ACCESS-NRI) has run an expensive spin-up configuration and want to provide the restarts as a starting point for others in the community. The configuration may have 20+ stages, but the full suite of restarts may contain a restart from the 1st stage, one from the 5th stage and the rest from the most recent stage. A supported ```payu``` workflow is to take a restart from elsewhere and effectively plug it in, so ```payu``` transparently restarts from this restart (I think this is true- I have memory of this being demonstrated on the training day).

I'm imagining a scenario where I run a configuration to create a starting point for community science. I'll likely save the data to my non-permanent scratch space. I want to take the most recent ```payu``` set of restarts and store it someone accessible on ```gdata```. If the necessary older restarts are only known by symlink, then they'll disappear relatively soon and whichever user takes this ```payu``` restart may not have access to my scratch space where the symlinked files live.

ccarouge: It's because the `archive` space is created under `/scratch` by default and it's the user's responsibility to save it to `/g/data` if required. I'd forgotten about that bit for payu.

So copy makes sense.

ccarouge (dismissed):
It looks ok to me.

I had to go read the whole file to remind myself of what was going on here, so I have a comment on something you haven't changed. Feel free to ignore.

In this block of code:
https://github.com/payu-org/payu/blob/b69b26d331a49b40f7acd10ffd7976b2e3599da9/payu/models/staged_cable.py#L178-L188

I think it would be more readable if you define `master_nml` at the same place as `stage_nml` and use it in the IF condition on file existence (l.182):

```python
if os.path.isfile(master_nml): 
```

payu/models/staged_cable.py
aidanheerdegen: You can use `cp` with the `-L` option to "dereference" links, so the destination will contain all the files referenced by links

https://superuser.com/a/216920

test/models/test_staged_cable.py
```diff
@@ -107,10 +109,10 @@ def teardown_module(module):
...
         print("teardown_module   module:%s" % module.__name__)
 
-    try:
-        shutil.rmtree(tmpdir)
-    except Exception as e:
-        print(e)
+    # try:
+        # shutil.rmtree(tmpdir)
+    # except Exception as e:
+        # print(e)
```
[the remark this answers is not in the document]
Whyborn: Did this to work out why some tests were failing, should be uncommented.

Issue #538, opened 2024-11-28
CICE in ESM1.5 can repeatedly use the same restart file


blimlim:
In the ESM1.5 configurations, we changed the CICE `dumpfreq` parameter from `m` (monthly) to `y` (yearly), reducing the number of restart files produced by CICE.

We'd expect that running ESM1.5 in monthly segments would fail, as there would be no valid CICE restart files for subsequent runs. 
This is not the case – instead CICE repeatedly uses the same restart files. The following shows the restart directories produced by `payu run -n 3` with a 1 month runtime:
```
ls restart00*/ice
restart000/ice:
cice_in.nml  iced.01010101  ice.restart_file  ice.restart_file-01001231  input_ice.nml  mice.nc  mice.nc-01001231  restart_date.nml

restart001/ice:
cice_in.nml  iced.01010101  ice.restart_file  ice.restart_file-01001231  input_ice.nml  mice.nc  mice.nc-01001231  restart_date.nml

restart002/ice:
cice_in.nml  iced.01010101  ice.restart_file  ice.restart_file-01001231  input_ice.nml  mice.nc  mice.nc-01001231  restart_date.nml
```
In each of the above the restart pointer file `ice.restart_file` points to the `./iced.01010101` file. 

The `restart_date.nml` files do increment as they are calculated by payu:
```
cat restart00*/ice/restart_date.nml
&coupling
    inidate = 1010201
    init_date = 10101
/
&coupling
    inidate = 1010301
    init_date = 10101
/
&coupling
    inidate = 1010401
    init_date = 10101
/
```

In the second and third run, no history is written (perhaps due to time mismatches between the restart file and the calendar file).
```
ls output*/ice
output000/ice:
cice_in.nml    HISTORY     ice_diag_out  iceout086  iceout088  iceout090  iceout092  iceout094  iceout096
debug.root.03  ice_diag.d  iceout085     iceout087  iceout089  iceout091  iceout093  iceout095  input_ice.nml

output001/ice:
cice_in.nml    ice_diag.d    iceout085  iceout087  iceout089  iceout091  iceout093  iceout095  input_ice.nml
debug.root.03  ice_diag_out  iceout086  iceout088  iceout090  iceout092  iceout094  iceout096

output002/ice:
cice_in.nml    ice_diag.d    iceout085  iceout087  iceout089  iceout091  iceout093  iceout095  input_ice.nml
[...]

PR #539, opened 2024-12-03
Check CICE4 restart file dates

blimlim:
_Note: I've started this pull request as a draft as it currently has no unit tests. However, I had quite a bit of trouble working out the best way to structure these changes (which drivers to put them in etc), and hence would appreciate any comments/feedback/ideas/suggestions on the structure while I'm working on the unit tests._ 

This draft PR addresses #538, introducing two guards against CICE4 using a restart file whose date clashes with the date in the restart calendar file. This should prevent:
1. Payu from picking up the wrong restart file
2. Users from mistakenly changing the name of an `iced.YYYYMMDD` restart file.

These are both handled by the `cice4_make_restart_pointer()` function. This checks that an `iced.YYYYMMDD` file exists with a name matching the start date calculated from the calendar restart file (`restart_date.nml`). It then checks that the `time` in the `iced` file's header matches the previous runtime calculated from the calendar restart file, before writing the new pointer file.

This function is called during the `setup` stage of the `access` driver. 
```Python
            if model.model_type == 'cice':
                # Set up and check the cice restart files.
                cice4_make_restart_pointer(model,
                                           run_start_date,
                                           previous_runtime)
```

I think it would have made sense to place these steps in the `cice` driver, and include necessary overrides in the `cice5` driver, rather than putting them in the `access` driver as I've done here. The challenge was that for CICE4 in ESM1.5, the `access` driver calculates the start date and run time used by the model, and hence it contains the data needed to correctly identify the `iced.YYYYMMDD` file and check its header. 

Meanwhile the start date and prior runtime calculated in the `cice` driver (based on the `cice_in.nml` files) appear to be mostly (but not completely) unused by the model, are incorrect (i.e. in the PI configuration's `cice.py` calendar has always been 100 years off the actual date used by the model), and use the wrong type of calendar. Because of this, I thought it was simplest to implement the changes in the `access` driver where the available information is correct.

I've included a blank `make_restart_ptr()` method in the `cice` driver, and overridden it in the `cice5` driver so that CICE5 is unaffected by these changes.

I'll get started on some testing, but it would be great to hear if you have any feedback on this overall structure. It's clearly not the cleanest or most intuitive, and so any ideas would be welcome.


payu/models/access.py
```diff
@@ -218,6 +219,12 @@ def setup(self):
                 f90nml.write(cpl_nml, nml_work_path + '~')
                 shutil.move(nml_work_path + '~', nml_work_path)
 
+            if model.model_type == 'cice':
+                # Set up and check the cice restart files.
+                cice4_make_restart_pointer(model,
+                                           run_start_date,
+                                           previous_runtime)
+
```
anton-seaice, line 227: I probably just didn't read the description carefully enough.

Can this be a function in the cice class, and it just's called here?

i.e. move cice4_make_restart_pointer & associated files to the Cice class. and then this line becomes

`model.cice4_make_restart_pointer(run_start_date,
                                           previous_runtime)`

If you do this, a better name might be `overwrite_restart_ptr()` ?

payu/models/cice.py
```diff
@@ -162,29 +162,10 @@ def setup(self):
...
-
-            res_ptr_path = os.path.join(self.work_init_path,
-                                        'ice.restart_file')
-            if os.path.islink(res_ptr_path):
-                # If we've linked in a previous pointer it should be deleted
-                os.remove(res_ptr_path)
-            with open(res_ptr_path, 'w') as res_ptr:
-                res_dir = self.get_ptr_restart_dir()
-                print(os.path.join(res_dir, iced_restart_file), file=res_ptr)
+            self.make_restart_ptr()
```
anton-seaice, line 168:
```suggestion
            self._make_restart_ptr()
```
I think these functions are intended to be private to this class ?

anton-seaice (commented):
Thanks @blimlim - if its feasible to check the value of `dumpfreq` in the cice namelist that would be great. 

I think whatever we do will be unavoidably messy, i made some suggestions, not sure if they are useful or not!

payu/models/cice.py
blimlim: Good point! I've added this in c1b3f57

payu/models/access.py
blimlim: That's a good idea, I think it helps to clean up the structure. I've added this in c1b3f57

payu/models/access.py
```diff
@@ -333,3 +340,65 @@ def set_model_output_paths(self):
...
+               )
+        raise RuntimeError(msg)
+
+
+def read_binary_iced_header(iced_path):
+    """
+    Read header information from a CICE4 binary restart file.
+    """
+    with open(iced_path, 'rb') as iced_file:
+        header = iced_file.read(24)
```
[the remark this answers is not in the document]
blimlim: Yeah good idea! Added in c1b3f57

anton-seaice (commented):
I ran this and checked it makes the expected error, ill finish the review tomorrow :)

anton-seaice:
I was thinking about this yesterday - it leaves the config in a kind of weird state. e.g.

The model runs & completes (the archive step succeeds) however there is no restart saved to restart the model. The user only gets an error once they try to restart and is it clear they need to go back one model run to fix this ? I don't think payu has an easy way of going back one run ? ( I guess deleting the output & restart folder is the way to do that?)

Obviously once the check on the dump dates is done at the start of the run, then the issue goes away. I guess we could move the check in this PR to the end of a run, e.g. the archive step fails if the cice restart file is at the "wrong" time?. For a release, we can put this in the known issues and provide advice on the workaround as needed.

test/models/test_cice.py
```diff
@@ -311,6 +323,172 @@ def test_no_restart_ptr(config, cice_nml, cice_history_nml):
...
+    # Run timing information
+    previous_runtime = 500000
+    run_start_date = datetime.date(500, 12, 3)
+    # Expected iced file
+    run_start_date_int = payu.calendar.date_to_int(run_start_date)
+    iced_name = f"iced.{run_start_date_int:08d}"
+
+    # Create iced restart with wrong date
+    wrong_iced_name = "iced.01010101"
+    wrong_runtime = 1000
```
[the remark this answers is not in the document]
blimlim: Can do! Just to clarify, do you mean the date in the `iced.YYYYMMDD` filename? This test checks that an error is raised if there are no files with the expected date, and so isn't using the `previous_runtime`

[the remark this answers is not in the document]
anton-seaice: So i think that means check 01010102 ? and 01001231 ?

blimlim:
That's a good point!
Since we'll still be writing the restart pointer during setup, I'm wondering if it's worth keeping the check during setup to make sure that what we're writing is correct, and adding an additional one during archive. What would you think of adding a separate small check to the archive stage to see if the expected `iced.YYYYMM` file exists, so that we have:

Setup stage:
- Check `iced` file matching run start date exists, and if so write it to the pointer file
- Check the header to make sure the previous runtime is as expected

Archive stage:
- Check for an `iced` file whose name matches the run end date

And once we add the `dumpfreq` checks we could remove/consolidate the different checks.

anton-seaice:
I am ok with all the checks being there :-). If you have time, then go ahead, if you don't and want to get a release out - then what we have prevents already in this PR prevents the initial error, its just messy to recover from .

test/models/test_access.py
```diff
@@ -229,7 +230,14 @@ def test_access_cice_calendar_cycling_500(
             # which we are trying to bypass.
             shutil.copy(default_input_ice, cice_model.work_path)
 
-            access_model.setup()
+            # Skip writing restart pointer as it requires iced file
+            # with valid header
+            with patch(
```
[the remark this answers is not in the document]
blimlim: Good idea! I've added this in 9135f89

payu/models/cice.py
```diff
@@ -351,3 +331,101 @@ def link_restart(self, fpath):
...
+            Seconds between experiment initialisation date and start date
+        calendar_file:  str
+            Calendar restart file used to calculate timing information
+        """
+        # Expected iced restart file name
+        run_start_date_int = cal.date_to_int(run_start_date)
+        iced_restart_file = f"iced.{run_start_date_int:08d}"
+
+        if iced_restart_file not in self.get_prior_restart_files():
+            msg = (f"Restart file {iced_restart_file} matching "
```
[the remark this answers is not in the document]
blimlim: Yep that's clearer! I've added this in 9135f89

test/models/test_cice.py
```diff
@@ -311,6 +323,172 @@ def test_no_restart_ptr(config, cice_nml, cice_history_nml):
...
+    wrong_iced_path = prior_restart_dir / wrong_iced_name
+    write_iced_header(wrong_iced_path,
+                      bint=0,
+                      istep0=0,
+                      time=wrong_runtime,
+                      time_forc=0)
+
+    # Check test set up correctly
+    assert iced_name != wrong_iced_name
```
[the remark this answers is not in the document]
blimlim: Oh good point! I've added this and fixed a couple of other test errors in 9135f89

test/models/test_cice.py
```diff
@@ -311,6 +323,172 @@ def test_no_restart_ptr(config, cice_nml, cice_history_nml):
...
+
+@pytest.mark.parametrize("config", [DEFAULT_CONFIG],
+                         indirect=["config"])
+@pytest.mark.parametrize(
+     'run_start_date, previous_runtime',
+     [
+        (datetime.datetime(1500, 1, 1), 500000),
+        (datetime.datetime(1, 7, 3), 900000000),
+        (datetime.datetime(3753, 9, 23), 123456789),
+     ]
```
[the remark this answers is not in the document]
blimlim: Can do! I've done first valid date, 1/1/1, with the shortest previous runtime, 1, and the last valid date 9999/12/31, with the longest possible runtime, 315537811200. These are based on the dates that payu would be able to handle - though I'm not sure what limits CICE itself has.

payu/models/access.py
```diff
@@ -300,6 +284,33 @@ def archive(self):
...
+                                        model.restart_path,
+                                        model.ice_nml_fname
+                )
+
+                if os.path.exists(work_ice_nml_path):
+                    shutil.copy2(work_ice_nml_path, restart_ice_nml_path)
+
+                # Check that run produced iced.YYYYMMDD restart file matching
+                # run end date.
+                model.find_matching_iced(model.restart_path, run_end_date)
```
anton-seaice, line 312: Can this be run earlier ? By this point some files have already been archived, I think you can run it using the mode.work_path before the restart files are moved

blimlim: Ah this can only be run once the correct `run_end_date` has been calculated, and so I'm not sure if it can be run earlier

blimlim: I could move it up a little though, just before
```python
# Copy supplemental restart files to RESTART path
                for f_name in model.access_restarts:
                    f_src = os.path.join(model.work_path, f_name)
                    f_dst = os.path.join(model.restart_path, f_name)
```

anton-seaice: What is happening is that half the restarts get moved the the archive folder, and `payu sweep` won't clean them up. They get found when trying to run again

blimlim: Hmm yeah that's not ideal. The top level `access.py` driver's `archive()` is always called after the submodels: https://github.com/payu-org/payu/blob/da047828f8eee8bd7746f0b7cb3df75044cb94b8/payu/experiment.py#L798-L803

and so things will always be partially archived as long as we run the check in the `access` driver, which has access to the correct end date.

Perhaps it's better to just omit this additional test in that case? Let me know what you think

anton-seaice: Could you force the cice submodel part to run first, before any other models archive ?

anton-seaice (dismissed):
I did adhoc tests for :

- Trying to restart with a cice restart with the wrong date - payu raised error
- Two consecutive months with correct restarts - ran ok

anton-seaice:
Thanks @blimlim this is neat piece of work :)

payu/models/access.py
anton-seaice: We've decided to defer this part for now

blimlim:
Awesome, thanks @anton-seaice for your help with this :) I'll go make that issue about the additional tests

I've just made one more tiny commit fixing a typo, and undoing an unnecessary change I'd made earlier.

@aidanheerdegen if you have any time to look over this, that would be much appreciated, apologies for cutting it so close time wise with this!

anton-seaice (approved):

PR #542, opened 2024-12-16
Fix log compression bug when no logs specified

### LINKED ISSUE #541: Allow for no log files to be compressed
The implementation of model log compression in #532 has a bug which causes OM2 runs to fail.

Compression is currently run from the `cice.py` driver, using the [`get_log_files()` and `compress_log_files()`](https://github.com/payu-org/payu/blob/0c2a837bb9fd0b1546de85945413e6989a7131fd/payu/models/cice.py#L326-L353) methods. The `get_log_files_method()` searches the work directory for files matching regex patterns contained in the  `model.logs_to_compress` list.

For example in `cice.py` we set:
https://github.com/payu-org/payu/blob/0c2a837bb9fd0b1546de85945413e6989a7131fd/payu/models/cice.py#L58-L61
which matches all the desired log files for CICE4 in ESM1.5.

While for `cice5.py` we set:
https://github.com/payu-org/payu/blob/0c2a837bb9fd0b1546de85945413e6989a7131fd/payu/models/cice5.py#L45-L46

The intended behaviour is for no files to be compressed when `model.logs_to_compress` is empty.

In the actual implementation, the opposite happens when `model.logs_to_compress` is empty:
https://github.com/payu-org/payu/blob/0c2a837bb9fd0b1546de85945413e6989a7131fd/payu/models/cice.py#L335-L339

The `"|".join(self.logs_to_compress)` statement produces `""`, which then matches _every_ file in the model's work directory.  `model.compress_log_files()` then tries to compress and delete everything.

In OM2, this causes the following error.

```python
  File "/g/data/vk83/prerelease/apps/payu/dev/bin/payu-run", line 8, in <module>
    sys.exit(runscript())
  File "/g/data/vk83/prerelease/apps/payu/dev/lib/python3.10/site-packages/payu/subcommands/run_cmd.py", line 135, in runscript
    expt.archive(force_prune_restarts=run_args.force_prune_restarts)
  File "/g/data/vk83/prerelease/apps/payu/dev/lib/python3.10/site-packages/payu/experiment.py", line 799, in archive
    model.archive()
  File "/g/data/vk83/prerelease/apps/payu/dev/lib/python3.10/site-packages/payu/models/cice.py", line 324, in archive
    self.compress_log_files()
  File "/g/data/vk83/prerelease/apps/payu/dev/lib/python3.10/site-packages/payu/models/cice.py", line 353, in compress_log_files
    os.remove(file)
IsADirectoryError: [Errno 21] Is a directory: '/scratch/tm70/sw6175/access-om2/work/cice-resdate-changes-test-release-025deg_jra55_ryf-bc5b6712/ice/OUTPUT'
```

blimlim:
This PR closes #541.

Regex patterns for finding log files previously broke when no log files were specified, as done for CICE5. Instead of matching no files, the `get_log_files()` method would match every file in the model work directory.

This PR adds checks at the start of the `get_log_files()` and `compress_log_files()` methods, causing them to exit early when no log files are specified, and when no log files are found.

It also adds a unit test `test_log_compression_no_logs()` to `test_cice.py` which makes sure that the compression method does nothing when no log files are specified.

With these changes, OM2 runs without any issues.

### DIFF
--- a/payu/models/cice.py
+++ b/payu/models/cice.py
@@ -332,6 +332,9 @@ class Cice(Model):
         -------
         log_files: list of paths to model log files.
         """
+        if not self.logs_to_compress:
+            return
+
         log_files = []
         for filename in os.listdir(self.work_path):
             if re.match("|".join(self.logs_to_compress), filename):
@@ -343,6 +346,9 @@ class Cice(Model):
         Compress model log files into tarball.
         """
         log_files = self.get_log_files()
+        if not log_files:
+            return
+
         with tarfile.open(name=os.path.join(self.work_path, self.log_tar_name),
                           mode="w:gz") as tar:
             for file in log_files:
--- a/test/models/test_cice.py
+++ b/test/models/test_cice.py
@@ -595,3 +595,37 @@ def test_log_compression(config, cice4_log_files, non_log_file,
             with tar.extractfile(entry) as open_entry:
                 file_contents = open_entry.read().decode("utf-8")
                 assert file_contents == cice4_log_files[entry_name]
+
+
+@pytest.mark.parametrize("config", [CONFIG_WITH_COMPRESSION],
+                         indirect=True)
+def test_log_compression_no_logs(config, cice4_log_files, non_log_file,
+                                 cice_nml   # Required by expt.__init__
+                                 ):
+    """
+    Check that log compression does nothing when no logfiles are
+    specifed.
+    """
+    with cd(ctrldir):
+        # Initialise laboratory and experiment
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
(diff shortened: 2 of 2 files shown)


anton-seaice (approved):
The update version runs OM2 successfully - I didn't test with esm1.5. 

You can set 

```
       self.logs_to_compress = [r"ice.log.task_[0-9]{4}"]
       self.log_tar_name = "ice.logfiles.tar.gz"
```

for cice5 if you'd like, but no particular need to.

blimlim:
Thanks for the review @anton-seaice! I'll leave out the CICE5 log files for now, but we can always add them in later if they do cause issue. I've just done an extra test with ESM1.5 and everything's working there

PR #550, opened 2025-01-20
Fix job.yaml qstat call

### LINKED ISSUE #545: `job.yaml` doesn't include any pbs specific info
There's a `job.yaml` that gets written out during `payu run` job once the model has been run that outputs information about the payu job, e.g.
```
PAYU_CONTROL_DIR: /home/$USERID/test-payu/mom6-double-grye-base
PAYU_CURRENT_RUN: 0
PAYU_FINISH_TIME: '2024-12-20T08:25:39.134174'
PAYU_JOB_STATUS: 0
PAYU_N_RUNS: 1
PAYU_PATH: /g/data/vk83/prerelease/apps/base_conda/envs/payu-1.1.6/bin
PAYU_RUN_ID: b35e7409aed05466e0d8215727df3b107f05be85
PAYU_START_TIME: '2024-12-20T08:25:34.287522'
PAYU_WALLTIME: 4.846652 s
```

Currently it is being generated from within the `Experiment.run()` method:
https://github.com/payu-org/payu/blob/20a8e76fc249e6feac07598673a4b3dd222be788/payu/experiment.py#L675-L701

 `get_job_info()` uses `qstat` to obtain the job information. This is wrapped in a tenancity retry block that retries the qstat command for 10s. However it's been failing as the 'PBS_EXEC' environment variable does not exist during the jobs, so no PBS job info is being logged out.

https://github.com/payu-org/payu/blob/20a8e76fc249e6feac07598673a4b3dd222be788/payu/schedulers/pbs.py#L206-L207

Changing the above line to `cmd = f"qstat {qflag}"` seems to work OK.

There's also an issue with PBS scheduler specific functions getting imported in the main experiment class - which may not work once running with other schedulers:
https://github.com/payu-org/payu/blob/20a8e76fc249e6feac07598673a4b3dd222be788/payu/experiment.py#L34

jo-basevi:
Closes #545 

This fixes the PBS scheduler `get_qstat_info()` method which was always failing due to `PBS_EXEC` no longer being defined in the environment. Because it is wrapped in a  `tenancity` retry, it would retry the command for 10s. I've changed  to call `qstat` command directly as this seems available on the `$PATH` in PBS jobs. This now results in PBS specific job information being added to the `job.yaml` in the output archive directory.

I've also done a small change to move `get_job_info()` and `get_job_id()` to scheduler class methods - that are extended in the PBS scheduler subclass. This to avoid importing `scheduler.pbs` in `experiment.py`.

Tested that the `job.yaml` populated with PBS run info when running a small [MOM6 config](https://github.com/aidanheerdegen/mom6_double_gyre). Also tested it didn't add PBS specific info when running on the model on the login node using `payu-run`.

### DIFF
--- a/payu/schedulers/scheduler.py
+++ b/payu/schedulers/scheduler.py
@@ -8,6 +8,9 @@
 # expanded to provide greater functionality in the future.
 
 
+from typing import Any, Dict, Optional
+
+
 class Scheduler(object):
     """Abstract scheduler class."""
 
@@ -17,3 +20,28 @@ class Scheduler(object):
 
     def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
         raise NotImplementedError
+
+    def get_job_info(self) -> Optional[Dict[str, Any]]:
+        """Get information about the currently running job
+
+        Returns
+        ----------
+        Optional[Dict[str, Any]]
+            Dictionary of information queried from the scheduler
+        """
+        pass
+
+    def get_job_id(self, short: bool = True) -> Optional[str]:
+        """Get scheduler-specific job ID
+
+        Parameters
+        ----------
+        short: bool, default True
+            Return shortened form of the job ID
+
+        Returns
+        ----------
+        Optional[str]
+            Job id if defined, None otherwise
+        """
+        pass
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -21,7 +21,7 @@ import payu
 import payu.envmod as envmod
 from payu.fsops import is_conda
 from payu.models import index as supported_models
-from payu.schedulers import index as scheduler_index
+from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 import payu.subcommands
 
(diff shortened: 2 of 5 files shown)


payu/schedulers/scheduler.py
```diff
@@ -17,3 +20,28 @@ def __init__(self):
...
+
+    def get_job_info(self) -> Optional[Dict[str, Any]]:
+        """Get information about the currently running job
+
+        Returns
+        ----------
+        Optional[Dict[str, Any]]
+            Dictionary of information queried from the scheduler
+        """
+        pass
```
CodeGat, line 32: Is this like defining an abstract method for the subclass to implement?

CodeGat (approved):
Just the one comment, wouldn't hold up approval!

payu/schedulers/scheduler.py
CodeGat: Ah, I see!

tmcadam (approved):
Looks good

PR #553, opened 2025-01-28
Add UM date-based restart pruning

### LINKED ISSUE #552: Add date-based restart pruning for the UM
ESM1.5/6 uses restart dates from the MOM submodel for the date-based restart pruning: https://github.com/payu-org/payu/blob/7ecb96208f6da32acfb20f83e1ce0938a57a9e0b/payu/models/access.py#L321-L327

We'll (hopefully) shortly be adding AMIP configurations of ESM1.5 and ESM1.6, which have no ocean or sea ice components, running only using the UM atmosphere. These will be based on the [COE configuration](https://github.com/coecms/access-esm-amip), and prototypes appear to run fine through payu. Enabling date-based restart pruning fails though as there is no MOM submodel.

Payu runs of ESM1.5 produce a UM calendar restart file:
```um.res.yaml
end_date: 1978-02-01 00:00:00
```

We could add a `get_restart_datetime` method to the UM driver which parses this file to support date-based restart pruning in the AMIP configurations.

blimlim:
Closes #552 
The AMIP configurations of ESM1.5 and 1.6 only use the UM atmosphere submodel, without the MOM ocean or CICE sea-ice components. 

This PR adds a `get_restart_datetime()` method to the UM driver, allowing for date-based restart pruning via the UM model. The implementation uses the `um.res.yaml` calendar restart file produced by the driver at each run's archive step.

I've set the top-level `access.py` model to use  the MOM driver's `get_restart_datetime()` method by default, and to fall back to the UM driver's method when the MOM submodel is missing.

Any ideas or suggested feedback is very welcome! I'll add some specific comments outlining reasons behind some of the changes.


payu/calendar.py
```diff
@@ -127,25 +127,40 @@ def seconds_between_dates(start_date, end_date, caltype_int):
...
+        # Assume date refers to start of day
+        hour = 0
+        minute = 0
+        second = 0
+    else:
+        raise TypeError(
+            "Expected datetime.date or datetime.datetime. "
+            f"Recieved {type(date)}"
+        )
+
```
blimlim, line 156: The UM reads the restart date as a `datetime.datetime` object instead of a `datetime.date`. I've expanded this to handle this extra case.

payu/models/mom.py
```diff
@@ -225,19 +225,3 @@ def create_mask_table(self, input_nml):
...
-    for sub_model in model.expt.models:
-        if sub_model.model_type == 'mom' :
-            mom_restart_path = os.path.join(restart_path, sub_model.name)
-            return sub_model.get_restart_datetime(mom_restart_path)
-
-    raise NotImplementedError(
-        f'Cannot find mom sub-model: {model.model_type} date-based '
-        'restart pruning requires the mom sub-model to '
-        'determine restart dates'
-    )
```
blimlim, line 243: I'd added this function a while ago when we first added date-based restart pruning for ESM1.5. It was added as OM2 and ESM1.5 both handled the restart-pruning in the exact same way. I've removed it now that this is no longer the case.

We could add a generic `get_restart_datetime_using_submodel` function which takes in an argument for which model to use, though I'm unsure whether this would better. Let me know what you think!

payu/models/um.py
```diff
@@ -203,6 +200,58 @@ def setup(self):
 
         work_nml.write(work_nml_path, force=True)
 
+    def read_calendar_file(self, restart_calendar_path):
```
blimlim, line 203: I've added this extra method so that it can be used for the date-based restart pruning and also during the model setup where it reads the start date

payu/models/mom.py
jo-basevi: A generic `get_restart_datetime_using_submodel` could work? It could be added the base Model class or file, and an input could be a list of model types in priority order to use their get_restart_datetime methods. Something similar to:

```python
def get_restart_datetime_using_submodel(self, restart_path, model_types):
    for model_type in model_types:
        for model in self.expt.models:
            if model.model_type == model_type:
                model_restart_path = os.path.join(restart_path, model.name)
                return model.get_restart_datetime(model_restart_path)

    raise NotImplementedError(
        f'Cannot find any of the specified sub-models: {", ".join(model_types)}. '
        f'{self.model_type} date-based restart pruning requires one of these sub-models '
        'to determine restart dates.'
    )
```

But at the end of the day, I'm not sure on many lines of code this will reduce so I think having separate functions for OM2 and ESM1.5 is also OK

test/models/test_access.py
```diff
@@ -370,3 +373,82 @@ def test_access_cice_1year_runtimes(
...
+    write_config(config)
+
+    # Run test
+    yield
+
+    # Teardown
+    os.remove(config_path)
+
+
+def test_access_get_mom_restart_datetime(mom_only_config, remove_restart_dirs):
```
jo-basevi, line 401: It might be worth testing that the `mom` submodel method is used in priority over the `um` - so have a config with both `mom` and `um` submodels

payu/calendar.py
```diff
@@ -127,25 +127,40 @@ def seconds_between_dates(start_date, end_date, caltype_int):
...
+        second = date.second
+    elif isinstance(date, datetime.date):
+        # Assume date refers to start of day
+        hour = 0
+        minute = 0
+        second = 0
+    else:
+        raise TypeError(
+            "Expected datetime.date or datetime.datetime. "
+            f"Recieved {type(date)}"
```
jo-basevi, line 154:
```suggestion
            f"Received {type(date)}"
```

jo-basevi (dismissed):
Thanks @blimlim, it all looks good! Just a couple small comments

payu/calendar.py
blimlim: Good catch, thanks!

test/models/test_access.py
blimlim: Good idea! I've absorbed this into the above test in 590b17c

blimlim:
Thanks @jo-basevi for the suggestions. All good ideas and I've added them in!

blimlim:
I'll just fix those tests...

jo-basevi (approved):
Added changes look good - Thanks @blimlim!

blimlim:
Thanks @jo-basevi! merging now

PR #556, opened 2025-02-06
Remove setting restart time in access-om3 driver

### LINKED ISSUE #555: payu setting restart time in access-om3 driver
In the CesmCmeps / AccessOM3 driver, [payu current overrides any user setting of the restart output frequency to the length of the run.](https://github.com/payu-org/payu/blob/083a22b25d5940d9b6ec4ddb51708b405f0d03ed/payu/models/cesm_cmeps.py#L166-L167)

We should remove this and let the user configure this option. This lets the user write extra restarts if desired. Also new builds have a run time option to [set the restart output interval to the length of the run](https://github.com/ACCESS-NRI/access-om3-configs/pull/161#discussion_r1943815662), which provides a good default value for configurations.

I also think our general approach is to set the restart interval in the configuration instead of through payu ?

anton-seaice:
Closes #555 

This lets the user set the restart time as they desire!

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -159,13 +159,7 @@ class CesmCmeps(Model):
 
         # run checks on nuopc.runfig
         self._setup_checks()
-
-        # Ensure that restarts will be written at the end of each run
-        stop_n = self.runconfig.get("CLOCK_attributes", "stop_n")
-        stop_option = self.runconfig.get("CLOCK_attributes", "stop_option")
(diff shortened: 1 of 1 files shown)


dougiesquire (approved):
LGTM, thanks @anton-seaice

dougiesquire:
Oh, let me fix the PyPA build issue then we can rebase this

jo-basevi:
> Oh, let me fix the PyPA build issue then we can rebase this

Oops sorry I should've updated these PRs but it's hopefully already fixed in https://github.com/payu-org/payu/pull/560 that was merged this morning

PR #560, opened 2025-02-10
Update CI

### LINKED ISSUE #559: CI PyPI upload artefact error
PyPI is using a deprecated version of upload artefact which is causing CI to automatically fail (see https://github.com/payu-org/payu/actions/runs/13233350950/job/36933981604?pr=558). I don't think it requires an upload artefact step anyway so might be best to remove it all together

jo-basevi:
- Removed upload artefact step in Build-PyPA job as its not used anywhere
- Try installing test dependencies using optional test dependencies in `pyproject.toml` and removing `test/requirements_test.txt`

Closes #559

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,10 +31,6 @@ jobs:
       - run: |
           python3 -m pip install --upgrade build && python3 -m build
 
-      - uses: actions/upload-artifact@v3
-        with:
-          path: ./dist
-
   conda-build:
     name: Conda Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


CodeGat (approved):
Looks good! Why was the CI failing initially?

jo-basevi:
Yeah so the it was failing because it was using `actions/upload-artifact@v3` action which has now been deprecated and since January 30th (see [deprecation notice](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)), it can no longer be used - so the job automatically failed. The `CD` workflow uses the `v4` versions, I must have accidentally missed updating the `CI` workflow...

Issue #509, opened 2024-09-13
Multiple top-level model instances


jo-basevi:
The `Experiment` in `experiment.py` stores two separate instances of the top-level model driver in `Experiment.model` and in the list `Experiment.models`. For model drivers such as `access`, setup and archive is not run for the `access` instance in `Experiment.models` list, only for the top-level instance `Experiment.model`.

There was already a TODO to rename models to submodels to make it more clear:
https://github.com/payu-org/payu/blob/bdf0c66f143d64b2ecf06289d5cb93e994822f68/payu/experiment.py#L150-L151

Could still have `self.models = [self.model] + self.submodels` for when wanting to iterate over all model drivers, but have only one instance of the top-level model.

What might need to be checked is `accessom2` driver archive method. For the `Experiment.models` instance which archives before other submodel model drivers run `archive()`, it copies directories from work directories restart path to the output restart path. Then when the top level `Experiment.model` instance run after other submodel archives, it copies a file from a `mom` work path to a `cice5` restart path.

So maybe there just needs to be a top-level `post_setup`/`post_archive` method or similar, that runs after other submodel setup/archives methods have run? This might make it more clear what runs before submodel methods and what runs after submodel methods.

anton-seaice:
I think payu needs a design document to explain this stuff 🥲

aidanheerdegen:
A refactor of some of this would be good I think. Would need to spend time thinking about it to have any useful input.

aidanheerdegen:
Related issue: #632

PR #562, opened 2025-02-10
Add restart pruning to OM3 driver

### LINKED ISSUE #506: Support for date-based restart pruning in OM3
We should add support for date based restart pruning when running ACCESS-OM3

@jo-basevi says this: 

_To setup date-based pruning for OM3, just need a method to parse out a date from the restart files  - other model drivers use the ocean_solo.res files from the mom model._

Chances are we can do the same thing (and maybe a shared function with the MOM6 driver makes sense).

There's a bit of refactoring and test addition going on in #496, which may provide some relevant changes to also implementing this.

anton-seaice:
Closes #506 

This adds support for date based restart pruning to the access-om3 driver.

This uses the payu framework for restart pruning, the only addition is a `get_restart_datetime()`
 function with the a access-om3 class.

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -217,9 +217,9 @@ configuration.
    For example, ``restart_freq: 10YS`` would save earliest restart of the year,
    10 years from the last permanently archived restart's datetime.
 
-   Please note that currently, only ACCESS-OM2, MOM5 and MOM6 models support
-   date-based restart frequency, as it depends on the payu model driver being
-   able to parse restarts files for a datetime.
+   Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
+   and MOM6 models support  date-based restart frequency, as it depends on the payu 
+   model driver being able to parse restarts files for a datetime.
 
 ``restart_history``
     Specifies how many of the most recent restart files to retain regardless of 
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -218,7 +218,7 @@ configuration.
    10 years from the last permanently archived restart's datetime.
 
    Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
-   , MOM6 and UM7 models support  date-based restart frequency, as it depends on the payu 
+   , MOM6 and UM7 models support date-based restart frequency, as it depends on the payu 
    model driver being able to parse restarts files for a datetime.
 
 ``restart_history``
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -218,7 +218,7 @@ configuration.
    10 years from the last permanently archived restart's datetime.
 
    Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
-   and MOM6 models support  date-based restart frequency, as it depends on the payu 
+   , MOM6 and UM7 models support  date-based restart frequency, as it depends on the payu 
    model driver being able to parse restarts files for a datetime.
 
 ``restart_history``
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -362,6 +362,7 @@ class CesmCmeps(Model):
         seconds = int(seconds)
         hour = seconds // 3600 ; min = (seconds % 3600) // 60 ; sec = seconds % 60
 
+        # TODO: change to self.control_path if https://github.com/payu-org/payu/issues/509 is implemented
         self.get_runconfig(self.expt.control_path)
         run_calendar = self.runconfig.get("CLOCK_attributes", "calendar")
 
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -24,6 +24,12 @@ from payu.models.mom6 import mom6_add_parameter_files
 NUOPC_CONFIG = "nuopc.runconfig"
 NUOPC_RUNSEQ = "nuopc.runseq"
 
+#mapping of runconfig to cftime calendars:
+CFTIME_CALENDARS = {
+    "NO_LEAP" : "noleap" ,
+    "GREGORIAN" : "proleptic_gregorian"
+}
+
 # Add as needed
 component_info = {
     "mom": {
@@ -339,7 +345,8 @@ class CesmCmeps(Model):
 
     def get_restart_datetime(self, restart_path):
         """Given a restart path, parse the restart files and
-        return a cftime datetime (for date-based restart pruning)"""
+        return a cftime datetime (for date-based restart pruning)
+        Supports noleap and proleptic gregorian calendars"""
 
         # Check for rpointer.cpl file
         rpointer_path = os.path.join(restart_path, 'rpointer.cpl')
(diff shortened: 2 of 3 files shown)


anton-seaice:
Sorry yet another review :-) @dougiesquire and @jo-basevi

payu/models/cesm_cmeps.py
```diff
@@ -336,6 +343,46 @@ def collate(self):
...
+        run_calendar = self.runconfig.get("CLOCK_attributes", "calendar")
+
+        try:
+            cf_cal = CFTIME_CALENDARS[run_calendar]
+        except KeyError:
+            warn(
+                f"Unsupported calendar for restart pruning: {run_calendar},"
+                f" try {' or '.join(CFTIME_CALENDARS.keys())}"
+            )
+            return False
```
jo-basevi, line 372: It might be best to raise an exception here as the caller method in `Experiment.py` will eventually raise a TypeError with bool not being a cftime. Any exceptions raised during the determining restarts to prune will be caught during the archive method, but it might make the error trace more clear

jo-basevi (dismissed):
Changes look good to me! Just one comment really on the return type when there's an unsupported calendar

docs/source/config.rst
```diff
@@ -217,9 +217,9 @@ configuration.
    For example, ``restart_freq: 10YS`` would save earliest restart of the year,
    10 years from the last permanently archived restart's datetime.
 
-   Please note that currently, only ACCESS-OM2, MOM5 and MOM6 models support
-   date-based restart frequency, as it depends on the payu model driver being
-   able to parse restarts files for a datetime.
+   Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
+   and MOM6 models support  date-based restart frequency, as it depends on the payu
```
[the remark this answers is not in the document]
anton-seaice: Is UM only a real configuration ?

Is it ok to put ESM1.6 in? Its a quite developmenty still

payu/models/cesm_cmeps.py
anton-seaice: Great thanks, will do

payu/models/cesm_cmeps.py
```diff
@@ -330,6 +337,46 @@ def collate(self):
...
+            lines = ocean_solo.readlines()
+            # example lines would be access-om3.cpl.r.1900-01-02-00000.nc
+
+        date_str = lines[0].split('.')[3]
+        year, month, day, seconds = date_str.split('-')
+
+        seconds = int(seconds)
+        hour = seconds // 3600 ; min = (seconds % 3600) // 60 ; sec = seconds % 60
+
+        self.get_runconfig(self.expt.control_path)
```
anton-seaice, line 362: @jo-basevi - is it correct this uses `self.expt.control_path` and not `self.control_path` ?

payu/models/cesm_cmeps.py
```diff
@@ -23,6 +24,12 @@
 NUOPC_CONFIG = "nuopc.runconfig"
 NUOPC_RUNSEQ = "nuopc.runseq"
 
+#mapping of runconfig to cftime calendars:
```
anton-seaice, line 27:
```suggestion
# mapping of runconfig to cftime calendars:
# these match the supported calendars in CMEPS
# https://github.com/ESCOMP/CMEPS/blob/bc29792d76c16911046dbbfcfc7f4c3ae89a6f00/cesm/driver/ensemble_driver.F90#L196
```

docs/source/config.rst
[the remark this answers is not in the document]
jo-basevi: Yeah good point - maybe leave ESM1.6 for now.. I don't know whether a configuration could run UM only - @blimlim said it might be possible (?)

payu/models/cesm_cmeps.py
jo-basevi: I think probably safest to use the model driver's `self.control_path` - like I don't think it will be different from the experiment control path as `cesm_cmeps` is not a sub-model component but it probably doesn't hurt just incase the model's control path does get modified.

docs/source/config.rst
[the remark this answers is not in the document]
blimlim: We have some UM-only AMIP [configurations](https://github.com/ACCESS-NRI/access-esm1.5-configs/blob/amip/config.yaml) in development, but these still place the UM as a sub-model of ESM1.5/6. Perhaps it makes sense for it to come under ESM1.5 here.

Would it be similar for MOM5 vs OM2, and MOM6 vs OM3?

It's possible to also run the same AMIP configs with the UM as the 'top level model', which was done in the [clex configuration](https://github.com/coecms/access-esm-amip/blob/main/config.yaml), but I don't think we're planning to release similar configurations.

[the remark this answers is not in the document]
anton-seaice: MOM6 standalone is a thing that folks use for regional work and is a different executable than OM3, not sure about MOM5 ?

Anyway ill add the UM, do we just call it the UM or UM7 ?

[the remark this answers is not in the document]
blimlim: I think UM7 is good!

payu/models/cesm_cmeps.py
anton-seaice: I tried this - I get a 

`AttributeError: 'AccessOm3' object has no attribute 'control_path'`

error ?

docs/source/config.rst
anton-seaice, line 221:
```suggestion
   Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
   , MOM6 and UM7 models support  date-based restart frequency, as it depends on the payu 
```

payu/models/cesm_cmeps.py
jo-basevi: Aaah oh no - I think this is an issue with multiple top-level model instances (see https://github.com/payu-org/payu/issues/509).. so the code that sets up `model.control_path` is only run for `expt.models` not the `expt.model` instance. So in the test code if you used `model = expt.models[0]` to get the model from the experiment - it would be the one with control_path defined.

anton-seaice: Im a bit lost but this sounds like no change ?

The normal configuration fails too with `self.control_path` not just the tests

jo-basevi: Aaah yeah -  the restart pruning in `Experiment` class uses `expt.model`.. Ok I think best to use `self.expt.control_path` then so you don't have to worry about modifying the `Experiment` class in this PR. Then once I've gotten around to fixing the above issue, then using the model's control path should work in future

anton-seaice: Ok - do we need to make a note to change this ?

jo-basevi: Yeah I think a TODO/note wouldn't hurt

payu/models/cesm_cmeps.py
```diff
@@ -330,6 +339,47 @@ def collate(self):
...
+            # example lines would be access-om3.cpl.r.1900-01-02-00000.nc
+
+        date_str = lines[0].split('.')[3]
+        year, month, day, seconds = date_str.split('-')
+
+        # convert seconds into hours, mins, seconds
+        seconds = int(seconds)
+        hour = seconds // 3600 ; min = (seconds % 3600) // 60 ; sec = seconds % 60
+
+        self.get_runconfig(self.expt.control_path)
```
anton-seaice, line 365:
```suggestion
        # TODO: change to self.control_path if https://github.com/payu-org/payu/issues/509 is implemented
        self.get_runconfig(self.expt.control_path)
```

jo-basevi (approved):
Thanks @anton-seaice - it looks good to me

docs/source/config.rst
```diff
@@ -217,9 +217,9 @@ configuration.
    For example, ``restart_freq: 10YS`` would save earliest restart of the year,
    10 years from the last permanently archived restart's datetime.
 
-   Please note that currently, only ACCESS-OM2, MOM5 and MOM6 models support
-   date-based restart frequency, as it depends on the payu model driver being
-   able to parse restarts files for a datetime.
+   Please note that currently, only ACCESS-ESM1.5, ACCESS-OM2, ACCESS-OM3, MOM5 
+   , MOM6 and UM7 models support  date-based restart frequency, as it depends on the payu
```
dougiesquire, line 221:
```suggestion
   , MOM6 and UM7 models support date-based restart frequency, as it depends on the payu 
```

dougiesquire (dismissed):
LGTM - thanks @anton-seaice! I only flagged the additional space below to prove I did look at this 🙂

jo-basevi (approved):

Issue #554, opened 2025-02-04
Drivers for ESM1.6


blimlim:
ESM1.6 will need some changes in the model drivers compared to ESM1.5: two new cable namelists will be added to the configurations, and the `um.py` driver will need to pick up as configuration files.

I'm hoping to get ideas/thoughts on the best way of implementing this. One challenge is that we are currently using the same drivers for ESM1.5 and 1.6, which may get more difficult depending on how different the models become.

 A couple ideas I had were:

- Add the new namelists as optional_config_files in the [um driver](https://github.com/payu-org/payu/blob/083a22b25d5940d9b6ec4ddb51708b405f0d03ed/payu/models/um.py#L46). This would "work" in that ESM1.5 and 1.6 would both run provided you used the correct config files for each one, but I don't think this is a good option as I the files aren't really optional.

- Add a new `UnifiedModel1p6` model as a child of the UnifiedModel class which includes the extra configuration files. This would be fairly simple.

- Add a new `access_esm1p6` top-level driver, which can update the `config_files` in a similar way to the existing [access.py driver](https://github.com/payu-org/payu/blob/083a22b25d5940d9b6ec4ddb51708b405f0d03ed/payu/models/access.py#L38-L39). This might also let us simplify the access.py driver by separating the cice4 vs cice5 sections.

Let me know if you have any suggestions or other ideas!

aidanheerdegen:
I'm happy with option 3. It also means the driver for ESM1.6 can evolve more freely, as it doesn't have to maintain backwards compatibility with ESM1.5.

The downside would be some extra/duplicated work if there was some feature we want to add to both drivers.

blimlim:
That sounds good! I'm thinking of having a first PR which duplicates the current `access` driver which we know works as is, and then additional PRs which cut out extraneous parts and add on esm1.6 specific parts, in order to have clearer diffs for the review.

anton-seaice:
> 3\. This might also let us simplify the access.py driver by separating the cice4 vs cice5 sections.

I think well want the option for either for a while to test the various development permutations. e.g. the [spack package](https://github.com/ACCESS-NRI/spack-packages/blob/0330062f0a23306eec0a202ae1ec52f17b24061e/packages/access-esm1p6/package.py#L37) currently allows for both options.

PR #563, opened 2025-02-11
Add initial driver for ACCESS-ESM1.6

blimlim:
Closes the first part of #554

This PR adds an initial driver for ESM 1.6. This first version duplicates the ESM1.5 driver which until now has been used for ESM1.6. ESM1.6 specific changes (e.g. adding new configuration files, removing `matm` support) will be added in separate PRs in order to have clearer diffs. 

The most contentious part could be the file and model names – I tried to follow what I'd seen elsewhere but let me know if there are different conventions that would be better.
- I chose the filename `access_esm1p6.py` based on `accessom2.py`, however added an underscore for clarity since they are used in other drivers like `cesm_cmeps.py` include them.
- For naming the model class, I used `AccessEsm1p6`, following other classes like `AccessOm2`.
- For the `model-type` I used `access-esm1.6`, following `access-om2` in the OM2 driver, and also matching what is used in the spack packages.

### DIFF
--- a/payu/models/__init__.py
+++ b/payu/models/__init__.py
@@ -1,4 +1,5 @@
 from payu.models.access import Access
+from payu.models.access_esm1p6 import AccessEsm1p6
 from payu.models.accessom2 import AccessOm2
 from payu.models.cesm_cmeps import AccessOm3
 from payu.models.cable import Cable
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM. I like your naming choices. Consistent with previous practice, and where it differs it improves clarity.

blimlim:
Thanks @aidanheerdegen! Merged

PR #567, opened 2025-02-13
Commit MOM_parameter_doc.* into docs folder of config

### LINKED ISSUE #565: Commit _MOM_parameter_docs_ in access-om3 driver
As noted in https://github.com/COSIMA/access-om3/issues/117

We would like to capture some MOM6 output, which are four output files from MOM6 which capture the configuration parameters used for a model run. See [this commit](https://github.com/ACCESS-NRI/access-om3-configs/commit/32d084c0c5019ae0d958be4d1a85c464b71a7a6a)

During payu archive, the four files would be copied from the work directory into the the docs folder of the config directory. 

I don't know if its best that payu commits these files automatically (if `runlog:True`) or we expect the user to commit them.

We can follow up in the repro test in https://github.com/ACCESS-NRI/model-config-tests/ with a test that these four files output during the repro test are the same as the ones committed to the branch under test (to ensure they are correct in released configs).

anton-seaice:
Closes #565 

When using the access-om3 driver, the commits the MOM output documentation (`MOM_parameter_doc.*`) into the `docs` folder of the config being run.

If `runlog: False` it adds the docs folder it without committing

If `runlog: True` it commits it, with the message "payu archive: documentation of MOM6 run-time configuration" 

@dougiesquire - do you have a MOM6 standalone config i can test this with ?

@chrisb13 - if you'd like, we could try and do an "online/in zoom" review for training purposes. We probably need @jo-basevi because i'm a hack

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -33,7 +33,7 @@ component_info = {
         "optional_config_files": [
             "field_table",
             "data_table",
-        ]
+        ],
     },
     "cice": {
         "config_files": ["ice_in"],
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -18,7 +18,7 @@ from warnings import warn
 from payu.fsops import mkdir_p, make_symlink
 from payu.models.model import Model
 from payu.models.fms import fms_collate
-from payu.models.mom6 import mom6_add_parameter_files
+from payu.models.mom6 import mom6_add_parameter_files, mom6_save_docs_files
 
 NUOPC_CONFIG = "nuopc.runconfig"
 NUOPC_RUNSEQ = "nuopc.runseq"
@@ -33,7 +33,7 @@ component_info = {
         "optional_config_files": [
             "field_table",
             "data_table",
-        ],
+        ]
     },
     "cice": {
         "config_files": ["ice_in"],
@@ -263,6 +263,11 @@ class CesmCmeps(Model):
         return True
 
     def archive(self):
+
+        # Move any the MOM_parameter_docs output back into the control repo and commit it for documentation
+        if 'mom' in self.components.values() :
+            mom6_save_docs_files(self)
+
         super().archive()
 
         mkdir_p(self.restart_path)
@@ -336,14 +341,21 @@ class AccessOm3(CesmCmeps):
     def get_components(self):
         super().get_components()
 
-        assert self.components["atm"] == "datm", (
-            "Access-OM3 comprises a data atmosphere model, but the atmospheric model in nuopc.runconfig is set "
-            f"to {self.components['atm']}."
-        )
-        assert self.components["rof"] == "drof", (
-            "Access-OM3 comprises a data runoff model, but the runoff model in nuopc.runconfig is set "
-            f"to {self.components['rof']}."
-        )
+        if (self.components["atm"] != "datm") :
+            raise RuntimeError(
+                "Access-OM3 comprises a data atmosphere model, but the atmospheric model in nuopc.runconfig is set "
+                f"to {self.components['atm']}."
+            )
+        if (self.components["rof"] != "drof") :
+            raise RuntimeError(
+                "Access-OM3 comprises a data runoff model, but the runoff model in nuopc.runconfig is set "
+                f"to {self.components['rof']}."
+            )
+        
+    def archive(self):
+        
+
+        super().archive()
 
 
 class Runconfig:
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -402,11 +402,6 @@ class AccessOm3(CesmCmeps):
                 "Access-OM3 comprises a data runoff model, but the runoff model in nuopc.runconfig is set "
                 f"to {self.components['rof']}."
             )
-        
-    def archive(self):
-        
-
-        super().archive()
 
 
 class Runconfig:
--- a/payu/models/mom6.py
+++ b/payu/models/mom6.py
@@ -59,7 +59,7 @@ def mom6_save_docs_files(model):
         except Exception as e:
             warn(e)
 
-    if model.expt.config.get('runlog', True): #if runlog true, default to true
+    if model.expt.runlog.enabled: #if runlog true, default to true
         # commit new files to the control dir
         repo = GitRepository(repo_path = model.control_path)
 
--- a/payu/models/mom6.py
+++ b/payu/models/mom6.py
@@ -23,7 +23,7 @@ from payu.models.fms import Fms
 from payu.models.mom_mixin import MomMixin
 from payu.git_utils import GitRepository
 
-MOM6_DOCS = "MOM_parameter_doc.*"
+MOM6_DOCS = ["MOM_parameter_doc.*","available_diags.*"]
 
 def mom6_add_parameter_files(model):
     """Add parameter files defined in input.nml to model configuration files.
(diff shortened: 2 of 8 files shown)


payu/models/mom6.py
```diff
@@ -41,6 +47,28 @@ def mom6_add_parameter_files(model):
         else:
             model.config_files.extend(filenames)
 
+def mom6_save_docs_files(model):
+    """Add docs files created as MOM output back to the configuration repository"""
```
dougiesquire, line 51:
```suggestion
    """Add docs files created as MOM output back to the control directory"""
```

dougiesquire (commented):
Look great @anton-seaice. I've one question for now and a few suggestions to take or leave.

I haven't actually tested this yet, but will now.

payu/models/mom6.py
```diff
@@ -41,6 +47,28 @@ def mom6_add_parameter_files(model):
...
+    docs_folder = os.path.join(model.control_path, 'docs')
+    mkdir_p(docs_folder)
+
+    # copy everything that matches MOM_parameter_doc.* to the control dir
+    for f in glob(os.path.join(model.work_path,MOM6_DOCS)):
+        try:
+            shutil.copy(f, docs_folder)
+        except Exception as e:
+            warn(e)
```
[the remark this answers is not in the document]
anton-seaice: My thinking is that we don't know if old versions of MOM6 or future versions of MOM6 will handle these files in the same way. If an exception occurs here, it might mean the model works and runs but payu prevents someone using it because of post-processing.

I think if we don't catch this Exception, then we would need to make committing runtime configurable to improve future/past/unknown compatibility between different payu & MOM versions?

payu/models/cesm_cmeps.py
```diff
@@ -336,14 +341,21 @@ class AccessOm3(CesmCmeps):
...
+        if (self.components["atm"] != "datm") :
+            raise RuntimeError(
+                "Access-OM3 comprises a data atmosphere model, but the atmospheric model in nuopc.runconfig is set "
+                f"to {self.components['atm']}."
+            )
+        if (self.components["rof"] != "drof") :
+            raise RuntimeError(
+                "Access-OM3 comprises a data runoff model, but the runoff model in nuopc.runconfig is set "
+                f"to {self.components['rof']}."
+            )
```
aidanheerdegen, line 353: &lt; Nods appreciatively at removing the [anti-pattern of using `assert` in non-test code](https://snyk.io/blog/the-dangers-of-assert-in-python/) &gt;

anton-seaice: Just for you @aidanheerdegen - apologies for sliding this into an unrelated PR 🤣

payu/models/cesm_cmeps.py
```diff
@@ -263,6 +263,11 @@ def _setup_checks(self):
         return True
 
     def archive(self):
+
+        # Move any the MOM_parameter_docs output back into the control repo and commit it for documentation
+        if 'mom' in self.components.values() :
+            mom6_save_docs_files(self)
```
aidanheerdegen, line 269: So it's necessary to call this explicitly because the `cesm_cmeps` driver doesn't directly call the `mom6.archive` function?

dougiesquire:
(Sorry, wrote this comment and forgot to click "Comment")

I tested configurations for:
- `AccessOm3` driver
  - [ ] `runlog: False` : **commits docs**
  - [x] `runlog: True` : works as expected
- `Mom6` driver
  - [ ] `runlog: False` : again, **commits docs**
  - [x] `runlog: True` : works as expected
  
So it's committing when it shouldn't but otherwise working as expected.

anton-seaice:
> So it's committing when it shouldn't but otherwise working as expected.

I think i fixed this.

I changed `model.config` to `model.expt.config`, I think thats correct :)

dougiesquire (approved):
I tested the `AccessOm3` driver again and now works as expected. Maybe worth waiting for approval from a real Payu developer before merging though.

payu/models/mom6.py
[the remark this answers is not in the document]
dougiesquire: Ok fair points - happy to keep it as a warning.

payu/models/mom6.py
```diff
@@ -41,6 +47,26 @@ def mom6_add_parameter_files(model):
...
+    mkdir_p(docs_folder)
+
+    # copy everything that matches MOM_parameter_doc.* to the control dir
+    for f in glob(os.path.join(model.work_path, MOM6_DOCS)):
+        try:
+            shutil.copy(f, docs_folder)
+        except Exception as e:
+            warn(e)
+
+    if model.expt.config.get('runlog', True): #if runlog true, default to true
```
jo-basevi, line 62: I think it's also possible to for runlog to be set as
```
runlog:
  enable:  true
```
When there's multiple options for `config.yaml`, it's normally transformed in `read_config()` in `fsops.py` to one format, but instead it's being parsed here: https://github.com/payu-org/payu/blob/d0fdbff1ca9b52c687d31dbb7773752b0b5223f9/payu/runlog.py#L39-L47
I don't think `model.expt.runlog.enabled` is accessible as runlog is initialised in experiment  class after the models were created.. 

If it's helpful, I can add a commit transforming the runlog config in read_config, so you just need to check `if model.expt.config.get('runlog', {}).get('enable', True)` above?

anton-seaice: Thanks Jo!

I didn't quite follow this bit:

> I don't think `model.expt.runlog.enabled` is accessible as runlog is initialised in experiment class after the models were created..

The new function runs after the model runs, so won't this `model.expt.runlog.enabled` be available still ?

anton-seaice: I tried `model.expt.runlog.enabled` instead of `model.expt.config.get('runlog', True)` and it worked fine, let me know which option you prefer!

jo-basevi: > The new function runs after the model runs, so won't this model.expt.runlog.enabled be available still ?

Yeah you are right - that's my bad sorry! The model class would be pointing to the same Experiment class instance that was updated..

In that case, I think `model.expt.runlog.enabled` would work fine

anton-seaice: Thanks Jo! - I made this change, is there a test for checking 

`model.expt.runlog.enabled` for the options of runlog True/False and runlog.enable True/False

I added one to the om3 test but there is probably a better place for it ? (test_setup?)

jo-basevi: Changes look good thanks. Yeah I can't see a current generic test for `model.expt.runlog.enabled`, so I think `test_setup.py` out of the existing test files makes the most sense.

anton-seaice: okie done

jo-basevi (dismissed):
The changes look good to me - thanks @anton-seaice!

jo-basevi (approved):

payu/models/mom6.py
```diff
@@ -41,6 +47,26 @@ def mom6_add_parameter_files(model):
         else:
             model.config_files.extend(filenames)
 
+def mom6_save_docs_files(model):
+    """Add docs files created as MOM output back to the control directory"""
+    docs_folder = os.path.join(model.control_path, 'docs')
+    mkdir_p(docs_folder)
+
+    # copy everything that matches MOM_parameter_doc.* to the control dir
+    for f in glob(os.path.join(model.work_path, MOM6_DOCS)):
```
chrisb13, line 56: What happens if glob returns empty? (consider error trap [example](https://github.com/chrisb13/GoGGoNEMO/blob/main/setupNEMO_ARC2.py#L56))

test/models/test_mom6.py
```diff
@@ -126,6 +127,126 @@ def test_mom6_add_parameter_files(input_nml,
...
+    # Function to test
+    payu.models.mom6.mom6_save_docs_files(mom_parameter_doc)
+
+    # Check files are added to control_path
+    for file in ["MOM_parameter_doc.layout"]:
+        filename = os.path.join(mom_parameter_doc.control_path, "docs", file)
+        assert os.path.isfile(filename)==True , "docs/MOM_parameter_doc.* do not exist"
+        os.remove(filename)
+    
+    assert repo.head.commit == initial_commit,  "Payu did not commit MOM_parameter_doc.layout"
```
chrisb13, line 248:
```suggestion
    assert repo.head.commit == initial_commit,  "Payu incorrectly commited the mom_parameter_doc.layout"
```

payu/models/mom6.py
anton-seaice: Turn out this is fine. I will add a test but 

```
for i in []:
...     print(i)
... 
``` 

prints nothing and doesn't give an error

anton-seaice:
Its me again ... needing something approved again :-)

dougiesquire (dismissed):

anton-seaice:
This is ready to go (again!) @dougiesquire

dougiesquire (approved):
Thanks @anton-seaice. Looks good to me and tested successfully

PR #570, opened 2025-02-18
Add new cable 3 namelists to ESM1.6 driver

### LINKED ISSUE #569: Add ESM1.6 specific config files to driver
ESM1.6 adds two cable namelists to be read by the UM: `pft_params.nml` and `soil_params.nml`. These should be added to the ESM1.6 driver. Jhan reports that they should be optional configuration files.

blimlim:
Closes #569 

This PR adds two optional configuration files, `soil.nml` and `pft_params.nml` to the UM via the `access-esm1.6` driver. 

I'll add some comments elaborating on the implementation, but any feedback and suggested changes are welcome!

### DIFF
--- a/payu/models/access_esm1p6.py
+++ b/payu/models/access_esm1p6.py
@@ -81,6 +81,12 @@ class AccessEsm1p6(Model):
                 # Simulation length in seconds for new run
                 model.runtime_key = "runtime"
 
+            if model.model_type == 'um':
+                # Additional Cable 3 namelists
+                model.optional_config_files = list(
+                    set(['pft_params.nml', 'soil.nml']) |
+                    set(model.optional_config_files)
+                )
 
     def setup(self):
         if not self.top_level_model:
--- a/payu/models/access_esm1p6.py
+++ b/payu/models/access_esm1p6.py
@@ -83,6 +83,9 @@ class AccessEsm1p6(Model):
 
             if model.model_type == 'um':
                 # Additional Cable 3 namelists
+
+                # Using set as this initialised twice and would otherwise
+                # contain duplicates
                 model.optional_config_files = list(
                     set(['pft_params.nml', 'soil.nml']) |
                     set(model.optional_config_files)
--- a/test/models/test_access_esm1p6.py
+++ b/test/models/test_access_esm1p6.py
@@ -1,0 +1,150 @@
+import copy
+import os
+import shutil
+
+import pytest
+
+import payu
+
+from test.common import cd, expt_workdir
+from test.common import tmpdir, ctrldir, labdir, workdir, archive_dir
+from test.common import config as config_orig
+from test.common import write_config
+from test.common import make_all_files
+from test.common import config_path
+
+verbose = True
+
+
+def setup_module(module):
+    """
+    Put any test-wide setup code in here, e.g. creating test files
+    """
+    if verbose:
+        print("setup_module      module:%s" % module.__name__)
+
+    # Should be taken care of by teardown, in case remnants lying around
+    try:
+        shutil.rmtree(tmpdir)
+    except FileNotFoundError:
+        pass
+
+    try:
+        tmpdir.mkdir()
+        labdir.mkdir()
+        ctrldir.mkdir()
+        archive_dir.mkdir()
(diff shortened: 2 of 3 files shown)


payu/models/access_esm1p6.py
```diff
@@ -81,6 +81,12 @@ def __init__(self, expt, name, config):
                 # Simulation length in seconds for new run
                 model.runtime_key = "runtime"
 
+            if model.model_type == 'um':
+                # Additional Cable 3 namelists
+                model.optional_config_files = list(
+                    set(['pft_params.nml', 'soil.nml']) |
+                    set(model.optional_config_files)
+                )
```
blimlim, line 89: I'd planned  to use `model.optional_config_files.extend(...)` in both the `um.py` and `access_esm1p6.py` drivers, to avoid dependence on the order that the drivers were called. 

However, since the `access_esm1p6` model is actually initialised twice: [first as a submodel](https://github.com/payu-org/payu/blob/c45258797fd489e8af9cb4b09f66298cc5c0f4a6/payu/experiment.py#L162-L166), and [then as the top-level model](https://github.com/payu-org/payu/blob/c45258797fd489e8af9cb4b09f66298cc5c0f4a6/payu/experiment.py#L169-L174), this would result in the new namelists being added twice. The following is the output from printing `model.optional_config_files` during the init stage:
```
 ['input_atm.nml', 'parexe', 'pft_params.nml', 'soil.nml']
 ['input_atm.nml', 'parexe', 'pft_params.nml', 'soil.nml', 'pft_params.nml', 'soil.nml']
```

Hence I ended up using `set` instead. It is a bit more ungainly and less obvious what it is doing though.

aidanheerdegen: Probably worth just adding to the code comment about why `set` is used. e.g. 

```
# Using set as this is initialised twice and would otherwise contain duplicates
```

Arguably should use `set` rather than `list` in all drivers eh?

aidanheerdegen: Also initialising twice seems kinda mad. Good catch.

aidanheerdegen (dismissed):
Approved, but is now the right time to start adding tests for this driver?

payu/models/access_esm1p6.py
blimlim: Good idea, I've just added this comment in

blimlim:
> is now the right time to start adding tests for this driver?

Good idea. I've just added in some small tests for this driver, which check:
1. No duplicates appear in the `um.optional_config_files`
2. The new driver correctly adds new files to the `um.optional_config_files`.

Let me know if these look ok to you

test/models/test_access_esm1p6.py
```diff
@@ -0,0 +1,150 @@
...
+        len(esm1p6_um_model.optional_config_files) ==
+        len(set(esm1p6_um_model.optional_config_files))
+    )
+
+    # Check that esm1p6 driver added new config files compared
+    # to the standalone UM.
+    assert (
+        set(um_standalone_model.optional_config_files) <
+        set(esm1p6_um_model.optional_config_files)
+    )
```
aidanheerdegen, line 150: This tests that the `esm1p6` driver has added files, but not which ones. Is this worth checking explicitly or would you rather allow the driver to add more files and not have to alter the test?

aidanheerdegen (approved):
Looks good. Happy to approve, just had one small question about one of the tests.

test/models/test_access_esm1p6.py
blimlim: That was my original thinking, though it's a good point that it makes the test not very specific... Perhaps being more explicit is helpful in this case, so that you know that the right files are actually being added

aidanheerdegen: I'll leave it up to you to decide, but in general I don't think there is a massive downside to it being more specific.

blimlim: Yeah, that makes sense. Added in [add4045](https://github.com/payu-org/payu/pull/570/commits/add40455181a161bde9a6a776fb537218ac54598)

aidanheerdegen (approved):
Nice!

Issue #571, opened 2025-02-19
Drop MATM support from ESM1.6 driver


blimlim:
The `access_esm1p6.py` driver contains several sections which are specific to the matm model. E.g:
https://github.com/payu-org/payu/blob/b00677fa86c220ea3aedbd59bfd2852a63257866/payu/models/access_esm1p6.py#L70-L74

These are carried over from the ESM1.5 driver, however are not relevant for ESM1.6 and can be removed.

anton-seaice:
Implemented in #573

PR #573, opened 2025-04-03
remove matm

### LINKED ISSUE #486: Drop support for matm model driver
The [`matm` model driver ](https://github.com/payu-org/payu/blob/master/payu/models/matm.py) is no longer being used, supported or updated.

https://github.com/COSIMA/matm

It was superseded by `yatm` and is not being actively used. Because of this we have no useful way of knowing if we're breaking it.

I propose we drop support for it, and delete it. 

Older versions of `payu` can always be used if, in extremis, in the highly unlikely event someone wants to use it.

anton-seaice:
Closes #486 and #571 

Deprecate matm


aidanheerdegen (changes requested):
LGTM, but it's [still referenced in the docs](https://github.com/payu-org/payu/blob/39e4b9bdabceca3836bc25f0fcfe411d0c0b9f13/docs/source/config.rst?plain=1#L183-L186). 

We should update those to a newer example.

Any suggestions @jo-basevi?

blimlim (changes requested):
Thanks @anton-seaice for picking this up. This looks good to me, I just had a couple small suggestions outside of the changed lines.

In `access.py`, I think

```
if model.model_type in ('cice'):
```
in [line 102](https://github.com/payu-org/payu/blob/42a7c0e25be03de0246b7531f978e2e91e3c7960/payu/models/access.py#L102) and 
```python
if model.model_type == 'cice':
```
in [line 206](https://github.com/payu-org/payu/blob/42a7c0e25be03de0246b7531f978e2e91e3c7960/payu/models/access.py#L206) can be merged together.

I think the if statement
```
if model.model_type == 'cice':
```
in [line 192](https://github.com/payu-org/payu/blob/42a7c0e25be03de0246b7531f978e2e91e3c7960/payu/models/access.py#L192) can also be removed.

Likewise the two if statements in [line 237 ](https://github.com/payu-org/payu/blob/42a7c0e25be03de0246b7531f978e2e91e3c7960/payu/models/access.py#L237) and [260](https://github.com/payu-org/payu/blob/42a7c0e25be03de0246b7531f978e2e91e3c7960/payu/models/access.py#L260) should be able to be combined


Just remembered the same changes can also be made in the `access-esm1p6.py` driver too

jo-basevi:
> Any suggestions @jo-basevi?

Looks good to me - I don't have any additional suggestions

blimlim:
> We should update those to a newer example.

As the example's using the ESM1.5 driver, I'm wondering if it would be easiest to swap it over to the UM?

```
 - name: atmosphere
      model: um
      ncpus: 240
      exe: um_hg3.exe
```

anton-seaice:
> LGTM, but it's [still referenced in the docs](https://github.com/payu-org/payu/blob/39e4b9bdabceca3836bc25f0fcfe411d0c0b9f13/docs/source/config.rst?plain=1#L183-L186).
> 
> We should update those to a newer example.
> 
> Any suggestions @jo-basevi?

The whole example is quite dated:

https://github.com/payu-org/payu/blob/39e4b9bdabceca3836bc25f0fcfe411d0c0b9f13/docs/source/config.rst?plain=1#L180-L200

Should I set based on esm1.5 ?

anton-seaice:
Thanks @blimlim - I think I made the changed suggested

payu/models/access.py
```diff
@@ -217,8 +202,9 @@ def setup(self):
                 f90nml.write(cpl_nml, nml_work_path + '~')
                 shutil.move(nml_work_path + '~', nml_work_path)
 
-            if model.model_type == 'cice':
-                if model.prior_restart_path and not self.expt.repeat_run:
+            if ( model.model_type == 'cice'
+                and model.prior_restart_path
+                and not self.expt.repeat_run ):
```
blimlim, line 207:
```suggestion
                if  model.prior_restart_path and not self.expt.repeat_run:
```

blimlim: Originally thinking something like this, but now that I think about it more, I think it's ok either way!

blimlim (commented):
Thanks @anton-seaice, just to clarify the suggestion around [this block](https://github.com/payu-org/payu/blob/f7ebb9b66a7a93029ddda74fe14745df101eed90/payu/models/access.py#L191-L197) in both access.py and access-esm1p6.py, I think the following change can be made, as the code already falls under the `if model.model_type == 'cice'` in line 102:
```diff
-                if model.model_type == 'cice':
-                    if self.expt.counter and not self.expt.repeat_run:
-                        cpl_group['jobnum'] = (
-                            1 + self.expt.counter
-                        )
-                    else:
-                        cpl_group['jobnum'] = 1
+                if self.expt.counter and not self.expt.repeat_run:
+                    cpl_group['jobnum'] = (
+                        1 + self.expt.counter
+                    )
+                else:
+                    cpl_group['jobnum'] = 1
```

payu/models/access.py
anton-seaice: Ah yes - much better !

blimlim:
> Should I set based on esm1.5 ?

That sounds good. I think either an ESM1.5 or OM2 config should be ok there

anton-seaice:
I pushed a newer config ... i wasn't really sure what to do with the `input` lines ?

jo-basevi:
I'm not really sure what to do for the inputs either.. I feel most of the value of the example is just seeing what the `submodel` section of the configuration could look like, so I think it achieves that.

blimlim (approved):
Thanks @anton-seaice, this is looking good to me! I think the changes do the docs look good

jo-basevi (approved):
Looks good to me

aidanheerdegen (approved):
LGTM2

PR #582, opened 2025-06-06
Add exe_prefix option to documentation

### LINKED ISSUE #578: Document exe_prefix configuration option
I would benefit if payu had a way to prepend an arbitrary string in front of the model exe. For example, when I am running the vtune profilers, my mpirun command looks like:

```
mpirun <mpi-params> <aps -r aps-output-dir> <model-exe>
mpirun <mpi-params> <vtune -r vtune-output-dir -- > <model-exe>
```

Now, from what I can tell, there is currently no way to specify the arbitrary string `<aps -r aps-output-dir>` to prepend the actual exe. Would it be possible to add such an "exe-loader" feature? I could also use it to run `valgrind` - that would help me debug the [division by zero error](https://github.com/ACCESS-NRI/MOM5/issues/34) affecting the oneAPI ESM1.6 PI config at the moment.

Such a feature could also be used for "LD_PRELOAD" and swap out dynamic libraries at runtime.

jo-basevi:
Add a brief description of  `exe_prefix` to the Model configuration section.
Closes #578

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -172,6 +172,10 @@ configuration.
    modules).
    Various model drivers typically define their own default executable names.
 
+``exe_prefix``
+   Optional string to add in front of the model executable in the MPI run
+   command. This can be useful for configuring profilers or valgrind.
+
 ``submodels``
    If one is running a coupled model containing several submodels, then each
    model is configured individually within a ``submodel`` namespace, such as in
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -173,8 +173,10 @@ configuration.
    Various model drivers typically define their own default executable names.
 
 ``exe_prefix``
-   Optional string to add in front of the model executable in the MPI run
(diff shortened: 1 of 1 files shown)


docs/source/config.rst
```diff
@@ -172,6 +172,10 @@ configuration.
    modules).
    Various model drivers typically define their own default executable names.
 
+``exe_prefix``
```
manodeep, line 175: Do we want explicitly say that the `exe_prefix` strings comes before the `exe` but after the `mpirun` (+ any mpi flags)?

manodeep: It's a minor comment but in case that helps future users.

manodeep (approved):
Excellent - thanks! LGTM

manodeep:
Hmm may be I don't have write access to the org and my approving review does not count.

jo-basevi:
Thanks again for the review :)

PR #584, opened 2025-06-13
Move running archive userscript to before collation job is submitted

### LINKED ISSUE #583: Payu collate being submitted before archive userscripts are run
The [payu docs](https://payu.readthedocs.io/en/stable/config.html#user-processing) for `archive` user-scripts has: 

```
archive
     User-defined command to be called after model archival, but prior to any post-processing operations, such as payu collate
```

However payu submits the collate job before running the archive user-scripts:
https://github.com/payu-org/payu/blob/c2a1b5865dc859451b669470b1e1a9d73df5d5db/payu/experiment.py#L840-L858

This means collation and archive userscript could modify the outputs at the same time. I think the archive userscript just needs to be moved up to before the collate job is submitted.

jo-basevi:
Switched the order of collation submission and running the archive userscript to ensure the archive user script completes before collation starts.

Closes #583

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -835,6 +835,11 @@ class Experiment(object):
         elif py_libpath not in ld_libpaths.split(':'):
             os.environ['LD_LIBRARY_PATH'] = f'{py_libpath}:{ld_libpaths}'
 
+        # Run archive user script before collation job is submitted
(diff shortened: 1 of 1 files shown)


CodeGat (approved):
Nothing I can think of improving here, looks good

Issue #600, opened 2025-07-10
FileNotFoundError when restart files are processor-split pieces


minghangli-uni:
`cesm_cmeps.archive()` currently assumes every filename listed in the component rpointer files exists verbatim.

However when `PARALLEL_RESTARTFILES` is enabled, MOM6 writes processor-split restarts (*.nc.0000, .0001, …). Payu therefore raises FileNotFoundError during the archive step.

A related issue can be found https://github.com/ACCESS-NRI/access-om3-configs/issues/592

anton-seaice:
I have deja-vue, although that was a subtly different was of MOM6 making multiple restart files :)

https://github.com/payu-org/payu/pull/432

minghangli-uni:
Thanks @anton-seaice This is a little bit different problem

PR #601, opened 2025-07-11
Support processor-split restart files in cesm_cmeps.archive()

minghangli-uni:
Closes https://github.com/payu-org/payu/issues/600

When `PARALLEL_RESTARTFILES` is enabled, MOM6 writes restart files in a processor-split format such as *.nc.0000, *.nc.0001, but the corresponding rpointer files still refer to the unified name (e.g., access-om3.mom6.r.1900-01-02-00000.nc).

Previously payu assumed this exact filename exists and attempted to archive it directly, resulting in FileNotFoundError.


minghangli-uni:
No, we can’t restart directly from uncollated restart files. It will result in file not found errors. I’ve been considering if this could be done through payu, but as far as I can tell, payu currently only supports collating the previous restart files.

```
payu collate -i K

This will also collate restart ``K-1`` if ``restart: true`` in the ``collate``
section of the configuration file.
```

dougiesquire:
>  It will result in file not found errors

From MOM or Payu?

minghangli-uni:
For mom6, I think yes

https://github.com/ACCESS-NRI/MOM6/blob/df33a6743d0bf4a0ea9d710e9dd18319ac63d4d9/src/framework/MOM_restart.F90#L1925-L1935

minghangli-uni:
yes correct

dougiesquire (commented):
Thanks @minghangli-uni. I find the approach of adding the additional loop to create `tmp_restart_files` that then overwrites `restart_files` a little unclear. Could we instead just add the logic to include uncollated restarts into the original `for pointer in pointer_files` loop (L311-L322)?

Also I think we should restrict the logic a little so it only looks for uncollated mom restarts of the expected form `*.nc.[0-9]{4}`

So something like (untested code):

```python
for pointer in pointer_files:
    try:
        with open(pointer,"r") as f:
            for restart in f.readlines():
                restart_file = os.path.join(
                    self.work_path,
                    restart.rstrip()
                )
                if os.path.exists(restart_file):
                    restart_files.append(restart_file)
                elif "mom" in restart_file:
                    # Check for uncollated MOM output
                    uncollated_restarts = glob.glob(
                        restart_file + ".[0-9][0-9][0-9][0-9]"
                    )
                    if uncollated_restarts:
                        restart_files.extend(uncollated_restarts)
                else:
                    raise FileNotFoundError(
                        f"Restart file {restart_file} does not exist"
                    )
    except FileNotFoundError:
        print(f"payu: error: Model has not produced pointer file {pointer}")
```

As @anton-seaice suggested, I think it would be a good idea to also add a test or two.

minghangli-uni:
Thanks @dougiesquire. I did a similar thing. I listed the uncollated restart files directly in rpointer.ocn like this:

```
$ cat work/rpointer.ocn 
access-om3.mom6.r.1900-01-02-00000.nc.0000
access-om3.mom6.r.1900-01-02-00000.nc.0001
access-om3.mom6.r.1900-01-02-00000.nc.0002
access-om3.mom6.r.1900-01-02-00000.nc.0003
access-om3.mom6.r.1900-01-02-00000.nc.0004
access-om3.mom6.r.1900-01-02-00000.nc.0005
```

but this resulted in the following error:

```
   15 FATAL from PE     4: NetCDF: Index exceeds dimension bound: netcdf_read_data_3d: file:./access-om3.mom6.r.1900-01-02-00000.nc.0000- variable:Temp
   16 
   17 
   18 FATAL from PE    90: NetCDF: Index exceeds dimension bound: netcdf_read_data_3d: file:./access-om3.mom6.r.1900-01-02-00000.nc.0000- variable:Temp
   19 
   20 
   21 FATAL from PE    97: NetCDF: Index exceeds dimension bound: netcdf_read_data_3d: file:./access-om3.mom6.r.1900-01-02-00000.nc.0000- variable:Temp
   22 
   23 
   24 
   25 FATAL from PE   191: NetCDF: Index exceeds dimension bound: netcdf_read_data_3d: file:./access-om3.mom6.r.1900-01-02-00000.nc.0000- variable:Temp
   26 
   27 FATAL from PE   185: NetCDF: Index exceeds dimension bound: netcdf_read_data_3d: file:./access-om3.mom6.r.1900-01-02-00000.nc.0000- variable:Temp
```

It seems like there are still something unclear to me whether mom6 can actually restart from uncollated restart files.

dougiesquire:
> It seems like there are still something unclear to me whether mom6 can actually restart from uncollated restart files.

Yeah okay. We definitely need to understand MOM behaviour before deciding what to do with Payu

dougiesquire:
> That’s why I attempted to list the uncollated restart files explicitly in rpointer.ocn, rather than using just the basename.

It would definitely be better if the pointer was correct for the restart files that exist. Although that requires a MOM code change (ideally) and that Payu collate updates the pointer file.

Before we get too deep into discussions about the best approach in Payu, I think we need to understand [MOM's restrictions](https://github.com/payu-org/payu/pull/601#issuecomment-3060816309)

dougiesquire:
Yeah, unfortunately it seems that the NUOPC cap imposes limitations

dougiesquire:
> Yeah, unfortunately it seems that the NUOPC cap imposes limitations

Actually not really. I think restarting from uncollated files only works if using restart files that have been automatically named by MOM. The code snippet you linked above is within `if ((fname(1:1)=='r') .and. ( len_trim(fname) == 1)) then`.

minghangli-uni:
I’m not entirely sure, but I ran a test without the nuopc cap and renamed the restart files. MOM6 was still able to read the uncollated restart files successfully.
```
$ ls access-om3.mom6.r.1900-01-02-00000.nc.00*
access-om3.mom6.r.1900-01-02-00000.nc.0000  access-om3.mom6.r.1900-01-02-00000.nc.0006
access-om3.mom6.r.1900-01-02-00000.nc.0001  access-om3.mom6.r.1900-01-02-00000.nc.0007
access-om3.mom6.r.1900-01-02-00000.nc.0002  access-om3.mom6.r.1900-01-02-00000.nc.0008
access-om3.mom6.r.1900-01-02-00000.nc.0003  access-om3.mom6.r.1900-01-02-00000.nc.0009
access-om3.mom6.r.1900-01-02-00000.nc.0004  access-om3.mom6.r.1900-01-02-00000.nc.0010
access-om3.mom6.r.1900-01-02-00000.nc.0005  access-om3.mom6.r.1900-01-02-00000.nc.0011

```

dougiesquire:
Sorry, could you provide a bit more detail about what exactly you tested?

dougiesquire:
> I’m not entirely sure, but I ran a test without the nuopc cap and renamed the restart files. MOM6 was still able to read the uncollated restart files successfully.
> 
> ```
> $ ls access-om3.mom6.r.1900-01-02-00000.nc.00*
> access-om3.mom6.r.1900-01-02-00000.nc.0000  access-om3.mom6.r.1900-01-02-00000.nc.0006
> access-om3.mom6.r.1900-01-02-00000.nc.0001  access-om3.mom6.r.1900-01-02-00000.nc.0007
> access-om3.mom6.r.1900-01-02-00000.nc.0002  access-om3.mom6.r.1900-01-02-00000.nc.0008
> access-om3.mom6.r.1900-01-02-00000.nc.0003  access-om3.mom6.r.1900-01-02-00000.nc.0009
> access-om3.mom6.r.1900-01-02-00000.nc.0004  access-om3.mom6.r.1900-01-02-00000.nc.0010
> access-om3.mom6.r.1900-01-02-00000.nc.0005  access-om3.mom6.r.1900-01-02-00000.nc.0011
> ```

> I was using the build and an example from https://github.com/NOAA-GFDL/MOM6-examples/tree/dev/gfdl/ocean_only/benchmark. I’m not sure if this is related to fms_io.

@minghangli-uni and I chatted about this is person. In this example, @minghangli-uni is setting the restart filename using the  `RESTARTFILE` parameter in `MOM_input` (i.e. not passing the restart filename to the restart routines). This is why things work in this case, but not in ACCESS-OM3 - see https://github.com/ACCESS-NRI/access-om3-configs/pull/637#pullrequestreview-3015787893.

PR #607, opened 2025-07-21
Prepend container launcher script to pbs command, if it exists

jo-basevi:
Related to https://github.com/ACCESS-NRI/model-release-condaenv/issues/17 which requires running payu from an environment within a singularity container. 
This allows a custom launcher script to run at the start of the PBS job which launches a singularity container with a SquashFS overlay for the environment, and then runs the python and payu executable within that environment.

Could use a shorter or other name than `ENV_LAUNCHER_SCRIPT_PATH`? Now would be the time to change it if so..

I've manually tested that this does not add `ENV_LAUNCHER_SCRIPT_PATH` if it isn't set in the environment, and if it is, it then works with an launcher script implementation in the https://github.com/ACCESS-NRI/model-release-condaenv/pull/19 PR. 

Note: qsub call looks like `qsub <qsub_flags> -- /path/to/env-launcher-scripts/bin/launcher.sh /path/to/envs/payu/bin/python3.10 /path/to/envs/payu/bin/payu-run` where `/path/to/envs/payu/bin` directory is only accessible when running within the singularity container.

### DIFF
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -132,7 +132,11 @@ class PBS(Scheduler):
 
         # Check for custom container launcher script environment variable
         launcher_script = os.environ.get('ENV_LAUNCHER_SCRIPT_PATH')
-        if launcher_script and Path(launcher_script).is_file():
+        if (
+            launcher_script
+            and Path(launcher_script).is_file()
+            and os.access(launcher_script, os.X_OK)
+        ):
             # Prepend the container launcher script to the python command
             # so the python executable is accessible in the container
             python_exe = f'{launcher_script} {python_exe}'
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -6,6 +6,7 @@
 
 # Standard library
 import os
+from pathlib import Path
 import re
 import sys
 import shlex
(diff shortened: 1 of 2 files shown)


payu/schedulers/pbs.py
```diff
@@ -129,6 +130,13 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
         envmod.setup()
         envmod.module('load', 'pbs')
 
+        # Check for custom container launcher script environment variable
+        launcher_script = os.environ.get('ENV_LAUNCHER_SCRIPT_PATH')
+        if launcher_script and Path(launcher_script).is_file():
```
aidanheerdegen, line 135: Is it worth checking [if the file is also executable](https://docs.python.org/3/library/os.html#os.access)?

jo-basevi: Good point - Added in [8bc93d2](https://github.com/payu-org/payu/pull/607/commits/8bc93d2a3f370fb5df0354b525725ad535d16aa9)

aidanheerdegen (approved):
LGTM. Thanks.

I think you'd said you might add a test, but I'll leave that up to your discretion.

aidanheerdegen (approved):
Awesome. Thanks!

PR #608, opened 2025-07-21
List loaded modules after openmpi has been updated

### LINKED ISSUE #605: Change where loaded modules are listed in output
Currently the currently loaded modulefiles are displayed during `Experiment.load_modules()`, but this is before every module is loaded:

https://github.com/payu-org/payu/blob/c8343f79c77559cf48569da696e0b69c437b45cb/payu/experiment.py#L276

Later in the function, profiler modules are loaded if configured. Then later in `Experiment.run()` method, the openmpi module may be updated:

https://github.com/payu-org/payu/blob/c8343f79c77559cf48569da696e0b69c437b45cb/payu/experiment.py#L572-L578

This was then a bit confusing figuring out what openmpi module was actually run for an ACCESS-OM2 configuration as the PBS error logs would have

```
Currently Loaded Modulefiles:
 1) oasis3-mct/2023.11.09_2024.03.0     4) mom5/2023.11.09_2024.03.0   7) pbs
 2) libaccessom2/2023.10.26_2024.03.0   5) access-om2/2024.03.0
 3) cice5/2023.10.19_2024.03.0          6) openmpi/4.1.7-5kn4pgy
```

By the time the model was run, the openmpi was changed so it would've been:

```
Currently Loaded Modulefiles:
 1) oasis3-mct/2023.11.09_2024.03.0     5) access-om2/2024.03.0
 2) libaccessom2/2023.10.26_2024.03.0   6) pbs
 3) cice5/2023.10.19_2024.03.0          7) openmpi/4.0.2
 4) mom5/2023.11.09_2024.03.0
```

Initially I thought it would make more sense to have `envmod.lib_update` logic as part of `load_modules`, and then run `envmod.module('list')` at the end of the function. However `load_modules` is reused in `models.fms` in `fms_collate()`, which also needs to use `envmod.libs_update` with different inputs:

https://github.com/payu-org/payu/blob/c8343f79c77559cf48569da696e0b69c437b45cb/payu/models/fms.py#L106-L111

So the only idea I have is to move `envmod.module('list')` to after all the modules have been updated.

jo-basevi:
Move listing loaded modules until after openmpi has been potentially been updated. This should make it more clear what modules are loaded by the time the model runs

Closes #605

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -273,8 +273,6 @@ class Experiment(object):
         for mod in self.modules:
             envmod.module('load', mod)
 
-        envmod.module('list')
-
         for prof in self.profilers:
             prof.load_modules()
 
@@ -625,6 +623,9 @@ class Experiment(object):
             model_prog.append(os.path.join(model.work_path, model.exec_name))
 
             mpi_progs.append(' '.join(model_prog))
+
+        # List all loaded environment modules
+        envmod.module("list")
 
         cmd = '{runcmd} {flags} {exes}'.format(
             runcmd=mpi_runcmd,
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -109,6 +109,8 @@ def fms_collate(model):
         # the output
         collate_flags = " ".join([collate_flags, '-o'])
         envmod.lib_update(required_libs(mppnc_path), 'libmpi.so')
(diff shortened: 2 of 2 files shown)


aidanheerdegen (approved):
LGTM.

PR #610, opened 2025-07-21
Revert um calendar format change

### LINKED ISSUE #609: Revert UM calendar format change
In #599 I tried to modify the UM driver so that it read and wrote the UM restart date using `cftime.datetime` rather than `datetime.datetime` objects.

This change meant that `pyyaml` would read and write the dates as formatted strings. The pytests were modified to be consistent with this change. What I failed to consider was what would happen with existing restart files.

The quickest fix is to revert to reading and writing the UM dates using `datetime.date` objects and convert to/from `cftime.datetime` after/prior to reading/writing.

blimlim:
Closes #609.

Consistently read and write the `um_res.yaml` file using `datetime.datetime` objects. This maintains compatibility with the files that the driver writes, and maintains compatibility with other date manipulations within the UM driver which expect datetime.datetime objects (e.g. [here](https://github.com/payu-org/payu/blob/3f79a0d0a57f1e9bb86b20fe2a11ef4f5991a5a1/payu/models/um.py#L192-L198)). Convert to `cftime.datetime` after reading.

### DIFF
--- a/payu/models/um.py
+++ b/payu/models/um.py
@@ -26,7 +26,6 @@ from payu.fsops import mkdir_p, make_symlink
 from payu.models.model import Model
 import payu.calendar as cal
 
-UM_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
 # UM in access always uses proleptic gregorian calendar
 UM_CFTIME_CALENDAR = "proleptic_gregorian"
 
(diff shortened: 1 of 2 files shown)


blimlim:
@anton-seaice and @jo-basevi, this one should now be ready for review. It now reverts the `um.py` driver to (almost) match prior to the recent changes (see: https://github.com/ACCESS-NRI/payu/compare/c92ba25cf7118198f251f62547f0f14b87ef3ded..e3c113f55c447fd7a6bcb3d527cfb9b329b2d35f#diff-99e9f3532c6e21fbcc909e4de67d084d66705434bfc6688fc982b60a04c4624b) for a comparison of the lates changes with the UM driver prior to #599

jo-basevi (approved):
Thanks @blimlim, looks good to me

PR #614, opened 2025-07-22
Add netCDF4 as a explicit dependency

### LINKED ISSUE #613: Explicitly add netCDF4 as a dependency
From looking at previous pip installs, it looks like `netCDF4` was installed as part of `yamanifest` - e.g. old pip build:

```
Collecting netCDF4 (from nchash>=0.1.5->yamanifest>=0.3.4->payu==1.1.6+31.g2b3feea)
  Downloading netCDF4-1.7.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (1.8 kB)
```

However yamanifest has been updated ( `0.3.8` to `0.3.14`), and no longer has nchash as dependency.

So when install payu in a virtual environment and running `payu --version` would now get errors:

```
  File "/home/189/jb4202/payu_fork/payu/models/__init__.py", line 7, in <module>
    from payu.models.cice5 import Cice5
  File "/home/189/jb4202/payu_fork/payu/models/cice5.py", line 18, in <module>
    from netCDF4 import Dataset
ModuleNotFoundError: No module named 'netCDF4'
```

Anyway, `netCDF4` should be an explicit dependency in pyproject.toml as it's being directly imported in the code.

jo-basevi:
`netCDF4` is imported in CICE5 model driver, but was previously available in the environment due to an older version of `yamanifest` dependency. This PR adds `netCDF4` explicitly to the project dependencies.

Closes #613

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,8 @@ dependencies = [
     "cftime",
     "GitPython >=3.1.40",
     "ruamel.yaml >=0.18.5",
-    "packaging"
+    "packaging",
+    "netCDF4",
 ]
 
 [project.optional-dependencies]


aidanheerdegen (approved):
LGTM

jo-basevi:
Thanks for the review!

PR #616, opened 2025-07-22
Update pypa/gh-action-pypi-publish to v1.12.4

### LINKED ISSUE #615: Fix PyPI deployment
Payu PyPI deployment (https://github.com/payu-org/payu/actions/runs/16437302675/job/46449777005) has the following error:

```
Checking dist/payu-1.1.7-py3-none-any.whl: ERROR    InvalidDistribution: Metadata is missing required fields: Name,        
         Version.                                                               
         Make sure the distribution includes the files where those fields are   
         specified, and is using a supported Metadata-Version: 1.0, 1.1, 1.2,   
         2.0, 2.1, 2.2.   
```

jo-basevi:
This PR updates the [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) action used in the CD workflow to `v1.12.4` so it can support newer versions of `twine`.

Closes #615

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -64,11 +64,10 @@ jobs:
           auto-activate-base: false
           show-channel-urls: true
 
-      - name: Build and upload the conda package
-        uses: uibcdf/action-build-and-upload-conda-packages@c6e7a90ad5e599d6cde76e130db4ee52ad733ecf # v1.2.0
+      - name: Build and upload conda package
+        uses: ACCESS-NRI/action-build-and-upload-conda-packages@v2.0.1
         with:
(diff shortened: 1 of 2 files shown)


jo-basevi:
I don't think the conda publishing worked either so that will need to be updated too - so converting this PR to draft

CodeGat (approved):

jo-basevi:
I've updated `action-build-and-upload-conda-packages` to use ACCESS-NRI version in the CD: https://github.com/ACCESS-NRI/action-build-and-upload-conda-packages, and also added a step to the CI so it runs the same action to build the package but it doesn't do the final upload to anaconda step.

PR #619, opened 2025-08-01
Add in HPCpy

### LINKED ISSUE #568: HPCPy integration
[HPCpy](https://github.com/ACCESS-NRI/hpcpy) is a Python package for interacting with HPC scheduling systems. Using this in payu, may reduce the need for scheduler specific code. Currently, there is some development on a SLURM client: https://github.com/ACCESS-NRI/hpcpy/issues/28. See related payu issue for supporting different schedulers -  https://github.com/payu-org/payu/issues/182

Payu currently runs `qsub` with executable commands directly rather than a script, (e.g., `qsub <directives> -- /g/data/vk83/prerelease/apps/payu/dev/bin/python /g/data/vk83/prerelease/apps/payu/dev/bin/payu-run`). See related issue in `HPCpy` to support this: https://github.com/ACCESS-NRI/hpcpy/issues/24.

However, `HPCpy` supports script jinja templates which can be used to generate a script. So it could be possible to generate a job script that contains `{{python}} {{payu-cmd}}`. This could potentially help:
- module initialisation issues which requires a shell to be created (see https://github.com/payu-org/payu/issues/522). This will allow modules to be used in bash user-scripts. 
-  maybe support different compiler/mpi combinations with different script templates (see https://github.com/payu-org/payu/issues/336#issuecomment-2384586369)

There is also some support for dependency  strategy in `HPCpy`, this could be useful for running `sync` jobs - after a user-defined `postscript` (see some related discussion in this issue: https://github.com/payu-org/payu/issues/463#issuecomment-2244215195)

jo-basevi:
Storing some initial changes of adding in hpcpy. Installed hpcpy from the github repository into a python virtual environment as hpcpy is currently not available on PyPI (so I imagine the CI tests will fail...)

TODO:
- Test slurm client
- Use HPCpy for submitting post-scripts - add in job dependency for sync scripts?
- Add hpcpy to PyPI?
- Disable hpcpy debug logs but print job submission command (available when using `dry_run=True`)

A number of PBS specific directives are still required so can't use a generic client currently. 

Closes #568

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -167,7 +167,6 @@ def submit_job(script, config, vars=None):
     sched_name = config.get('scheduler', DEFAULT_SCHEDULER_CONFIG)
     sched_type = scheduler_index[sched_name]
     sched = sched_type()
-    cmd = sched.submit(script, config, vars)
-    print(cmd)
-
-    subprocess.check_call(shlex.split(cmd))
+    job = sched.submit(script, config, vars)
+    print(job.id)
+    return job
--- a/payu/schedulers/payu-submit.sh
+++ b/payu/schedulers/payu-submit.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+# Script to run payu command on a scheduled job, e.g. path/to/python path/to/payu-run
+{{python_exe}} {{payu_exe}}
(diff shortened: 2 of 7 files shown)


jo-basevi:
Implemented in #802

Issue #620, opened 2025-08-06
restart collation failed with FileNotFoundError


minghangli-uni:
`payu collate -d archive/restartxxx` failed with,

```
FileNotFoundError: [Errno 2] No such file or directory: 'archive/restart000/nuopc.runconfig'
```

`nuopc.runconfig` is not in restartxxx. The file is normally only present in either,
1. the matching `archive/outputxxx/`, or
2. the experiment `control/` directory.

PR #621, opened 2025-08-06
Autodetect nuopc.runconfig for restart-only collate

minghangli-uni:
closes https://github.com/payu-org/payu/issues/620

This PR patches collate method in CesmCmeps to automatically locate `nuopc.runconfig` when the user collates a bare restart directory with `payu collate -d archive/restartxxx` or `payu collate -d restartxxx`,

1. lookinig for sibling `outputxxx` directory obtained by `restartxxx -> outputxxx`
2. If absent, fall back to the copy in `control/`.

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -334,18 +334,14 @@ class CesmCmeps(Model):
             os.rmdir(self.work_restart_path)
 
     def collate(self):
-        
         self.output_path = os.path.realpath(os.path.abspath(self.output_path))
-        runconfig_path = os.path.join(self.output_path, NUOPC_CONFIG)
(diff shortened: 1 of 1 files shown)


anton-seaice:
> 2\. If absent, fall back to the copy in `control/`.

Can we just always do this ? 

The only place in sync we use the runconfig from is:

`         if "mom" in self.components.values(): `

So it's unlikely that would be different between control and output

aidanheerdegen:
> Is it correct to assume that if manually triggered, `payu collate -d ...` would be run from the control directory

Yeah pretty much. `payu collate` needs to read a `config.yaml` file in the control dir to be able to run. It is possible to specify a control directory from the command line. I don't know anyone who uses this functionality, but is that a scenario that would break this specific change?

anton-seaice (approved):
Thanks @minghangli-uni !

PR #627, opened 2025-08-26
fix: properly handle provision of optional configuration files

phellipecouto:
Enabled `optional_config_files` provision through `config.yaml`. Prior to this change, ROMS `varinfo` file was being considered a mandatory model config file and mistakenly hard coded into `config_files` list, leading to model initialisation failure when file name differed from `varinfo_seacofs.yaml`.

### DIFF
--- a/payu/models/roms.py
+++ b/payu/models/roms.py
@@ -12,7 +12,7 @@ from pathlib import Path
 from payu.models.model import Model
 from payu.fsops import mkdir_p
 
-config_files = ['varinfo_seacofs.yaml']
+config_files = []
 optional_config_files = []
 
 class Roms(Model):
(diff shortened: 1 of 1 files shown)


payu/models/roms.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, expt, name, config):
         self.optional_config_files = optional_config_files
```
aidanheerdegen, line 29: May as well just set these arrays to empty here and remove the global definitions above.

aidanheerdegen (approved):
LGTM. We can deal with docs later ...

payu/models/roms.py
phellipecouto: solved by this: https://github.com/payu-org/payu/pull/627/commits/270ea77a3d8817348a16361c8aed1cfab0ed2a59

aidanheerdegen:
> Can you point me to the right direction and I will assess if this would still work for the ROMS case?

I think what you've done is fine. Off the top of my head I can't think of any other drivers that have dynamic model config files, so I think it's fine.

I'll start a separate PR to add driver docs for more of the models, and get your input for ROMS then.

PR #630, opened 2025-09-17
Add copy restarts flag to ROMS driver

### LINKED ISSUE #629: ROMS driver does not support automatic restarts
## Context

When ROMS is run from a restart, rather than initial conditions, it requires two modifications to the input configuration file:

1. [Modify `NRREC`, usually setting it to -1](https://github.com/ACCESS-Community-Hub/roms-configs/blob/release-upwelling%2Btestcase/roms_upwelling.in#L1615)
2. [Change `ININAME` to point to the a restart file](https://github.com/ACCESS-Community-Hub/roms-configs/blob/release-upwelling%2Btestcase/roms_upwelling.in#L1076)

Using the [roms_upwelling case](https://github.com/ACCESS-Community-Hub/roms-configs/blob/release-upwelling%2Btestcase/roms_upwelling.in) as an example, this would mean changing `ININAME` to
```
ININAME == roms_rst.nc
```
because [the restart output file is defined to be `roms_rst.nc`](https://github.com/ACCESS-Community-Hub/roms-configs/blob/release-upwelling%2Btestcase/roms_upwelling.in#L1170).

## Problem

Using this above configuration as is, the ROMS driver does not support automatic restarts because `roms_rst.nc` is symlinked from `archive/restart000` into the `work` directory, which means it is read correctly, but the next restart output is written to the same name, which is the symlink to `archive/restart000`. 

This does not throw an error, instead the symptom is that `archive/restart001` is empty, it contains no `roms_rst.nc` because the file written to in the `work` directory is a symlink which is deleted during the archive step.

## Solutions

There are a number of different ways to solve this, but a couple that match current usage patterns:

1. Copy, rather than symlink, the restart file. This is the solution that is [already used in the `cice5` driver](https://github.com/payu-org/payu/blob/master/payu/models/cice5.py#L40), but would require a change to the model driver
2. Read and write restarts from different locations, e.g. read from `work/roms_rst.nc` as is done currently, but write to `work/RESTART/roms_rst.nc`

Solution 2 requires a change in the model configuration, e.g.
```
     RSTNAME == RESTART/roms_rst.nc
```
but this is not supported by the model driver, as it doesn't currently look in subdirectories of `work` for restart files

https://github.com/payu-org/payu/blob/master/payu/models/roms.py#L74-L79

so it would also require a change to the model driver.

With the current driver code, the only way to successfully restart is to change the name of the restart file every time the model is run, and make the same change to the model configuration.

aidanheerdegen:
Closes #629 

ROMS reads from and writes to the same restart file, so added a copy restarts flag as the simplest and easiest solution.

### DIFF
--- a/payu/models/roms.py
+++ b/payu/models/roms.py
@@ -25,6 +25,8 @@ class Roms(Model):
         self.config_files = []
         self.optional_config_files = []
 
+        self.copy_restarts = True
+
     def setup(self):
         ## handle mandatory config files
         if 'model_config' not in self.config:


aidanheerdegen:
I haven't added any tests, but I have run 3 cycles of the [upwelling roms test case](https://github.com/ACCESS-Community-Hub/roms-configs/blob/release-upwelling%2Btestcase/) and it works as expected: a new restart file is created each time and used correctly in subsequent runs.

aidanheerdegen:
This is a diff showing the restart files being picked up in the manifests

```diff
$ git diff HEAD^ manifests/restart.yaml
diff --git a/manifests/restart.yaml b/manifests/restart.yaml
index d4700cf..b205b05 100644
--- a/manifests/restart.yaml
+++ b/manifests/restart.yaml
@@ -3,7 +3,7 @@ version: 1.0
 ---
 work/roms_rst.nc:
   copy: true
-  fullpath: /scratch/xxXX/xxxXXX/roms/archive/roms-upwelling-release-upwelling+testcase-f4a04601/restart002/roms_rst.nc
+  fullpath: /scratch/xxXX/xxxXXX/roms/archive/roms-upwelling-release-upwelling+testcase-f4a04601/restart003/roms_rst.nc
   hashes:
-    binhash: 7c0dcd63936ddac5d02ea53adcdd1580
-    md5: 5db062f00903347876457375c2670754
+    binhash: edbcaa6317cb8d3862e98e189787d727
+    md5: 910af540519ad3ba3078355e5d0a7fd0
```

Note that `copy: true` because the restart manifest is copied, not symlinked.

helenmacdonald (approved):
Thanks @aidanheerdegen  - looks good to me

phellipecouto (approved):
@aidanheerdegen I'm happy with that solution and can see this working well, based on the results you obtained with the upwelling test run, for hindcast and forecast simulations. Thank you.

PR #639, opened 2025-10-27
Allow optional proxy url in telemetry config for local testing

jo-basevi:
In this  #637 PR, the telemetry config was changed as the persistent sessions for forwarding requests now uses an http proxy. However when testing with a local ACCESS-NRI services django server, I am unable to set a persistent sessions up with an http proxy. This PR makes the `telemetry_proxy_url` setting optional. 

I've tested sending records to my persistent session which has requests going to my local server, and sending records to the production server (that uses the http proxy setting  in requests).


aidanheerdegen (approved):
LGTM

Issue #648, opened 2025-12-11
exclude_uncollated=True is a dangerous default


aekiss:
Payu excludes uncollated files from the gdata sync by default
https://payu.readthedocs.io/en/stable/config.html#user-processing
> `exclude_uncollated` (Default: True if collation is enabled)
Flag to exclude uncollated files from being synced. This is equivalent to adding `--exclude *.nc.*.`

This is may lead to data loss if the uncollated files aren't manually collated before they're automatically deleted from scratch.

I'd advocate setting the default to False.

Related: https://github.com/ACCESS-NRI/access-om3-configs/issues/983

anton-seaice:
I can implement if folks agree

jo-basevi:
This change should now be in the latest `payu/dev` module. Thanks for raising this issue @aekiss!

PR #649, opened 2025-12-12
payu sync: Default exclude_uncollated to false

jo-basevi:
This PR changes the default setting of sync to `exclude_uncollated` files to always False. 

See related issue #648

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -463,7 +463,7 @@ User Processing
             - 'iceh.????-??-??.nc'
             - '*-IN-PROGRESS'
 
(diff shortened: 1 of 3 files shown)


jo-basevi:
Tested with a configuration with collation disabled for a run so there was uncollated files, then checked `--exclude *.nc.*` was added to rsync commands when `exclude_uncollated: true`, and no `--exclude *.nc.*`  was added when collation was re-enabled, and `exclude_uncollated` setting was removed, and all previous uncollated files were synced.

aidanheerdegen (approved):
LGTM

PR #664, opened 2026-01-20
657 check work dir payu run

### LINKED ISSUE #657: Better handling of existing work directory when invoking payu run
#### Description

When `payu run` is executed it does not check for the existence of a `work` directory until the the job submitted to the queue runs `payu-run`. That means a user wastes time while the job waits in the queue and fails almost instantly because there is already a `work` directory.

The current behaviour was intentional: it was important to note remove a possibly failed run until the user had examined it and determined the cause of the failure. So the onus was put on the user to `payu sweep` before invoking `payu run`.
 
However it is easy to forget to `sweep`, and it is also considered good practice to run `payu setup` at least once after modifications to the configuration, to ensure all manifests have been updated in an environment that has access to all filesystems so that the correct storage flags can be added to the PBS submission command.

The convenience option `-f` was added, so that `payu run -f` would automatically run `sweep` if the `work` directory exists.

However this can still catch out new users.

#### Proposed solution

if `-f` option is not specified, check if the `work` directory exists before queue submission, and warn the user that they either need to `sweep` or use `-f`, and abort the submission.

If we do have a beginner mode (see #655) this could offer to run `payu sweep` for them.

KAUR1984:
Fixes #657. 

**PR Changes:**
- [x] Added a conditional in `run_cmd.py` to check if `work` directory already exists when `payu-run` command is executed. If the `work` directory already exists, it shows an error message to the user. 
- [x] Color coded the error message in `red` using _Colorama_ and _logging_ libraries for users to easily find the error message.
- [x] Since we are using two libraries - _Colorama_ (third-party import) and _logging_ (standard library), the `logger.py` integrates both these libraries - and an error message could be written as `logger.error()` function, and it would automatically be color-coded.
- [x] Since  _Colorama_ is a third-party import, have added it as a dependency in `pyproject.toml`.

---

**Output: `payu run`**

Screenshot 2026-01-20 at 12 25 54

---

**Output: `payu-run`**

```
(payu-training-venv) [jk1943@gadi-login-06 mom6_double_gyre]$ payu-run
Use of these keys is deprecated: collate_walltime, collate_mem.
Instead use collate dictionary and subkey without 'collate_' prefix
laboratory path:  /scratch/tm70/jk1943/mom6
binary path:  /scratch/tm70/jk1943/mom6/bin
input path:  /scratch/tm70/jk1943/mom6/input
work path:  /scratch/tm70/jk1943/mom6/work
archive path:  /scratch/tm70/jk1943/mom6/archive
Use of these keys is deprecated: collate_walltime, collate_mem.
Instead use collate dictionary and subkey without 'collate_' prefix
Found experiment archive: /scratch/tm70/jk1943/mom6/archive/mom6_double_gyre-0364d62b
Use of these keys is deprecated: collate_walltime, collate_mem.
Instead use collate dictionary and subkey without 'collate_' prefix
nruns: 1 nruns_per_submit: 10 subrun: 1
payu: error: work path already exists: /scratch/tm70/jk1943/mom6/work/mom6_double_gyre-0364d62b.
             payu sweep and then payu run
```

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -6,7 +6,7 @@
    :copyright: Copyright 2011 Marshall Ward, see AUTHORS for details.
    :license: Apache License, Version 2.0, see LICENSE for details
 """
-
+# Standard imports
 import argparse
 import sysconfig
 import importlib
@@ -17,12 +17,14 @@ import subprocess
 import sys
 import warnings
 
+# Local imports
 import payu
 import payu.envmod as envmod
 from payu.fsops import is_conda
 from payu.models import index as supported_models
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 import payu.subcommands
+from payu.logger import setup_logger
 
 # Default configuration
 DEFAULT_CONFIG = 'config.yaml'
@@ -38,6 +40,7 @@ warnings.formatwarning = (
 
 def parse():
     """Parse the command line inputs and execute the subcommand."""
+    setup_logger()
     parser = generate_parser()
 
     # Display help if no arguments are provided
--- a/payu/logger.py
+++ b/payu/logger.py
@@ -4,9 +4,6 @@ import logging
 # Third Party imports
 from colorama import init, Fore, Style
 
-# Color Formatter: Initialize Colorama for cross-platform compatibility and auto-reset
-init(autoreset=True)
-
 class ColoredFormatter(logging.Formatter):
     """A custom formatter to add colors based on log level."""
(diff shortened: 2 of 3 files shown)


Qian-HuiChen:
It looks very nice!
Would it be good to show what the current error message looks like? I am a bit curious.

KAUR1984:
> It looks very nice! Would it be good to show what the current error message looks like? I am a bit curious.

Great point! Have added screenshot in the [original description](https://github.com/payu-org/payu/pull/664#issue-3831551790).

PR #670, opened 2026-01-21
Added shared_path in sync path. 

### LINKED ISSUE #462: Sync Path
Should the sync path configure follow the style as short path? Where the experiment name is added to the path provided.

i.e. if:

```
sync: 
    enable: True
    path: /g/data/$PROJECT/$USER
```

should this create a folder 

`/g/data/$PROJECT/$USER/<<experiment_name+uuid>>`

automatically to sync to?

This seems especially useful when using the `payu branch` from a point in another experiment and you don't want to override the existing results ?

Current behavior is it will create the output folders directly in the _path_ directory

Qian-HuiChen:
**Function**
Added `shared_path` in `sync` section of `config.yml`.
Prioritise cutomised `path`, sync to `path` when provided.
When `path` is not provided, sync to `shared_path/<expt_name>`.
When either path is provided, an `ValueError` will be raised.

**Tests**
Passed pytest by provided only `shared_path`, `path` or neither.
Another testing is through running `payu-sync`. Data are sync to the expected archive directory.
Closes #462

### DIFF
--- a/payu/sync.py
+++ b/payu/sync.py
@@ -148,8 +148,8 @@ class SyncToRemoteArchive():
             # Check if sync archive UUID {dest_uuid} matches 
             # the local payu archive UUID {self.expt.metadata.uuid}
             if dest_uuid != None and dest_uuid != self.expt.metadata.uuid:
-                print(f"payu: error: UUID of experiment metadata in sync archive"
-                      f"({self.destination_path})"
+                print(f"payu: error: UUID of experiment metadata in sync archive "
+                      f"({self.destination_path}) "
                       f"does not match current experiment UUID. "
                       "Refusing to sync to avoid overwriting existing output.")
                 raise ValueError("payu: error: Mismatched experiment UUIDs in sync destination.")


payu/sync.py
```diff
@@ -124,18 +137,36 @@ def add_extra_source_paths(self):
...
+        If yes, check if UUID is the same as current experiment UUID.
+        If not the same, raise error."""
+        metadata_path = Path(self.destination_path) / METADATA_FILENAME
+        if os.path.exists(metadata_path):
+            metadata = YAML().load(metadata_path)
+            dest_uuid = metadata.get(UUID_FIELD, None)
+            if dest_uuid != self.expt.metadata.uuid:
+                print(f"payu: error: UUID of experiment metadata in sync "
+                      f"archive does not match current experiment UUID. "
+                      "Refusing to sync to avoid overwriting existing output.")
```
jo-basevi: Could add a line of what the uuid is and what the destination path is. E.g. something like

```
{dest_uuid} (sync archive UUID) != {self.expt.metadata.uuid} (local payu archive)
Sync archive metadata path: {self.metadata_path}
```

jo-basevi (commented):
Thanks for adding in the metadata UUID check! It looks good to me :)

Qian-HuiChen:
Ready for review 😃 @KAUR1984

KAUR1984 (dismissed):
Lgtm! Thanks!

payu/sync.py
Qian-HuiChen: What do you think of print this inside `check_UUID()`?
```
print(f"payu: error: UUID of experiment metadata in sync archive"
                      f"({self.destination_path})"
                      f"does not match current experiment UUID. "
                      "Refusing to sync to avoid overwriting existing output.")
```
I am thinking only print out the `destination_path` when there is an error. When it goes well, payu won't show up sync path in case the user get information overloaded (?)

jo-basevi: That looks good to me! 

I think it'll be worth adding the destination path still as it's probably the most important bit of information from this job for the user. It is possible to infer it from the `rsync` commands in the output, but it'll be good to be clear where everything is being synced to.

PR #677, opened 2026-01-30
Improve handling hidden files and directories in manifests.

### LINKED ISSUE #377: Handling of hidden input files and directories
We recently realised that payu was including the `/g/data/ik11/inputs/JRA-55/RYF/v1-4/make_rhuss/.git ` directory and files therein in the input of a given run. That means it creates symlinks for all those files and also includes them in the manifest.

Although it has no effect on the calculation, the fact that those files are in the manifest is a bit annoying, as it makes it unnecessarily long and harder to read.

I was wondering if there was a legitimate case where a user would want to use hidden files/directories in a calculation. If not, them maybe payu should simply ignore those files. What do you think?

Qian-HuiChen:
Hidden files and directories are in default set to be ignored. 
Users **do not need to include** `ignore` or `ignore_path` entry in `manifest` of `config.yaml`.

If users would like to include the hidden files (directories), users can add
```
ignore:
ignore_path:
```
and leave them empty in `manifest` section of `config.yaml`.

Users can also customize their ignore pattern through 
```
ignore: 
    - <custom1>
    - <custom2>
ignore_path:
    - <custom_path1>
    - <custom_path2>
```
Closes #377

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -275,7 +275,7 @@ section for details.
       ``restart`` (*Default: global reproduce flag*)
             Enforce restart file reproducibility.
 
-``ignore`` (*Default: [.\*]*):
+``ignore`` (*Default:* ``[.*]``):
       List of ``glob`` patterns which match files to ignore when scanning input directories.
       A file will be ignored if any part of its full path matches one of the patterns.
       This is an array, so multiple patterns can be specified on multiple lines. 
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -275,17 +275,18 @@ section for details.
       ``restart`` (*Default: global reproduce flag*)
             Enforce restart file reproducibility.
 
-``ignore`` (*Default: .*):
+``ignore`` (*Default: [.\*]*):
       List of ``glob`` patterns which match files to ignore when scanning input
       directories. This is an array, so multiple patterns can be specified on
-      multiple lines. The default is .** which ignores all hidden files on a
-      POSIX filesystem. To include hidden files, set this option to an empty.
-``ignore_path`` (*Default: */.*):
+      multiple lines. 
+      The default is .** which ignores all hidden files on a POSIX filesystem. 
+      To include hidden files, set this option to an empty list ``[]``.
+``ignore_path`` (*Default: [*/.\*]*):
       List of ``glob`` patterns which match directories to ignore when
       scanning input. This is an array, so multiple patterns can be
-      specified on multiple lines. The default is ***/.** which ignores all
-      hidden directories on a POSIX filesystem. To include hidden directories,
-      set this option to an empty.
+      specified on multiple lines. 
+      The default is ***/.** which ignores all hidden directories on a POSIX filesystem. 
+      To include hidden directories, set this option to an empty list ``[]``.
 
 
 Archiving
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -276,18 +276,11 @@ section for details.
             Enforce restart file reproducibility.
 
 ``ignore`` (*Default: [.\*]*):
-      List of ``glob`` patterns which match files to ignore when scanning input
-      directories. This is an array, so multiple patterns can be specified on
-      multiple lines. 
-      The default is .** which ignores all hidden files on a POSIX filesystem. 
-      To include hidden files, set this option to an empty list ``[]``.
-``ignore_path`` (*Default: [*/.\*]*):
-      List of ``glob`` patterns which match directories to ignore when
-      scanning input. This is an array, so multiple patterns can be
-      specified on multiple lines. 
-      The default is ***/.** which ignores all hidden directories on a POSIX filesystem. 
-      To include hidden directories, set this option to an empty list ``[]``.
-
+      List of ``glob`` patterns which match files to ignore when scanning input directories.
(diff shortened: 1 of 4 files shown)


docs/source/config.rst
```diff
@@ -275,11 +275,17 @@ section for details.
       ``restart`` (*Default: global reproduce flag*)
             Enforce restart file reproducibility.
 
-``ignore`` (*Default: .\**):
+``ignore`` (*Default: .*):
```
jo-basevi: I think it needs the `\` so the displayed string is  `.*`. 

The docs preview (found at the CI checks at the bottom of the PR): https://payu--677.org.readthedocs.build/en/677/config.html#manifests displays  `(Default: .):`

jo-basevi (commented):
Just have been thinking more about how similar `ignore` and `ignore_path` config options are..

So `ignore` is a list of patterns for filenames, and this PR adds `ignore_path` to have glob patterns values  for paths (so we can exclude files that are in hidden directories by default). 

I initially didn't want to change the behaviour of the `ignore` option to preserve backwards compatibility (though I don't know how often `ignore` is actually used...). We could add a deprecate notice to `ignore` and suggest the patterns get converted to `ignore_paths`.  Then in future, remove `ignore` as an option. 

Or we could convert automatically convert the `ignore` patterns to `ignore_path` patterns?  The current method uses `fnmatch` which does have limited special characters - https://docs.python.org/3.11/library/fnmatch.html. It would've possible to add `**/` to the  `ignore` patterns using [pathlib.Path.full_match()](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.full_match), but this was only added in Python3.13 (payu needs to still be able to run in python3.11).  

Or there is the [pathlib.Path.match()](https://docs.python.org/3.10/library/pathlib.html#pathlib.PurePath.match) which does matching from the right of the path? So it could possible that all `ignore` patterns could be applied to the full input path and get the same files being excluded (as all old `ignore` patterns would have been relative paths). 
Edit: It'll be difficult to match hidden directories with `pathlib.Path.match()` as `*` isn't recursive, and the recursive `**` isn't supported

payu/manifest.py
```diff
@@ -121,6 +121,16 @@ def check_reproduce(self, previous_manifest):
...
+    def path_full_match(self, fullpath, ignore_patterns):
+        """
+        Recursively check if ignore patterns match filepath. If any pattern matches, return True.
+        """
+        for pattern in ignore_patterns:
+            # Check each part of the filepath against the pattern
+            for p in Path(fullpath).parts:  
+                if Path(p).match(pattern):
+                    return True
+        return False
```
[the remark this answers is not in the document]
aidanheerdegen: It is a very nice and clean implementation.

Strictly speaking it isn't recursive, as that [involves a function calling itself](https://realpython.com/python-recursion/). So more an iterative check with early termination.

docs/source/config.rst
```diff
@@ -275,12 +275,12 @@ section for details.
       ``restart`` (*Default: global reproduce flag*)
             Enforce restart file reproducibility.
 
-``ignore`` (*Default: .\**):
-      List of ``glob`` patterns which match files to ignore when scanning input
-      directories. This is an array, so multiple patterns can be specified on
-      multiple lines. The default is *.\** which ignores all hidden files on a
-      POSIX filesystem.
-
+``ignore`` (*Default: [.\*]*):
```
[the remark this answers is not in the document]
Qian-HuiChen: Thanks for the suggestion! It turns out that backslash is not needed outside italicize.
 ```
``ignore`` (*Default:* ``[.*]``):
```

File: test/test_manifest.py [hunk 4]
```diff
@@ -533,3 +533,23 @@ def test_hard_sweep():
...
+    from payu.manifest import path_full_match
+
+    # Test with no ignore patterns, should return False for all files
+    assert(path_full_match(tmpdir/'visible_file', []) == False)
+    assert(path_full_match(tmpdir/'.hidden_file', []) == False)
+
+    # Test with default pattern ['.*'], should match hidden file and directories
+    assert(path_full_match(tmpdir/'.hidden_file', ['.*']) == True)
+    assert(path_full_match(tmpdir/'visible_file', ['.*']) == False)
+    assert(path_full_match(tmpdir/'.hidden_dir/visible_file', ['.*']) == True)
```
aidanheerdegen: Can we have a test to make sure we're not getting a false positive for the default with a name with a `.` embedded, e.g. `test.nc`.

test/test_manifest.py
```diff
@@ -533,3 +533,23 @@ def test_hard_sweep():
...
+    
+    # Test with custom pattern, should match files and directories with pattern but not those without
+    assert(path_full_match(tmpdir/'pattern_test_004', ['pattern_*']) == True)
+    assert(path_full_match(tmpdir/'test_002', ['pattern_*']) == False)
+    assert(path_full_match(tmpdir/'pattern_dir/test_005', ['pattern_*']) == True)
```
aidanheerdegen: Would be good to have a test with multiple patterns to make sure they work as expected.

aidanheerdegen (changes requested):
Sorry, just a couple more changes!

aidanheerdegen (approved):
I think this is a nice clean implementation. Thank you.

PR #679, opened 2026-02-03
Change from coverall to CodeCov in Github workflow.

### LINKED ISSUE #469: Use CodeCov for CI code coverage
At the moment, the CI has a fix for broken CI coveralls step, but it's having to use a commit from a development branch rather than a tagged version of the action - see #468

An improvement could be to move to using the CodeCov - which is used for other projects: https://github.com/ACCESS-NRI/dev-docs/wiki/Continuous-integration-and-deployment#uploading-code-coverage-reports-to-codecov

Qian-HuiChen:
Use CodeCov instead of coverall for test coverage reporting in CI.
Closes #469

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -89,7 +89,7 @@ jobs:
         run: python -m pytest --cov-report=xml
 
       - name: Upload code coverage report to Codecov
-        uses: codecov/codecov-action@v5
+        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
         with:
           files: ./coverage.xml
           token: ${{ secrets.CODECOV_TOKEN }}


.github/workflows/CI.yml
```diff
@@ -86,22 +86,13 @@ jobs:
         run: pylint --extension-pkg-whitelist=netCDF4 --ignored-modules=backports -E payu
 
       - name: Run tests
-        run: |
-          PYTHONPATH=. pytest --cov=payu test;
+        run: python -m pytest -s --cov=payu --cov-report=xml .
```
atteggiani: Since the `--cov=payu --cov-report=xml` options are added in the `pyproject.toml`'s `[tool.pytest.ini_options]` below, we have the option to remove them here (as they would still be applied). 
```suggestion
        run: python -m pytest -s
```

pyproject.toml
```diff
@@ -68,10 +69,19 @@ requires = [
 include = ["payu*"]
 namespaces = false
 
+[tool.pytest.ini_options]
+addopts = "--cov=payu --cov-report=xml"
+
```
atteggiani:
```suggestion
testpaths = ["test/"]
```

I think this slightly speeds up running tests because instead of internally looking for tests, we provide pytest with test paths already. In the cases when a test path is provided in the manual command (e.g., `pytest ../relative/path/to/test` the command-line path would still take precedence over this configuration option.

atteggiani:
Also, I see there is the `Coveralls Finished` check that is required but never runs (as that job was removed with the changes in this PR). This requirement is probably set within a branch protection rule. I think `Coveralls Finished` should be removed as a branch protection requirement.

Issue #640, opened 2025-10-28
Payu status improvements


jo-basevi:
Just documenting suggested improvements to `payu status` command here:

- Add queue time for queued jobs. 
- For currently running jobs, add in current model time (see discussion here: https://github.com/payu-org/payu/issues/520#issuecomment-2795919247)
- Add job files for collate and sync, and maybe postscripts? (can remove log files from control directory)
- Use symlink paths in payu status output (would need some care to check symlinks are set up)

PR #686, opened 2026-02-11
Display queue time in payu status.

### LINKED ISSUE #684: Add queue time for queued jobs in payu status
In `payu status`, 
show queue time for queued jobs,
show total queue time for running or archived jobs.

This is a sub-issue for part of issue #640

Qian-HuiChen:
For jobs in queue, a queue time will be displayed when `payu status --update` is entered.
For jobs in `model-run`, `complete` or `archive`, a total queue time will be displayed when `payu status` is entered.

This is to solve the first item in #640 .
Closes #684


aidanheerdegen:
> Another option to avoid having to query the scheduler with `qstat`, (with `payu status --update`), could be for payu to store the current time when the job is submitted in the queued job file:

Wouldn't `qstat` have to be run in any case to establish if the job is still queued?

jo-basevi:
> Wouldn't qstat have to be run in any case to establish if the job is still queued?

Yeah true, so would still need to run `payu status --update` for payu jobs that were deleted while on queue to remove any payu job files. 

I normally spam use of `payu status` which doesn't run `qstat`, and only run `payu status --update` rarely to clean up some files. So with this change, the common usage pattern will be to run `payu status --update` more often to get the scheduler information. (Which we might have to do anyway in future to get resource usage updates)

payu/status.py
jo-basevi: So these are the blobs of data get printed  as json when `payu status --json` is run. As most of these fields will live in the `job_file`, I am a bit cautious over time of the size of this data just becoming a ever larger subset of the `job_file` so the important pieces of information just become harder to find.

It would still be possible to obtain the `qtime` and `stime` programmatically by getting the `job_file` and then parsing the scheduler information. E.g. this could be done at a later stage when printing the queue information.

On the other hand `--json` options can have a lot of data anyway and designed to be machine readable, so maybe this actually isn't an issue!

payu/status.py
```diff
@@ -210,10 +250,20 @@ def update_all_job_files(
...
             job_status = all_jobs.get(job_id)
             exit_status = job_status.get("exit_status") if job_status else None
+            
+            update_job_file(
+                file_path=job_file,
+                data={
+                    "qtime": job_status.get("qtime") if job_status else qtime,
+                    "stime": job_status.get("stime") if job_status else stime,
+                }
+            )
```
jo-basevi: I spoke a bit about this yesterday, but `scheduler.get_all_jobs_status` could instead get all jobs information and update the `scheduler_job_info` data. My reasons are that we are running `qstat` command anyway so we have access to the information and it means it'll be easier to display extra scheduler information in future, like `% Walltime Used` or `% Memory Used`. Also both `qtime` and `stime` will exist in the scheduler fields in the `scheduler_job_info`, so adding them in the top level of the data also might be a bit redundant. 

 `scheduler.get_all_jobs_status` could need to be re-written/replaced with a function that returns a dictionary of `job_id` -> `scheduler_job_info` arrays.

jo-basevi (commented):
Looks good @Qian-HuiChen!! Sorry that most of my comments are just comments rather than things that need to be changed in this PR (or at all!), so please feel free to push back on them. 

 Implementing some file-locking or another solution to handle concurrent writes to the job files would be good to add to this if we are updating the job file while payu is running.

payu/status.py
```diff
@@ -104,6 +105,28 @@ def get_job_file_list(
...
+    else:
+        start_time = datetime.strptime(stime, "%a %b %d %H:%M:%S %Y")
+
+    submit_time = datetime.strptime(qtime, "%a %b %d %H:%M:%S %Y")
+    wait_time = (start_time - submit_time).total_seconds()
+
+    # Convert the queue time to a human-readable format
+    wait_hr = int(wait_time // 3600)
+    wait_min = int((wait_time % 3600) // 60)
+    wait_sec = int(wait_time % 60)
```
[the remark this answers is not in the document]
Qian-HuiChen: Thanks! Turns out `divmod` is a built-in function and it makes the code simpler. I end up with:
```
wait_hr, rem = divmod(int(wait_time), 3600)
wait_min, wait_sec = divmod(rem, 60)
```
By converting `wait_time` to an integer first, the output will be integers as well.

payu/status.py
```diff
@@ -262,6 +297,23 @@ def display_job_info(data: dict[str, Any]) -> None:
...
+            job_file = run_info.get("job_file")
+            job_id = run_info.get("job_id")
+            all_job_info = read_job_file(Path(job_file)).get("scheduler_job_info", {})
+            if "qtime" in all_job_info:
+                run_info["qtime"] = all_job_info.get("qtime", None)
+                run_info["stime"] = all_job_info.get("stime", None)
+            else:
+                job_info = all_job_info.get("Jobs", {}).get(job_id, {})
+                run_info["qtime"] = job_info.get("qtime", None)
+                run_info["stime"] = job_info.get("stime", None)
```
```suggestion
            job_info = all_job_info.get("Jobs", {}).get(job_id, {})
            run_info["qtime"] = job_info.get("qtime", None)
            run_info["stime"] = job_info.get("stime", None)
```
This is related to keeping the same data structure comment in pbs.py

payu/status.py
```diff
@@ -211,11 +236,30 @@ def update_all_job_files(
...
             run_status = data.get("exit_status")
 
-            # Get job status from the scheduler data
-            job_status = all_jobs.get(job_id)
-            exit_status = job_status.get("exit_status") if job_status else None
-
-            if not job_status:
+            # Get job info from the scheduler data
+            job_info = all_jobs.get(job_id)
+            exit_status = job_info.get("Exit_status") if job_info else None
```
jo-basevi: If the job_info structure is changed, this will need to be changed to something similar to:

```suggestion
            exit_status = job_info.get("Jobs", {}).get(job_id, {}).get("Exit_status") if job_info else None
```

payu/status.py
```diff
@@ -211,11 +236,30 @@ def update_all_job_files(
...
+                if exit_status is not None:
+                    if stage == "queued":
+                        remove_job_file(file_path=job_file)
+                    elif run_status is None:
+                        update_job_file(
+                            file_path=job_file,
+                            data={
+                                "payu_run_status": exit_status
+                            }
+                        )
```
```suggestion
            if job_info and exit_status is not None and stage == "queued:
               # Job was likely deleted while on the queue and did not start
               remove_job_file(file_path=job_file)
            elif job_info:
               # Add updated scheduler information to the telemetry data
               update_data = {
                        "scheduler_job_info": job_info
                }
                
                if exit_status is not None and run_status is None:
                     # Update the payu status with the overall scheduler job exit status
                     update_data["payu_run_status"] = exit_status
               
                # Write to the job file
                update_job_file(
                    file_path=job_file,
                    data=update_data
                )
```

The above suggestion might not have correct syntax (or even work!), but it'll be good to just write to the job file only once here.

jo-basevi (commented):
Thanks for adding in the updates! I'm happy to chat in person over some of the suggestions

payu/schedulers/pbs.py
```diff
@@ -436,14 +436,11 @@ def get_all_jobs_status(self) -> Optional[Dict[str, Any]]:
...
+        all_job_info = {}
         jobs = info.get('Jobs', {})
         for job_id, job_info in jobs.items():
-            job_statuses[job_id] = {
-                'job_state': job_info.get('job_state'),
-                'exit_status': job_info.get('Exit_status'),
-            }
-        return job_statuses
+            all_job_info[job_id] = job_info
+        return all_job_info
```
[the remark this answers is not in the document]
Qian-HuiChen: Sounds good! It also keeps the structure consistent. Implemented in [020bca0](https://github.com/payu-org/payu/pull/686/commits/020bca065d487c0b35fa76a4a4518ce12d5fc31b)

payu/status.py
Qian-HuiChen: Doneeee in [020bca0](https://github.com/payu-org/payu/pull/686/commits/020bca065d487c0b35fa76a4a4518ce12d5fc31b)

payu/telemetry.py
```diff
@@ -482,9 +485,17 @@ def update_job_file(
...
-    run_info.update(data)
-    atomic_write_file(file_path=file_path, data=run_info)
+    lock = FileLock(str(file_path) + ".lock", lifetime=LOCK_LIFETIME, timeout=LOCK_TIMEOUT)
+    try:
+        with lock:
+            run_info = read_job_file(file_path)
+            run_info.update(data)
+            atomic_write_file(file_path=file_path, data=run_info)
+    except Timeout:
+        run_info = read_job_file(file_path)
```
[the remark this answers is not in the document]
jo-basevi: Yeah happy to go with that. The main thing is that the lock isn't held indefinitely. 

I still think it could be worth saving the "data" to a tmp file, so it can be manually recovered if it's useful, 

I wonder if the print statement for "File at {file_path} is locked.." be a warning instead? As it would be good for this case to be visible.

payu/status.py
```diff
@@ -262,6 +297,16 @@ def display_job_info(data: dict[str, Any]) -> None:
...
+
+            #read out qtime and stime from the job file
+            job_file = run_info.get("job_file")
+            job_id = run_info.get("job_id")
+            all_job_info = read_job_file(Path(job_file)).get("scheduler_job_info", {})
+            job_info = all_job_info.get("Jobs", {}).get(job_id, {})
+            run_info["qtime"] = job_info.get("qtime", None)
+            run_info["stime"] = job_info.get("stime", None)
+
+            display_wait_time(run_info.get("qtime"), run_info.get("stime"))
```
```suggestion
            display_wait_time(job_info.get("qtime", None), job_info.get("stime", None))
```

Feel free to leave this but `qtime` and `stime` probably doesn't need to be saved to the `run_info` dict.

payu/telemetry.py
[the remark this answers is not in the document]
jo-basevi: Thanks!

payu/telemetry.py
```diff
@@ -37,6 +39,8 @@
 }
 
 REQUEST_TIMEOUT = 10
+LOCK_TIMEOUT = 5
+LOCK_LIFETIME = 8
```
jo-basevi:
```suggestion
LOCK_TIMEOUT = 10
LOCK_LIFETIME = 20
```
I think 10 seconds is reasonable amount to wait. Surely, it'll never need 10 seconds

jo-basevi (dismissed):
Thanks for adding all those changes! I added two very minor comments but feel free to merge without adding them in. 

I ran a couple tests with `payu status --update` with telemetry configured to sent to my local desktop, and it seemed to run as expected :)

payu/telemetry.py
Qian-HuiChen: Sounds good 👍

jo-basevi (approved):
Thanks! Looks good to me :)

Issue #688, opened 2026-02-11
Increase timeout limit for `_run_pbsnodes_json`


dougiesquire:
Since #659, I have been getting this error fairly frequently (~10 times today):

```
RuntimeError: Unable to collect pbs node info: command timed out after 10 seconds
```

It could be worth increasing the timeout limit.

Ping @minghangli-uni

aidanheerdegen:
> [@jo-basevi](https://github.com/jo-basevi) I'm thinking of increasing `REQUEST_TIMEOUT` from 10 to 60 for this case. Would you prefer that, or should I introduce a separate variable instead?

I think we need to cache this data. It doesn't change often, so seems unnecessary to query for it every time, and can just fallback to using cached value if we do this check and it times out.

jo-basevi:
> @jo-basevi I'm thinking of increasing REQUEST_TIMEOUT from 10 to 60 for this case. Would you prefer that, or should I introduce a separate variable instead?

Yeah if longer timeouts are needed for this command, then I would prefer a separate variable. 

> I think we need to cache this data. It doesn't change often, so seems unnecessary to query for it every time, and can just fallback to using cached value if we do this check and it times out.

For caching, would this need to be written to a local file so it persists between payu submit calls?

jo-basevi:
> Depends what "once" means. It wouldn't do to store it centrally for all versions of payu as the code may change, or the output of pbsnodes may also change.

Would a path including a payu version work for differing payu versions? E.g. `~/.payu/<payu_version>/pbs/nodes`

minghangli-uni:
This issue seems to be causing some frustration at the moment... Would you mind if I open a quick PR to increase the timeout as an initial fix? I think this change would still make sense alongside the upcoming improvements as well.

jo-basevi:
> Would you mind if I open a quick PR to increase the timeout as an initial fix? 

That would be great thanks! I was just typing out that we should increase the timeout today to prevent issues

jo-basevi:
Yeah in the Repro CI, it'll currently use the default `$HOME/.cache/pbs/` so the cache will carry over between tests. To move away from using the service-user's home environment, we will need to set `PAYU_PBSNODES_CACHE` in the the repro tests CI (added issue: https://github.com/ACCESS-NRI/model-config-tests/issues/196)

PR #691, opened 2026-02-13
Increase pbsnode timeout limit for _run_pbsnodes_json

minghangli-uni:
Ref: https://github.com/payu-org/payu/issues/688#issuecomment-3894747587

Increase the timeout as an initial fix for repro errors: 

`Unable to collect pbs node info: command timed out after 10 seconds`


jo-basevi (approved):
LGTM

minghangli-uni:
Thanks @jo-basevi! I think you may also need to trigger https://github.com/ACCESS-NRI/model-release-condaenv
 after it's merged to get it deployed sooner :)

minghangli-uni:
We can also monitor how people' repros behave with the timeout increased to 60 seconds.

PR #694, opened 2026-02-18
Fix pbsnodes timeout by introducing cached pbsnodes.json file

### LINKED ISSUE #688: Increase timeout limit for `_run_pbsnodes_json`

Qian-HuiChen:
We introduce a caching mechanism for `pbsnodes -a -F json` to prevent timeout errors when the scheduler is busy.

Instead of running the command every time, the output is now saved to a cached `pbsnodes.json` file along with a timestamp. 
If the timestamp is found older than 7 days, `pbsnodes -a -F json` is re-run and the cache file is refreshed.

Fixes #688.

### DIFF
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -26,6 +26,21 @@ from payu.schedulers.scheduler import Scheduler
 
 PBSNODE_TIMEOUT = 60
 
+def get_pbsnodes_cache_path() -> Path:
+    """Get the path of pbsnodes.json cache file, 
+    If not set, then use default cache path."""
+
+    env_path = os.environ.get("PAYU_PBSNODES_CACHE")
+    if env_path:
+        return Path(env_path)
+    else:
+        xdg_cache = os.environ.get('XDG_CACHE_HOME')
+        if xdg_cache:
+            return Path(xdg_cache) / "pbs" / "pbsnodes.json"
+        else:
+            home_dir = os.environ.get("HOME")
+            return Path(home_dir) / ".cache" / "pbs" / "pbsnodes.json"
+
 def check_pbsnode_file(pbsnodes_json_path):
     """ Check if pbsnodes.json file is recent (< 7 days) and rerun pbsnodes if not. """
     expire_day = 7
@@ -48,13 +63,17 @@ def check_pbsnode_file(pbsnodes_json_path):
     # If the file is not recent, rerun pbsnodes and update the file
     if not is_recent:
         new_pbsnodes_json = _run_pbsnodes_json(timeout=PBSNODE_TIMEOUT)
-        new_pbsnodes_json["timestamp"] = time.time()
-        with pbsnodes_json_path.open("w") as f:
-            json.dump(new_pbsnodes_json, f, indent=2)
+        try:
+            with pbsnodes_json_path.open("w") as f:
+                json.dump(new_pbsnodes_json, f, indent=2)
+        except IOError as e:
+            raise RuntimeError(f"Failed to write pbsnodes JSON to {pbsnodes_json_path}") from e
 
-def read_pbsnode_file(archive_path) -> Dict[str, Any]:
+def read_pbsnode_file() -> Dict[str, Any]:
     """ Read pbsnodes.json file and return the parsed JSON data. """ 
-    pbsnodes_json_path = Path(archive_path) / "payu_jobs" / "pbsnodes.json"
+    pbsnodes_json_path = get_pbsnodes_cache_path()
+    pbsnodes_json_path.parent.mkdir(parents=True, exist_ok=True)
+
     check_pbsnode_file(pbsnodes_json_path)
     try:
         with pbsnodes_json_path.open() as f:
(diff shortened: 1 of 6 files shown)


aidanheerdegen (commented):
Preliminary partial review. Haven't looked at the tests yet.

payu/schedulers/pbs.py
```diff
@@ -25,6 +26,44 @@
...
+    # If the file is not recent, rerun pbsnodes and update the file
+    if not is_recent:
+        new_pbsnodes_json = _run_pbsnodes_json(timeout=PBSNODE_TIMEOUT)
+        new_pbsnodes_json["timestamp"] = time.time()
+        with pbsnodes_json_path.open("w") as f:
+            json.dump(new_pbsnodes_json, f, indent=2)
+
+def read_pbsnode_file(archive_path) -> Dict[str, Any]:
+    """ Read pbsnodes.json file and return the parsed JSON data. """ 
+    pbsnodes_json_path = Path(archive_path) / "payu_jobs" / "pbsnodes.json"
```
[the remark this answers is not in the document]
aidanheerdegen: Also the CI jobs need the capability to set an environment variable with a path to a `pbsnodes.json` file as we don't want the CI jobs to rely on the environment of the service user running the jobs.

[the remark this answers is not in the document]
jo-basevi: In python tests, you can use monkey patch to set environment variables, e.g. 
https://github.com/payu-org/payu/blob/0359e0df5d3bce98c68dd7ecf35e3d7720a6837b/test/test_setup.py#L335-L338

For manual testing, just add run `export PAYU_PBSNODES_CACHE=<add-test-path>` in the terminal session.

payu/schedulers/pbs.py
```diff
@@ -48,13 +63,17 @@ def check_pbsnode_file(pbsnodes_json_path):
...
+                json.dump(new_pbsnodes_json, f, indent=2)
+        except IOError as e:
+            raise RuntimeError(f"Failed to write pbsnodes JSON to {pbsnodes_json_path}") from e
 
-def read_pbsnode_file(archive_path) -> Dict[str, Any]:
+def read_pbsnode_file() -> Dict[str, Any]:
     """ Read pbsnodes.json file and return the parsed JSON data. """ 
-    pbsnodes_json_path = Path(archive_path) / "payu_jobs" / "pbsnodes.json"
+    pbsnodes_json_path = get_pbsnodes_cache_path()
+    pbsnodes_json_path.parent.mkdir(parents=True, exist_ok=True)
```
aidanheerdegen: I think it make more sense to put the parent directory creation step at the same place where the json file is dumped out.

aidanheerdegen (changes requested):
Thanks for adding in those changes. Just one or two more small suggestions.

payu/schedulers/pbs.py
```diff
@@ -17,6 +17,7 @@
 
 import json
 from tenacity import retry, stop_after_delay
+import time
```
aidanheerdegen:
```suggestion
from datetime import datetime, timedelta
```

payu/schedulers/pbs.py
```diff
@@ -25,6 +26,63 @@
...
+        if xdg_cache:
+            return Path(xdg_cache) / "pbs" / "pbsnodes.json"
+        else:
+            home_dir = os.environ.get("HOME")
+            return Path(home_dir) / ".cache" / "pbs" / "pbsnodes.json"
+
+def check_pbsnode_file(pbsnodes_json_path):
+    """ Check if pbsnodes.json file is recent (< 7 days) and rerun pbsnodes if not. """
+    expire_day = 7
+    is_recent = False
```
aidanheerdegen:
```suggestion
    recreate_pbsnodes_file = False
```
The logic below checks if the file exists, or if it is too old, so I'd suggest using a variable name that encompasses both cases for clarity. I don't think my suggested name is good, so if you can think of a better one please use it.

payu/schedulers/pbs.py
```diff
@@ -25,6 +26,63 @@
...
+    # Check if pbsnodes.json exists
+    if pbsnodes_json_path.exists():
+        try:
+            # Check if the file is recent (modified within the last 7 days)
+            with pbsnodes_json_path.open() as f:
+                pbsnodes_json = json.load(f)
+            timestamp = pbsnodes_json.get("timestamp", 0)
+
+            if timestamp > time.time() - expire_day*24*60*60:           
+                is_recent = True
```
aidanheerdegen:
```suggestion
            timestamp = datetime.fromtimestamp(pbsnodes_json.get("timestamp", 0))

            if datetime.now() - timestamp > timedelta(days=expire_day):           
                recreate_pbsnodes_file = True
```
It almost certainly takes longer to use the `datetime` methods I am suggesting, but I think they're more readable. 

If you agree use my suggestion. If you don't agree, feel free to ignore it.

aidanheerdegen (commented):
Sorry, a second review as I noticed some stuff I should have just included in the first one.

payu/schedulers/pbs.py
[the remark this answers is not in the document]
Qian-HuiChen: Sounds good. Since we are using filelock anyway, we should use it for pbsnodes file as well. Will implement it in the latest commit

aidanheerdegen (approved):
Looking really good, thanks heaps. 

I have one question, but have approved as I think it is fine to merge if writing a test is too time consuming.

PR #700, opened 2026-03-03
Replace mkdir_p with os.makedirs.

### LINKED ISSUE #460: Replace `mkdir_p` with built-in os.makedirs
As brought up in this PR review (https://github.com/payu-org/payu/pull/452#discussion_r1665678280 by @truth-quark), the mkdir_p() in fsops.py could be replaced by [os.makedirs()](https://docs.python.org/3/library/os.html#os.makedirs) using exist_ok=True.

Qian-HuiChen:
Replace `mkdir_p()` from `fsops.py` with `os.makedirs(path, exist_ok=True)` across all payu code.
All pytests pass and manual run with mom6 succeeds.

Closes #460

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -31,7 +31,7 @@ from packaging import version
 
 # Local
 from payu import envmod
-from payu.fsops import mkdir_p, make_symlink, read_config, movetree
+from payu.fsops import make_symlink, read_config, movetree
 from payu.fsops import list_archive_dirs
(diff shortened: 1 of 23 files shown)


aidanheerdegen (approved):
Good idea to use a standard function that has been available since python 3.2!

The testing coverage drops slightly, but we're removing code, which is good, and I can't think of any useful tests, so that is fine I think.

Thanks for the testing and being explicit what was tested.

PR #702, opened 2026-03-12
Current expt time for running jobs in payu status

### LINKED ISSUE #701: Add current model time for running jobs
In payu status,
show currently experiment time for running jobs.

This is a sub-issue for part of issue https://github.com/payu-org/payu/issues/640

Qian-HuiChen:
In `payu status`, display current experiment time for running jobs, as:
```
  Job ID:            xxx.gadi-pbs
  Run ID:            xxx
  Stage:             model-run
  Total queue time:  0h 1m 4s
  Current Expt Time: 1901-02-20T00:00:00
  ::
```
This is implemented in models:
 - OM2
 - ESM1.6
 - MOM6
 - OM3
For other models, no current expt time will be displayed.

In `payu status`, display current model finish time for archived jobs, e.g.,
 ```
 Job ID:            xxx.gadi-pbs
  Run ID:            xxx
  Stage:             archive
  Total queue time:  0h 0m 22s
  Model Finish Time: 1900-05-01T00:00:00
  ::
```

Closes #701 as part of #640.

### DIFF
--- a/payu/models/access_esm1p6.py
+++ b/payu/models/access_esm1p6.py
@@ -15,7 +15,6 @@ import os
 import re
 import shutil
 import sys
-import warnings
 
 # Extensions
 import f90nml
@@ -320,6 +319,11 @@ class AccessEsm1p6(Model):
         return self.get_restart_datetime_using_submodel(restart_path,
                                                         model_types)
 
+    def get_cur_expt_time(self):
+        """ Use UM submodel to get the current experiment time."""
+        model_types = ['um']
+        return self.get_cur_expt_time_using_submodel(model_types)
+
     def set_model_pathnames(self):
         pass
 
@@ -335,69 +339,6 @@ class AccessEsm1p6(Model):
     def collate(self):
         pass
 
-    def read_start_date(self, nl_path):
-        """Read the start date from the namelist file"""
-        try:
-            namelist = f90nml.read(nl_path)
-            model_basis_time = namelist.get('nlstcall', {}).get('model_basis_time', None)
-            if model_basis_time is None:
-                warnings.warn(f"model_basis_time not found in {nl_path}")
-                return None
-            return datetime(*model_basis_time)
-        except Exception as e:
-            warnings.warn(f"Could not read file {nl_path}: {e}")
-            return None
-
-    def convert_timestep(self, log_path):
-        """ Convert the timestep to experiment runtime 
-        based on the information in log file"""
-        timestep = None
-        step_per_period = 0
-        secs_per_period = 0
-
-        with open(log_path, 'r') as f:
-            for line in f:
-                try:
-                    if 'STEPS_PER_PERIODim' in line:
-                        step_per_period = int(line.split('=')[-1].strip())
-                    if 'SECS_PER_PERIODim' in line:
-                        secs_per_period = int(line.split('=')[-1].strip())
-                        print(f"Found secs_per_period: {secs_per_period}")
-                    if 'Atm_Step: Timestep' in line:
-                        timestep = int(line.split()[-1])
-                except ValueError:
-                    continue
-
-        if timestep is not None and secs_per_period > 0 and step_per_period > 0:
-            secs_per_step = secs_per_period / step_per_period 
-            runtime_sec = timestep * secs_per_step
-            return runtime_sec
-        
-        warnings.warn(
-                f"""Could not find all required entries in file {log_path}
-                to calculate run time"""
-            )
-        return None
-
-
-    def get_cur_expt_time(self):
-        """Get the current experiment time from file"""
-        nl_path = os.path.join(self.expt.work_path, 'atmosphere', 'namelists')
-        log_path = os.path.join(self.expt.work_path, 'atmosphere', 'atm.fort6.pe0')
-
-        if not os.path.exists(nl_path) or not os.path.exists(log_path):
-            warnings.warn(f"Could not find required files: {nl_path} or {log_path}")
-            return None
-        
-        start_date = self.read_start_date(nl_path)
-        runtime_sec = self.convert_timestep(log_path)
-        if start_date is None or runtime_sec is None:
-            return None
-            
-        # Both ESM1.6 and datetime package use Gregorian calendar, 
-        # so we can use datetime here for date calculations.
-        cur_expt_time = start_date + timedelta(seconds=runtime_sec)
-        return cur_expt_time.isoformat()
         
         
         
--- a/payu/models/access_esm1p6.py
+++ b/payu/models/access_esm1p6.py
@@ -338,8 +338,3 @@ class AccessEsm1p6(Model):
 
     def collate(self):
         pass
-
-        
-        
-        
-
--- a/payu/models/access_esm1p6.py
+++ b/payu/models/access_esm1p6.py
@@ -15,6 +15,7 @@ import os
 import re
 import shutil
 import sys
+import warnings
 
 # Extensions
 import f90nml
@@ -333,3 +334,71 @@ class AccessEsm1p6(Model):
 
     def collate(self):
         pass
+
+    def read_start_date(self, nl_path):
+        """Read the start date from the namelist file"""
+        try:
+            namelist = f90nml.read(nl_path)
+            model_basis_time = namelist.get('nlstcall', {}).get('model_basis_time', None)
+            if model_basis_time is None:
+                warnings.warn(f"model_basis_time not found in {nl_path}")
+                return None
+            return datetime(*model_basis_time)
+        except Exception as e:
+            warnings.warn(f"Could not read file {nl_path}: {e}")
+            return None
+
+    def convert_timestep(self, log_path):
+        """ Convert the timestep to experiment runtime 
+        based on the information in log file"""
+        timestep = None
+        step_per_period = 0
+        secs_per_period = 0
+
+        with open(log_path, 'r') as f:
+            for line in f:
+                try:
+                    if 'STEPS_PER_PERIODim' in line:
+                        step_per_period = int(line.split('=')[-1].strip())
+                    if 'SECS_PER_PERIODim' in line:
+                        secs_per_period = int(line.split('=')[-1].strip())
+                        print(f"Found secs_per_period: {secs_per_period}")
+                    if 'Atm_Step: Timestep' in line:
+                        timestep = int(line.split()[-1])
+                except ValueError:
+                    continue
+
+        if timestep is not None and secs_per_period > 0 and step_per_period > 0:
+            secs_per_step = secs_per_period / step_per_period 
+            runtime_sec = timestep * secs_per_step
+            return runtime_sec
+        
+        warnings.warn(
+                f"""Could not find all required entries in file {log_path}
+                to calculate run time"""
+            )
+        return None
+
+
+    def get_cur_expt_time(self):
+        """Get the current experiment time from file"""
+        nl_path = os.path.join(self.expt.work_path, 'atmosphere', 'namelists')
+        log_path = os.path.join(self.expt.work_path, 'atmosphere', 'atm.fort6.pe0')
+
+        if not os.path.exists(nl_path) or not os.path.exists(log_path):
+            warnings.warn(f"Could not find required files: {nl_path} or {log_path}")
+            return None
+        
+        start_date = self.read_start_date(nl_path)
+        runtime_sec = self.convert_timestep(log_path)
+        if start_date is None or runtime_sec is None:
+            return None
+            
+        # Both ESM1.6 and datetime package use Gregorian calendar, 
+        # so we can use datetime here for date calculations.
+        cur_expt_time = start_date + timedelta(seconds=runtime_sec)
+        return cur_expt_time.isoformat()
+        
+        
+        
+
--- a/payu/models/accessom2.py
+++ b/payu/models/accessom2.py
@@ -12,6 +12,8 @@ from __future__ import print_function
 
 import os
 import shutil
+import json
+import warnings
 
 from payu.models.model import Model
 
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

**Comment by Qian-HuiChen** [comment]
Good morning @anton-seaice @blimlim @ezhilsabareesh8, 
We are trying to display current model time (e.g., 1901-05-01T00:00:00) in `payu status`, which could be helpful during a long experiment run. 

It will be highly appreciated if you could provide some feedback from a user perspective: 
- Is this information useful?
- Is the current mode of presentation clear?
- Any other feedback :)

Thanks!

**Comment by jo-basevi** [comment]
Thanks for adding this in! Great to have this implemented for so many model drivers. 

I was testing running this branch with a pre-existing MOM6 configuration, and I am getting a warning when running `payu status`:

```
~/payu_development/payu_fork/payu/models/mom6.py:164: UserWarning: Could not find required files: /scratch/tm70/USER/mom6/work/mom6_double_gyre-expt-d9cb6783/input.nml or /scratch/tm70/USER/mom6/work/mom6_double_gyre-expt-d9cb6783/ocean.stats
```
As the run has completed, the work directory no longer exists so should change the code to only query the current run time only when the model is running. 

The other initial thing I noticed was `payu status --json` also prints warnings before printing a json blob. In this case the output probably should only be json so it can still be machine readable.

**Review by blimlim — COMMENTED** [review]
Thanks @Qian-HuiChen, I've had a look through the changes and I think they're looking really good. Being able to query the run date like this would be very useful.

I've added a few small comments and questions throughout. 

For the ESM1.6 implementation, all the relevant files (`work/atmosphere/namelists`, `work/atmosphere/atm.fort6.pe0`, and the suggested `work/atmosphere/um.res.yaml`) are all used and produced specifically by the UM submodel. Because of this, I'm wondering whether it could make sense to move the implementation of `get_curr_exp_time` from `access_esm1p6.py` to `um.py`, and then access it using the `get_cur_expt_time_using_submodel` method in ESM1.6?

File: payu/telemetry.py [hunk 1]
```diff
@@ -530,6 +531,8 @@ def update_run_job_file(
         run_info.update(extra_info)
     if timings:
         run_info.update(get_timings_isoformat(timings))
+    if cur_expt_time:
+        run_info.update({"cur_expt_time": cur_expt_time})
```
jo-basevi: Probably best to remove this as the up to date dates can be determined in `payu status` rather than updates to the the job files.

payu/models/cesm_cmeps.py
```diff
@@ -422,6 +422,27 @@ def get_components(self):
...
+                with open(log_path, 'r') as f:
+                    for line in reversed(f.readlines()):
+                        if line.startswith(" memory_write: model date"):
+                            cur_expt_time = line.split()[4]
+                            return cur_expt_time
+            
+            warn(f"Log file {log_path} does not exist or does not contain current model time.")
+            return None
+
+        except KeyError as e:
```
jo-basevi: Should this be an IndexError for `line.split()[4]`

jo-basevi (commented):
Just added a couple more comments. 

I wonder if each `get_cur_expt_time()` model driver method could return a `cftime.datetime` object. It would catch if a valid date was parsed from the files correctly, and it would also store calendar information that could be used at a later point

As there may (?) be updates to models and their file contents in future, I wonder if we could add a optional command-line flag to silence warnings? 

Another option to protect payu status erroring out in the future would be to have a higher level exception catch, e.g. something like:
```python
try:
   current_model_time = self.model.get_cur_expt_time()
except Exception:
   warnings.warn("Unexpected error parsing current experiment time")
```

 Could also raise Exceptions for FileNotFound, ValueErrors etc during parsing for current run time, and let those exceptions bubble up and then catch exceptions at a higher level and then display a warning of the error. The benefits of this is that you don't need to return and handle `None` values as much.

payu/models/cesm_cmeps.py
```diff
@@ -422,6 +422,27 @@ def get_components(self):
...
+    def get_cur_expt_time(self):
+        """Get the current experiment time from file work/log/med.log."""
+        try:
+            log_path = os.path.join(self.expt.work_path, 'log', 'med.log')
+            
+            # Read out the latest `cur_exp-datetime` from the log file
+            if os.path.exists(log_path):
+                with open(log_path, 'r') as f:
+                    for line in reversed(f.readlines()):
+                        if line.startswith(" memory_write: model date"):
```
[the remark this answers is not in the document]
Qian-HuiChen: Thanks for checking this in the model code 👍

anton-seaice:
> As there may (?) be updates to models and their file contents in future, I wonder if we could add a optional command-line flag to silence warnings?

I also wondered if the whole change could be configured in config.yaml, e.g. the filename, the search string etc. But payu already has lots of configuration items hard coded in, so it seemed unnecessarily complicated !

payu/telemetry.py
Qian-HuiChen: Sounds good! I have deleted this in the next commit.

payu/models/accessom2.py
```diff
@@ -12,6 +12,11 @@
 
 import os
 import shutil
+import json
+import warnings
+import cftime
+import logging
+logger = logging.getLogger(__name__)
 
 from payu.models.model import Model
```
jo-basevi:
```suggestion
import logging

from payu.models.model import Model

logger = logging.getLogger(__name__)
```

Just a small change to move setting the logger to after the imports. I think this applies to a couple other files

payu/models/accessom2.py
```diff
@@ -94,3 +99,24 @@ def get_restart_datetime(self, restart_path):
...
+        with open(log_path, 'r') as f:
+            for line in reversed(f.readlines()):
+                if 'cur_exp-datetime' in line:
+                    line_json = json.loads(line)
+                    time_str = line_json.get('cur_exp-datetime', None)
+                    if time_str is not None:
+                        return cftime.datetime.strptime(time_str, '%Y-%m-%dT%H:%M:%S')
+
+        logger.debug(f"cur_exp-datetime not found in {log_path}")
+        return None
```
jo-basevi: If catching exceptions higher up, could raise a ValueError here instead?

payu/subcommands/status_cmd.py
```diff
@@ -31,21 +34,18 @@ def runcmd(lab_path, config_path, json_output,
...
         # Determine archive path
-        lab = Laboratory(lab_path)
+        lab = Laboratory(config_path=config_path, lab_path=lab_path)
         warnings.filterwarnings("error", category=MetadataWarning)
         try:
-            metadata = Metadata(Path(lab.archive_path),
-                                config_path=config_path)
-            metadata.setup()
+            expt = Experiment(lab, config_path=config_path)
+            expt.init_models()
```
jo-basevi:
```suggestion
```
No need to run this as init_models() is already run as part of Experiment initialisation.

payu/status.py
```diff
@@ -299,10 +302,24 @@ def display_job_info(data: dict[str, Any]) -> None:
...
-            job_info = all_job_info.get("Jobs", {}).get(job_id, {})
+            all_job_info = read_job_file(Path(job_file))
+            job_info = all_job_info.get("scheduler_job_info", {}).get("Jobs", {}).get(job_id, {})
             display_wait_time(job_info.get("qtime", None), job_info.get("stime", None))
 
+            if run_info.get("stage") == "model-run":
+                cur_expt_time = None
+                try:
+                    cur_expt_time = expt.get_model_cur_expt_time()
+                except (FileNotFoundError, IndexError, OSError, json.JSONDecodeError, UnboundLocalError) as e:
```
jo-basevi: What causes the UnboundLocalError?

jo-basevi (commented):
Thank you for adding in all those changes!  Just added a few more comments.

payu/models/um.py
```diff
@@ -15,6 +15,7 @@
 import os
 import shutil
 import string
+import cftime
```
blimlim: It looks like this one isn't directly being used!

test/models/access-om3/test_access_om3.py
```diff
@@ -581,4 +581,62 @@ def test_collect_restart_files_incorrect_parallel():
...
+
+        cur_expt_time = model.get_cur_expt_time()
+
+        assert cur_expt_time.isoformat() == "1900-01-02T00:00:00"
+
+    teardown_cmeps_config()
+
+
+def test_get_cur_expt_time_no_log(tmp_path):
+    """ Test if get_cur_expt_time returns None if log file is missing. """
```
blimlim:
```suggestion
    """ Test if get_cur_expt_time raises an error if log file is missing. """
```
Wondering if this would this be a bit more accurate for the current test?

test/models/access-om3/test_access_om3.py
```diff
@@ -581,4 +581,62 @@ def test_collect_restart_files_incorrect_parallel():
...
+    cmeps_config(1)
+
+    with cd(ctrldir):
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
+        expt = payu.experiment.Experiment(lab, reproduce=False)
+        model = expt.models[0]
+
+        log_path = os.path.join(model.work_path, "log", "med.log")
+        if os.path.exists(log_path):
+            os.remove(log_path)
```
blimlim: I'm wondering if it would be useful to do something like `assert not(os.path.exists(log_path))`, rather than removing the log path if it exists. 

I'm thinking that if the log path does exist, it would suggest that one of the other tests isn't cleaning up properly, and it could be alerted that something must have gone wrong somewhere.

blimlim (commented):
Thanks @Qian-HuiChen for putting in those changes, they're looking really good! I've just had another read through, and added some small comments

payu/models/access_esm1p6.py
```diff
@@ -333,3 +338,8 @@ def set_model_output_paths(self):
 
     def collate(self):
         pass
+
+        
+        
+        
+
```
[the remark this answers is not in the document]
Qian-HuiChen: oops I didn't clean all that up. Did it in the latest commit

test/models/test_access_om2.py
```diff
@@ -0,0 +1,109 @@
...
+import logging
+
+import payu
+import cftime
+
+from test.common import cd, tmpdir, ctrldir, labdir, workdir, write_config, config_path
+from test.common import config as config_orig
+from test.common import make_inputs, make_exe
+from test.common import list_expt_archive_dirs, make_expt_archive_dir, remove_expt_archive_dirs
+
```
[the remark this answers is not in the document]
Qian-HuiChen: Good catch! Thanks :)

test/models/test_access_om2.py
```diff
@@ -0,0 +1,109 @@
...
+    config['model'] = MODEL
+    config['ncpus'] = ncpu
+
+    write_config(config)
+
+def teardown_config():
+    # Teardown
+    os.remove(config_path)
+
+def test_get_cur_expt_time(tmp_path):
```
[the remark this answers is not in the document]
Qian-HuiChen: Yeah, `tmp_path` is not used here. I have removed them in OM3 and OM2 tests now.

test/models/test_mom6.py
```diff
@@ -4,6 +4,7 @@
 
 import pytest
 import f90nml
+import logging
```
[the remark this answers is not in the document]
Qian-HuiChen: Good catch! Thanks :)

payu/models/accessom2.py
```diff
@@ -12,9 +12,14 @@
...
 import os
 import shutil
+import json
+import warnings
+import cftime
+import logging
 
 from payu.models.model import Model
 
+logger = logging.getLogger(__name__)
```
jo-basevi:
```suggestion

from payu.models.model import Model
```

I think these are safe to remove

test/models/test_mom6.py
```diff
@@ -310,3 +310,158 @@ def test_setup():
...
+    
+    # Write a calendar and current model time into ocean_solo.res
+    ocean_solo_path = os.path.join(expt.work_path, 'INPUT', 'ocean_solo.res')
+    os.makedirs(os.path.dirname(ocean_solo_path), exist_ok=True)
+    with open(ocean_solo_path, 'w') as f:
+        f.write("2\n")  # Use Julian calendar
+
+    with pytest.raises(ValueError, match=f"Key 'date_init' not found in {input_path}"):
+        cur_expt_time = expt.get_model_cur_expt_time()
+        assert cur_expt_time is None
```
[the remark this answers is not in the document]
jo-basevi: As this function is just testing `read_start_date`, instead of `expt.get_model_cur_expt_time()` could call the model method directly - might be `expt.model.read_start_date()`?

payu/status.py
```diff
@@ -197,6 +202,16 @@ def build_job_info(
         for run_num, run_jobs in status_data["runs"].items():
             run_jobs["run"] = [run_jobs["run"][-1]]
 
+    if run_info.get("stage") == "model-run":
```
jo-basevi: We only want to modify the `run_info` the for latest run (as that should be the only run that is running the model..), while currently it's modifying the last `run_info`  from the for loop - which might not always be ordered. With `payu status --all`, it can display all run summaries including failed runs. 

Could do something like
```
latest_run_num = max(status_data["runs"])
latest_run_info = status_data["runs"][latest_run_num]["run"][-1]
```

jo-basevi (commented):
Thanks for the chat yesterday and adding in those updates! I've couple more suggestions, but most of them are very minor.

payu/models/cesm_cmeps.py
```diff
@@ -16,12 +16,15 @@
...
 from warnings import warn
+import logging
 
 from payu.fsops import make_symlink
 from payu.models.model import Model
 from payu.models.fms import fms_collate
 from payu.models.mom6 import mom6_add_parameter_files, mom6_save_docs_files
 
+logger = logging.getLogger(__name__)
+
```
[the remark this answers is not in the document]
Qian-HuiChen: Thanks for spotting this!

test/models/test_access.py
```diff
@@ -461,3 +459,64 @@ def test_access_get_um_restart_datetime(um_only_config, remove_restart_dirs):
     restart_path = list_expt_archive_dirs()[0]
     parsed_run_dt = expt.model.get_restart_datetime(restart_path)
     assert parsed_run_dt == date
+
+def test_get_cur_expt_time(um_only_config):
+    """
+    Check that a debug message is logged when trying to get the current experiment time
+    for the access model, as this functionality is not yet implemented.
```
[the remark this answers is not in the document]
Qian-HuiChen: Thanks for pointing this out!

jo-basevi (approved):
Thanks for all the work @Qian-HuiChen! Changes good to me :)
I've run a few short runs with OM2 and OM3 with latest changes and it works as expected (thanks for running all the other models!)

anton-seaice (commented):
I didn't retest this, but the OM2/OM3 specific changes look good

blimlim (approved):
Thanks @Qian-HuiChen! I tested again with esm1.5 and 1.6 and had a look over the changes. Everything looks good to me!

PR #707, opened 2026-03-29
`Payu setup -r` handle empty manifest

### LINKED ISSUE #704: Handle empty or missing manifests in `payu setup --repro`
When running payu setup --repro, payu raises an error when restart manifests are empty, e.g. ACCESS-OM2:
```
 Failed: Failed to run payu setup with --reproduce. Error: Restart manifest must exist and be populated if reproduce is configured to True
```
It should not error here and only error if there was no pre-existing manifest and a new restart manifest was generated. 
Related to this [payu setup check issue](https://github.com/ACCESS-NRI/model-config-tests/pull/190).

Qian-HuiChen:
Closes #704 as part of this [payu setup repro check issue](https://github.com/ACCESS-NRI/model-config-tests/pull/190).

When manifest is empty (e.g., OM2 initial experiment), `payu setup -r` does not error immediately. It only errors when a new restart manifests file is created.

### DIFF
--- a/payu/manifest.py
+++ b/payu/manifest.py
@@ -333,12 +333,6 @@ class Manifest(object):
             self.manifests[mf].calculate_fast(self.previous_manifests[mf])
 
             if self.reproduce[mf]:
-                # When manifest is initially empty, check if it stays empty
-                if len(self.previous_manifests[mf]) == 0 and len(self.manifests[mf]) > 0:
-                    sys.stderr.write(
-                        f'Run cannot reproduce: {mf.capitalize()} manifest was empty but now has entries\n'
-                    )
-                    sys.exit(1)
                 # Compare manifest with previous manifest
                 self.manifests[mf].check_reproduce(self.previous_manifests[mf])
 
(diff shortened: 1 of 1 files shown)


payu/manifest.py
```diff
@@ -341,6 +333,12 @@ def check_manifests(self):
...
 
             if self.reproduce[mf]:
+                # When manifest is initially empty, check if it stays empty
+                if len(self.previous_manifests[mf]) == 0 and len(self.manifests[mf]) > 0:
+                    sys.stderr.write(
+                        f'Run cannot reproduce: {mf.capitalize()} manifest was empty but now has entries\n'
+                    )
+                    sys.exit(1)
                 # Compare manifest with previous manifest
                 self.manifests[mf].check_reproduce(self.previous_manifests[mf])
```
jo-basevi: Does this function not error out when the manifests were empty but now has entries? 

If possible, it might be good to write a test for this case. Could create some inputs/file, but pass an empty manifest?

jo-basevi (dismissed):
Thanks for adding this fix in! I've just had a super quick review

jo-basevi (approved):
Looks good to me

Issue #673, opened 2026-01-23
Handling errors and warnings in Payu


jo-basevi:
It would be good to develop a consistent approach to logging, handling warnings and errors in Payu. Currently for errors, there's a range of:

- `sys.exit('payu: error: ...')`
- Standard exceptions: e.g. `raise RuntimeError(...)`, `raise FileNotFoundError(...)`
- Custom exceptions: `raise PayuBranchError`

For warnings:
- `print('payu: warning: ...')`
- `warnings.warn(...)`

A benefit of using exceptions is that it is easy for them to be caught and handled. This is a large part why using `sys.exit()` is discouraged in library code. On the other hand, exceptions with stack traces (and warnings with code line numbers), aren't the most user friendly and it is often showing information that is not required. Though, especially in development, there are cases where it is useful to have stack traces.

An approach could be to catch exceptions in the main CLI entry point code, and then log the error messages and remove the stack trace, and have a clean exit with `sys.exit(1)`. Similarly, the warnings could be configured in the main CLI entry point code to format and log the warnings. We could add a "debug" option flag to the CLI (and/or `config.yaml`) to include debug logs with stack traces.

A similar approach could be to do above - so keep any global configuration or wrappers around the CLI entry point code, but have custom `PayuError` type (with more specific sub-classes like `PayuConfigError` or `PayuFileNotFound`), then only these exceptions will have a clean error message with no stack trace, and all other exceptions keep their stack trace as before. This will have the benefit that we know the customs exceptions should have a descriptive error message.

Some references:
- An explanation of configuring global logging near the CLI entry point of code: https://realpython.com/ref/best-practices/logging/
- Blog post on error handling which goes into high level catching exceptions with a sys.exit(): https://blog.miguelgrinberg.com/post/the-ultimate-guide-to-error-handling-in-python
- An explanation on why it's difficult to handle system exits: https://stackoverflow.com/a/48571382
- Python docs on when to use logging: https://docs.python.org/3/howto/logging.html#when-to-use-logging

PR #709, opened 2026-03-31
Payu general command wrapper for warning and exceptions formats

jo-basevi:
This PR:
- Removes `payu-branch`, `payu-clone`, `payu-checkout` runscript commands (I think I mistakingly added these..)
- Adds a wrapper script for running the `runscript` commands (e.g. `payu-run`, `payu-sync`, etc) and moved the parsing logic out of the subcommand modules. 
- Add another wrapper function to format logs, warnings, and catch exceptions. This is used by above wrapper and the general  parse() method (used by `payu setup`, `payu run`)
- Add `--stacktrace` flag for enabling stacktraces with exceptions and warnings (`--verbose` name for a flag was already taken by another payu command)
- Add `--log-level` flag for enabling different level of logs.

I've only done a couple tests so far - which are mostly just manually adding random `logger.debug`, `raise ValueError` and `warnings.warn()` lines to methods to trigger errors. 

Related to #673

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -35,8 +35,7 @@ warnings.formatwarning = (
     lambda message, category, filename, lineno, line=None: (
         formatwarning_orig(message, category, filename, lineno, line='')
     )
-)
-
+)    
 
 def parse():
     """Parse the command line inputs and execute the subcommand."""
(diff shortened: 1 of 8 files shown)


payu/cli.py
```diff
@@ -24,23 +25,47 @@
...
+            lambda message, category, filename, lineno, line=None: str(message)
+        )
+
+    try:
+        func(*args, **kwargs)
+    except SystemExit:
+        # TODO: We want to remove any sys.exit(1) calls in the code
+        # and replace with exceptions! 
+        raise
+    except Exception as exc:
```
jo-basevi, line 59: @KAUR1984 could catch the custom exceptions here? 

```python
except PayuError as exc:
  # Format payu exceptions to remove stacktrace by default
  if stacktrace:
      logger.exception(str(exc), exc_info=True)
  else:
      logger.error(str(exc))
except Exception: 
   # Unhandled exceptions - show stacktrace
   logger.exception(str(exc), exc_info=True)
  ```

jo-basevi:
Closing as similar changes are added in #750

PR #713, opened 2026-04-09
Pass warning to logger and add a stacktrace flag.

Qian-HuiChen:
With this new PR, all warnings in payu will be captured and passed through the logging system. By default, only warning messages are displayed.
When `--stacktrace` flag is enabled, the full warning details (including source, line number and message) will be shown.

Related to #673 .

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -41,8 +41,9 @@ def parse():
     parser = generate_parser(is_interactive = True)
 
     # filter out --stacktrace when counting argument numbers
-    filtered_args = [arg for arg in sys.argv if arg != '--stacktrace']
-    arg_count = len(filtered_args)
+    arg_count = len(sys.argv)
+    if '--stacktrace' in sys.argv:
+        arg_count = arg_count - 1
     # Display help if no arguments are provided
     if arg_count == 1:
         parser.print_help()
@@ -52,6 +53,7 @@ def parse():
     args = vars(parser.parse_args())
     run_cmd = args.pop('run_cmd')
 
+    # We pop --stacktrace here so it will not be propagated to runcmd() in subcommands
     stacktrace = args.pop('stacktrace')
     if not stacktrace:
         # Force warnings.warn() to omit the source code line in the message
@@ -85,6 +87,8 @@ def generate_parser(is_interactive=False):
     for cmd in subcmds:
         cmd_parser = subparsers.add_parser(cmd.title, **cmd.parameters)
         cmd_parser.set_defaults(run_cmd=cmd.runcmd)
+        # Add the stacktrace option to all subcommands for consitent CLI UX.
+        # It will be extracted in the parse() and not propagated to subcommand's runcmd()
         cmd_parser.add_argument(*arg_templates.stacktrace['flags'], **arg_templates.stacktrace['parameters'])
 
         for arg in cmd.arguments:
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -33,7 +33,6 @@ DEFAULT_CONFIG = 'config.yaml'
 
 # Pass the warning through the logger
 logging.captureWarnings(True)
-original_formatwarning = warnings.formatwarning
 
 def parse():
     """Parse the command line inputs and execute the subcommand."""
@@ -60,8 +59,6 @@ def parse():
         warnings.formatwarning = (
             lambda message, category, filename, lineno, line=None: f"{message}"
         )
-    else:
-        warnings.formatwarning = original_formatwarning
         
     run_cmd(**args)
 
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -16,6 +16,7 @@ import shlex
 import subprocess
 import sys
 import warnings
+import logging
 
 # Local imports
 import payu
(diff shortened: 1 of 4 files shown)


payu/cli.py
```diff
@@ -25,31 +26,41 @@
...
     run_cmd = args.pop('run_cmd')
+
+    stacktrace = args.pop('stacktrace')
+    if not stacktrace:
+        # Force warnings.warn() to omit the source code line in the message
+        warnings.formatwarning = (
+            lambda message, category, filename, lineno, line=None: f"{message}"
+        )
+    else:
+        warnings.formatwarning = original_formatwarning
```
aidanheerdegen: Is this `else` condition required? `warnings.formatwarning` wasn't changed from the default, so it shouldn't need to be reinstated?

payu/cli.py
```diff
@@ -25,31 +26,41 @@
...
+original_formatwarning = warnings.formatwarning
 
 def parse():
     """Parse the command line inputs and execute the subcommand."""
     setup_logger()
     parser = generate_parser(is_interactive = True)
+
+    # filter out --stacktrace when counting argument numbers
+    filtered_args = [arg for arg in sys.argv if arg != '--stacktrace']
+    arg_count = len(filtered_args)
```
aidanheerdegen:
```suggestion
    arg_count = len([arg for arg in sys.argv if arg != '--stacktrace'])
```
It doesn't appear that `filtered_args` is used again, so it isn't necessary to have a temporary variable.

Another option would be to use `in` and just decrement the variable if it is present:
```python
    arg_count = len(sys.argv)
    arg_count = arg_count - 1 if '--stacktrace' in sys.argv
```
I think that is more readable, and though it won't make any meaningful difference to the execution time, it is more "efficient" as the `in` operation should short-circuit as soon as it finds a matching element.

It is possible to do this in one line as a boolean is evaluated as `True=1` and `False=0`
```python
 arg_count = arg_count - ('--stacktrace' in sys.argv)
```
but it's not particularly readable so I wouldn't recommend it.

payu/cli.py
```diff
@@ -25,31 +26,41 @@
...
+    if arg_count == 1:
         parser.print_help()
         return
-    if len(sys.argv) > 2:
+    if arg_count > 2:
         parser = generate_parser()
     args = vars(parser.parse_args())
     run_cmd = args.pop('run_cmd')
+
+    stacktrace = args.pop('stacktrace')
```
aidanheerdegen, line 56: It would be good to have a short comment here to explain that `--stacktrace` is treated differently to other arguments, and is not propagated to sub-commands.

payu/cli.py
```diff
@@ -74,6 +85,7 @@ def generate_parser(is_interactive=False):
     for cmd in subcmds:
         cmd_parser = subparsers.add_parser(cmd.title, **cmd.parameters)
         cmd_parser.set_defaults(run_cmd=cmd.runcmd)
+        cmd_parser.add_argument(*arg_templates.stacktrace['flags'], **arg_templates.stacktrace['parameters'])
```
aidanheerdegen, line 89: Is it worth adding a code comment here why this is necessary, and refer to back to the comment above. e.g something like "It is necessary to add --stacktrace as a valid option for all subcommands but it is captured in the parse routine above.

Assuming I have understood this correctly!

test/test_cli.py
```diff
@@ -364,4 +367,43 @@ def test_parse_clone(parser):
...
+        payu.cli.parse()
+
+    w = caught[0]
+    formatted = warnings.formatwarning(
+        w.message, w.category, w.filename, w.lineno, w.line
+    )
+    assert formatted == "Test Warning"  
+    assert "UserWarning" not in formatted
+    assert str(w.filename) not in formatted
+    assert str(w.lineno) not in formatted
```
aidanheerdegen: Are these tests necessary? The first assertion tests that `formatted` is a specific string which clearly does not the other elements.

test/test_cli.py
aidanheerdegen, line 1: Is it worth adding a test that the `arg_count` logic is working as expected? e.g. calling `payu --stacktrace` will produce a help message, but `payu list --stacktrace` does not?

aidanheerdegen:
Oh and docs of course! Always forget docs ... or will that be addressed in a later PR?

payu/cli.py
Qian-HuiChen: I found that unit test `test_parse_setup_stacktrace_on` fail if it run right after `test_parse_setup_stacktrace_off`,  but it succeeds if I run it separately. 😕 
So I added this to reset the behavior every time.

test/test_cli.py
Qian-HuiChen: Sounds reasonable. I added a test `test_parse_arg_count` in the latest commit.

test/test_cli.py
Qian-HuiChen: I was thinking more tests are fine. But yes, `formatted == "Test Warning"` already confirm all these three lines will pass. I clean this up in the latest commit.

payu/cli.py
Qian-HuiChen: Sounds good. I added a comment in the latest commit.

payu/cli.py
Qian-HuiChen: Sounds good.  I added a comment in the latest commit.

payu/cli.py
Qian-HuiChen: Sounds good to simplify the code with `in`. I implemented this in the latest commit.

Qian-HuiChen:
Thanks @aidanheerdegen for catching the outdated documentation! 
Based on the big plan, there will be an other flag `--log-level` to be implemented. This is expected to start after this PR is merged, so we should not need to worry about the cross-file conflict for now.

payu/cli.py
aidanheerdegen: This is a problem if the `monkeypatch` changes persist between tests. All tests should be completely independent. Consequently I think it is best to remove this logic and fix the tests.

This [SO thread](https://stackoverflow.com/questions/69940759/monkeypatch-in-python-leaking-mocks-to-other-tests-causing-them-to-fail) suggests `monkeypatch` changes have a default function scope, should shouldn't pollute other tests.

Can you look into why this might be happening?

test/test_cli.py
aidanheerdegen: Thanks. I think it can be misleading/confusing if there are clearly redundant tests. I would think I'd misunderstood the logic.

payu/cli.py
Qian-HuiChen: Thanks for pointing this out! I agree that all tests should be completely independent. In the latest commit, I add
```
monkeypatch.setattr(warnings, "formatwarning", warnings.formatwarning)
```
to ensure the `warnings.formatting` setting is cleaned up when a test ends.

aidanheerdegen (approved):
Looks good, thanks for incorporating the feedback.

jo-basevi:
Hi @Qian-HuiChen, just confirming that this only works for `payu <subcommand>`, e.g. `payu setup`, `payu clone`, but not `payu-subcommand` jobs? Currently it'll display warnings for the qsub generation part of `payu run`, but it won't format warnings inside of PBS job (which runs the `payu-run` command). 

I think to parse the stacktrace (and later, the log level information) to the PBS jobs, we will might need to use environment variables (as briefly described here https://github.com/payu-org/payu/pull/709#issuecomment-4196032127).

Qian-HuiChen:
Hi @jo-basevi , thanks for pointing this out! Yes, PBS job will still show stacktrace (file name, file path, line number). I will create environment variables for stacktrace and log level in the later PR.

payu/cli.py
```diff
@@ -25,31 +26,40 @@
...
-# Force warnings.warn() to omit the source code line in the message
-formatwarning_orig = warnings.formatwarning
-warnings.formatwarning = (
-    lambda message, category, filename, lineno, line=None: (
-        formatwarning_orig(message, category, filename, lineno, line='')
-    )
-)
-
+# Pass the warning through the logger
+logging.captureWarnings(True)
```
jo-basevi, line 35: This is swallowing warnings for in the PBS jobs as a logger is not configured (PBS jobs run with `payu-*` commands)

Related to this comment: https://github.com/payu-org/payu/pull/740#discussion_r3285074649

PR #722, opened 2026-04-20
Create collate job file; log restart collation into job file

### LINKED ISSUE #438: Log collate operations
It would be good to have some machine-readable log of collate operations so there is a track of how files that were output from the model have been transformed. 

For example restart files that are collated break the link between the `restart.yaml` manifest and the restarts that were used in the model run. If that link could be re-established then collated restarts could be reliably used in place of the original tiled restarts.

For a collate log to be used in this manner it would have to include `md5` hashes for the collated and uncollated files and a relationship between them.

How to store this information in the commit history of the experiment repo is a problem that is shared with https://github.com/payu-org/payu/issues/437

Qian-HuiChen:
This PR introduces a job file for each payu collate PBS task. The job file records key metadata such as job ID, walltime, and requested CPU resources.

It also logs the mapping between MD5 hashes of collated files and their corresponding uncollated tiles. Note that this mapping only records restart collation, excluding output collation.

From a user perspective, a new “Collate Info” section is added to the `payu status` output.
Example `payu status` output:
```
Run: 7
----------------------------------------
  Job ID:            1100000.gadi-pbs
  Run ID:            bbbbbbb
  Stage:             archive
  ::
  Model Exit Code:   0 (Success)
  Output Log:        /home/$USER/access-om2/1deg_jra55_ryf/1deg_jra55_ryf.o1100000
  Error Log:         /home $USER/access-om2/1deg_jra55_ryf/1deg_jra55_ryf.e1100000
  Job File:          /scratch/$USER/access-om2/archive/1deg_jra55_ryf-expt-bbbbbb/payu_jobs/7/run/1100000.gadi-pbs.json
  ------------ Collate Info ------------
  Job ID:            1200000.gadi-pbs
  Stage:             exited
  Output Log:        /home/$USER/access-om2/1deg_jra55_ryf/1deg_jra55_ry_c.o1200000
  Error Log:         /home/$USER/access-om2/1deg_jra55_ryf/1deg_jra55_ry_c.e1200000
  Job File:          /scratch/$USER/access-om2/archive/1deg_jra55_ryf-expt-bbbbbb/payu_jobs/7/collate/1200000.gadi-pbs.json
```

Example payu collate job file:
```
"scheduler_job_id": "1200000.gadi-pbs",
    "scheduler_type": "pbs",
    "stage": "exited",
    "payu_current_run": "7",
::
"collate_mapping": {
        "restart006": {
            "367820a8431bb5fa74558e86e75fa89b": [
                "f6b029333dc65bf372653c01162e34cf",
                "c3761ce69020c21003d3a9688b0f2329"
            ],
            "81ee4afbe9ea803971e37d3bdbb739c9": [
                "f4285fe998595e5db48b2080be0da34f",
                "2558e637bd163ed7f1e52e2b7a35cb19"
                ::
            ]
        }
    },
    "payu_collate_status": 0,
    "timings": {
        "payu_start_time": "2026-04-20T15:21:11.220453",
        "payu_collate_duration_seconds": 21.013621598482132,
        "payu_finish_time": "2026-04-20T15:21:42.335297",
        "payu_total_duration_seconds": 31.114844
    }
}

```
Closes #438.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -414,7 +414,7 @@ This will also collate restart ``K-1`` if ``restart: true`` in the ``collate``
 section of the configuration file.
 
 Alternatively you can directly specify a directory name containing the uncollated files 
-(e.g., ``restart001/ocean``)::
+(e.g., ``archive/restart001/ocean``)::
 
   payu collate -d dir_name
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -413,7 +413,8 @@ collate output from run ``K``, type the following::
 This will also collate restart ``K-1`` if ``restart: true`` in the ``collate``
 section of the configuration file.
 
-Alternatively you can directly specify a directory name::
+Alternatively you can directly specify a directory name containing the uncollated files 
+(e.g., ``restart001/ocean``)::
 
   payu collate -d dir_name
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -167,13 +167,16 @@ class Experiment(object):
 
     def set_job_file(self, type='run'):
         """Set the job file for the payu job"""
-        self.job_file = telemetry.get_job_file_path(
+        job_file = telemetry.get_job_file_path(
             archive_path=Path(self.archive_path),
             run_number=self.counter,
             timings=self.timings,
             scheduler=self.scheduler,
             type=type,
         )
+        if type == 'run':
+            self.job_file = job_file
+        return job_file
 
 
     def init_timings(self):
@@ -218,7 +221,7 @@ class Experiment(object):
         else:
             self.model = None
 
-    def set_counters(self):
+    def set_counters(self, keep_run_number=False):
         # Assume that ``set_paths`` has already been called
         assert self.archive_path
 
@@ -235,13 +238,19 @@ class Experiment(object):
             # Check for restart index
             max_restart_index = self.max_output_index(output_type="restart")
             if max_restart_index is not None:
-                self.counter = 1 + max_restart_index
+                if keep_run_number:
+                    self.counter = int(max_restart_index)
+                else:
+                    self.counter = 1 + max_restart_index
             else:
                 # Now look for output directories,
                 # as repeat runs do not generate restart files.
                 max_output_index = self.max_output_index()
                 if max_output_index is not None:
-                    self.counter = 1 + max_output_index
+                    if keep_run_number:
+                        self.counter = int(max_output_index)
+                    else:
+                        self.counter = 1 + max_output_index
                 else:
                     self.counter = 0
 
@@ -968,9 +977,21 @@ class Experiment(object):
         """ Run model collation and record the time taken in seconds to run collation"""
         # Setup modules - load user-defined modules
         self.setup_modules()
+        full_mapping_collate_dict = {}
 
         for model in self.models:
-            model.collate()
+            mapping_collate_dict = model.collate()
+            if mapping_collate_dict is not None:
+                full_mapping_collate_dict.update(mapping_collate_dict)
+
+        # Write the full_mapping_collate_dict to job file in
+        # archive/payu_jobs/{latest_run_number}/collate/{job_id}-gadi-pbs.json
+        self.set_counters(keep_run_number=True)
+        job_file_path = self.set_job_file(type='collate')
+        telemetry.update_job_file(
+            file_path=job_file_path,
+            data={"collate_mapping": full_mapping_collate_dict}
+        )
 
     def profile(self):
         for model in self.models:
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -963,7 +963,9 @@ class Experiment(object):
         if not collating:
             self.postprocess()
 
+    @timeit("payu_collate_duration_seconds")
     def collate(self):
+        """ Run model collation and record the time taken in seconds to run collation"""
         # Setup modules - load user-defined modules
         self.setup_modules()
 
@@ -1010,6 +1012,7 @@ class Experiment(object):
 
             sp.check_call(shlex.split(cmd))
 
+    @timeit("payu_sync_duration_seconds")
     def sync(self):
         # RUN any user scripts before syncing archive
         envmod.setup()
@@ -1064,6 +1067,7 @@ class Experiment(object):
         default_job_name = os.path.basename(os.getcwd())
         short_job_name = str(self.config.get('jobname', default_job_name))[:15]
 
+        # find all PBS log files including payu runs, collate runs, postscript runs, and sync runs
         log_filenames = [short_job_name + '.o', short_job_name + '.e']
         for postfix in ['_c.o', '_c.e', '_p.o', '_p.e', '_s.o', '_s.e']:
             log_filenames.append(short_job_name[:13] + postfix)
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -977,8 +977,12 @@ class Experiment(object):
         """ Run model collation and record the time taken in seconds to run collation"""
         # Setup modules - load user-defined modules
         self.setup_modules()
-        full_mapping_collate_dict = {}
 
+        # Counters does not increase for collate runs
+        # This is used to determine the job file path
+        self.set_counters(keep_run_number=True)
+
+        full_mapping_collate_dict = {}
         for model in self.models:
             mapping_collate_dict = model.collate()
             if mapping_collate_dict is not None:
@@ -986,7 +990,7 @@ class Experiment(object):
 
         # Write the full_mapping_collate_dict to job file in
         # archive/payu_jobs/{latest_run_number}/collate/{job_id}-gadi-pbs.json
-        self.set_counters(keep_run_number=True)
+        
         job_file_path = self.set_job_file(type='collate')
         telemetry.update_job_file(
             file_path=job_file_path,
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -164,19 +164,19 @@ class Experiment(object):
         self.scheduler = scheduler_index[self.scheduler_name]()
         self.job_file = None
 
-
-    def set_job_file(self, type='run'):
-        """Set the job file for the payu job"""
-        job_file = telemetry.get_job_file_path(
+    def get_job_file(self, type='run'):
+        """ Get the job file for the payu job."""
+        return telemetry.get_job_file_path(
             archive_path=Path(self.archive_path),
             run_number=self.counter,
             timings=self.timings,
             scheduler=self.scheduler,
             type=type,
         )
-        if type == 'run':
-            self.job_file = job_file
-        return job_file
+
+    def set_job_file(self):
+        """Set the job file for the payu job"""
+        self.job_file = self.get_job_file(type='run')
 
 
     def init_timings(self):
@@ -977,19 +977,21 @@ class Experiment(object):
         """ Run model collation and record the time taken in seconds to run collation"""
         # Setup modules - load user-defined modules
         self.setup_modules()
-        full_mapping_collate_dict = {}
 
+        full_mapping_collate_dict = {}
         for model in self.models:
             mapping_collate_dict = model.collate()
             if mapping_collate_dict is not None:
                 full_mapping_collate_dict.update(mapping_collate_dict)
 
+        # Counters does not increase for collate runs
+        # This is used to determine the job file path
+        self.set_counters(keep_run_number=True)
+
         # Write the full_mapping_collate_dict to job file in
         # archive/payu_jobs/{latest_run_number}/collate/{job_id}-gadi-pbs.json
-        self.set_counters(keep_run_number=True)
-        job_file_path = self.set_job_file(type='collate')
         telemetry.update_job_file(
-            file_path=job_file_path,
+            file_path=self.get_job_file(type='collate'),
             data={"collate_mapping": full_mapping_collate_dict}
         )
 
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -21,7 +21,6 @@ import tempfile
 import json
 import stat
 import warnings
-import hashlib
 
 # Extensions
 import yaml
@@ -383,15 +382,3 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
             _run_script(cmd, control_path)
         else:
             raise
-
-def calculate_md5_hash(file_path):
-    """ Calculate the md5 hash of a file"""
-    # Make sure the file exists
-    if not os.path.isfile(file_path):
-        raise FileNotFoundError("File not found")
-
-    hash_md5 = hashlib.md5()
-    with open(file_path, "rb") as f:
-        for chunk in iter(lambda: f.read(4096), b""):
-            hash_md5.update(chunk)
-    return hash_md5.hexdigest()
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -21,6 +21,7 @@ import tempfile
 import json
 import stat
 import warnings
+import hashlib
 
 # Extensions
 import yaml
@@ -382,3 +383,15 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
             _run_script(cmd, control_path)
         else:
             raise
+
+def calculate_md5_hash(file_path):
+    """ Calculate the md5 hash of a file"""
+    # Make sure the file exists
+    if not os.path.isfile(file_path):
+        raise FileNotFoundError("File not found")
+
+    hash_md5 = hashlib.md5()
+    with open(file_path, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            hash_md5.update(chunk)
+    return hash_md5.hexdigest()
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -20,9 +20,11 @@ import re
 import warnings
 import json
 
+from yamanifest.hashing import _hashlib
+
 from payu.models.model import Model
 from payu import envmod
-from payu.fsops import required_libs, calculate_md5_hash
+from payu.fsops import required_libs
 from payu.telemetry import get_job_file_path, update_job_file
 
 # There is a limit on the number of command line arguments in a forked
@@ -95,7 +97,7 @@ def get_uncollate_hashes(mnc_tiles, prior_restart_path):
     for nc_fname, tiles in mnc_tiles[prior_restart_path].items():
         # record md5 hashes of all tiles in restart collation
         uncollate_hashes[nc_fname] = [
-            calculate_md5_hash(os.path.join(prior_restart_path, tile)) for tile in tiles
+            _hashlib(os.path.join(prior_restart_path, tile), hashfn='md5') for tile in tiles
         ]
 
     return uncollate_hashes
@@ -122,7 +124,7 @@ def mapping_log(model, mnc_tiles, uncollate_hashes_dict):
     for nc_fname in mnc_tiles[model.prior_restart_path]:
         # calculate md5 hash of the final collated file
         nc_path = os.path.join(model.prior_restart_path, nc_fname)
-        collate_hash = calculate_md5_hash(nc_path)
+        collate_hash = _hashlib(nc_path, hashfn='md5')
 
         # match the collated file hash with the list of uncollated tile hashes
         mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -18,14 +18,12 @@ from itertools import count
 import fnmatch
 import re
 import warnings
-import json
 
-from yamanifest.hashing import _hashlib
+from yamanifest.hashing import _hashlib as hashlib
 
 from payu.models.model import Model
 from payu import envmod
 from payu.fsops import required_libs
-from payu.telemetry import get_job_file_path, update_job_file
 
 # There is a limit on the number of command line arguments in a forked
 # MPI process. This applies only to mppnccombine-fast. The limit is higher
@@ -84,25 +82,27 @@ def get_uncollate_hashes(mnc_tiles, prior_restart_path):
 
     Args:
         mnc_tiles: A dictionary structured as:
-                   {output_path: {collated_filename: [list_of_tile_filenames]},
-                   restart_path: {collated_filename: [list_of_tile_filenames]}}
+                   {restart_path: {collated_filename: [list_of_tile_filenames]}}
+        prior_restart_path: The path to the prior restart directory, e.g., /path/to/restart001.
 
     Returns:
         A dictionary structured as:
         {future_collated_filename_1: [md5_hash_tile_1a, md5_hash_tile_1b, ...],
         ...}
     """
+    if prior_restart_path is None:
+        return {}
+    
     uncollate_hashes = {}
-
     for nc_fname, tiles in mnc_tiles[prior_restart_path].items():
-        # record md5 hashes of all tiles in restart collation
+        # calculate md5 hashes of all tiles in restart collation
         uncollate_hashes[nc_fname] = [
-            _hashlib(os.path.join(prior_restart_path, tile), hashfn='md5') for tile in tiles
+            hashlib(os.path.join(prior_restart_path, tile), hashfn='md5') for tile in tiles
         ]
 
     return uncollate_hashes
 
-def mapping_log(model, mnc_tiles, uncollate_hashes_dict):
+def mapping_log(model, uncollate_hashes_dict):
     """ Log a dictionary of collate mapping into job file.
     Example mapping_collate_dict structure:
     {   
@@ -116,18 +116,18 @@ def mapping_log(model, mnc_tiles, uncollate_hashes_dict):
     if model.prior_restart_path is None:
         return None
     else:
-        prior_restart_num = re.search(r'restart(\d+)', model.prior_restart_path).group()
+        prior_restart_num = os.path.basename(model.prior_restart_path)
 
     mapping_collate_dict = {}
     mapping_collate_dict[prior_restart_num] = {}
 
-    for nc_fname in mnc_tiles[model.prior_restart_path]:
+    for nc_fname, tile_hashes in uncollate_hashes_dict.items():
         # calculate md5 hash of the final collated file
         nc_path = os.path.join(model.prior_restart_path, nc_fname)
-        collate_hash = _hashlib(nc_path, hashfn='md5')
+        collate_hash = hashlib(nc_path, hashfn='md5')
 
         # match the collated file hash with the list of uncollated tile hashes
-        mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
+        mapping_collate_dict[prior_restart_num][collate_hash] = tile_hashes
 
     return mapping_collate_dict
 
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -18,10 +18,12 @@ from itertools import count
 import fnmatch
 import re
 import warnings
+import json
 
 from payu.models.model import Model
 from payu import envmod
-from payu.fsops import required_libs
+from payu.fsops import required_libs, calculate_md5_hash
+from payu.telemetry import get_job_file_path, update_job_file
 
 # There is a limit on the number of command line arguments in a forked
 # MPI process. This applies only to mppnccombine-fast. The limit is higher
@@ -74,6 +76,75 @@ def get_avail_collate_flags(mppnc_path):
 
     return set(re.findall(r"\B-[A-Za-z0-9]+\b", collate_help.stdout))
 
+
+def get_uncollate_hashes(mnc_tiles, prior_restart_path):
+    """ Map future collated filenames to the MD5 hashes of each uncollated tiles.
+
+    Args:
+        mnc_tiles: A dictionary structured as:
+                   {output_path: {collated_filename: [list_of_tile_filenames]},
+                   restart_path: {collated_filename: [list_of_tile_filenames]}}
+
+    Returns:
+        A dictionary structured as:
+        {future_collated_filename_1: [md5_hash_tile_1a, md5_hash_tile_1b, ...],
+        ...}
+    """
+    uncollate_hashes = {}
+
+    for nc_fname, tiles in mnc_tiles[prior_restart_path].items():
+        # record md5 hashes of all tiles in restart collation
+        uncollate_hashes[nc_fname] = [
+            calculate_md5_hash(os.path.join(prior_restart_path, tile)) for tile in tiles
+        ]
+
+    return uncollate_hashes
+
+def mapping_log(model, mnc_tiles, uncollate_hashes_dict):
+    """ Log a dictionary of collate mapping into job file.
+    Example mapping_collate_dict structure:
+    {   
+        "restart001": {
+            "md5_hash_collated_filename_1": ["md5_hash_tile_1a", "md5_hash_tile_1b", ...],
+            "md5_hash_collated_filename_2": ["md5_hash_tile_2a", "md5_hash_tile_2b", ...],
+            ...
+        }
+    }
+    """
+    if model.prior_restart_path is None:
+        return None
+    else:
+        prior_restart_num = re.search(r'restart(\d+)', model.prior_restart_path).group()
+
+    mapping_collate_dict = {}
+    mapping_collate_dict[prior_restart_num] = {}
+
+    for nc_fname in mnc_tiles[model.prior_restart_path]:
+        # calculate md5 hash of the final collated file
+        nc_path = os.path.join(model.prior_restart_path, nc_fname)
+        collate_hash = calculate_md5_hash(nc_path)
+
+        # match the collated file hash with the list of uncollated tile hashes
+        mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
+
+
+    # Write the mapping_collate_dict to job file in
+    # archive/payu_jobs/{current_run}/collate/{job_id}-gadi-pbs.json
+    job_id = os.environ.get('PBS_JOBID', '')
+    current_run = os.environ.get('PAYU_CURRENT_RUN', '')
+
+    job_file_path = get_job_file_path(
+            archive_path=Path(model.expt.archive_path),
+            run_number=model.expt.counter,
+            timings=model.expt.timings,
+            scheduler=model.expt.scheduler,
+            type='collate',
+        )
+    update_job_file(
+        file_path=job_file_path,
+        data={"collate_mapping": mapping_collate_dict}
+    )
+    return mapping_collate_dict
 
 def fms_collate(model):
     """
@@ -202,6 +273,9 @@ def fms_collate(model):
                           .format(len(mnc_tiles[t_base])))
                     print("Warning: collation will be slow and may fail")
 
+    # generate a dictionary of md5 hashes for uncollated tile files
+    uncollate_hashes_dict = get_uncollate_hashes(mnc_tiles, model.prior_restart_path)
+
     cpucount = int(collate_config.get('ncpus',
                    multiprocessing.cpu_count()))
 
@@ -259,6 +333,8 @@ def fms_collate(model):
                 print(op.decode(), file=sys.stderr)
         sys.exit(-1)
 
+    # Get md5 hash for collated files and write collate mapping into job file
+    mapping_log(model, mnc_tiles, uncollate_hashes_dict)
 
 class Fms(Model):
 
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -129,23 +129,6 @@ def mapping_log(model, mnc_tiles, uncollate_hashes_dict):
         # match the collated file hash with the list of uncollated tile hashes
         mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
 
-
-    # Write the mapping_collate_dict to job file in
-    # archive/payu_jobs/{current_run}/collate/{job_id}-gadi-pbs.json
-    job_id = os.environ.get('PBS_JOBID', '')
-    current_run = os.environ.get('PAYU_CURRENT_RUN', '')
-
-    job_file_path = get_job_file_path(
-            archive_path=Path(model.expt.archive_path),
-            run_number=model.expt.counter,
-            timings=model.expt.timings,
-            scheduler=model.expt.scheduler,
-            type='collate',
-        )
-    update_job_file(
-        file_path=job_file_path,
-        data={"collate_mapping": mapping_collate_dict}
-    )
     return mapping_collate_dict
 
 def fms_collate(model):
@@ -336,7 +319,8 @@ def fms_collate(model):
         sys.exit(-1)
 
     # Get md5 hash for collated files and write collate mapping into job file
-    mapping_log(model, mnc_tiles, uncollate_hashes_dict)
+    mapping_collate_dict = mapping_log(model, mnc_tiles, uncollate_hashes_dict)
+    return mapping_collate_dict
 
 class Fms(Model):
 
@@ -367,4 +351,5 @@ class Fms(Model):
         shutil.move(self.work_restart_path, self.restart_path)
 
     def collate(self):
-        fms_collate(self)
+        mapping_collate_dict = fms_collate(self)
+        return mapping_collate_dict
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -168,7 +168,8 @@ def fms_collate(model):
     else:
         mppnc_path = model.expand_executable_path(mppnc_path)
 
-    assert mppnc_path, 'No mppnccombine program found'
+    if not mppnc_path:
+        raise FileNotFoundError('No mppnccombine program found')
 
     # Check config for collate command line options
     collate_flags = collate_config.get('flags')
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -319,7 +319,7 @@ def fms_collate(model):
         sys.exit(-1)
 
     # Get md5 hash for collated files and write collate mapping into job file
-    mapping_collate_dict = mapping_log(model, mnc_tiles, uncollate_hashes_dict)
+    mapping_collate_dict = mapping_log(model, uncollate_hashes_dict)
     return mapping_collate_dict
 
 class Fms(Model):
--- a/payu/models/fms.py
+++ b/payu/models/fms.py
@@ -81,33 +81,49 @@ def get_avail_collate_flags(mppnc_path):
     return set(re.findall(r"\B-[A-Za-z0-9]+\b", collate_help.stdout))
 
 
-def get_uncollate_hashes(mnc_tiles, prior_restart_path):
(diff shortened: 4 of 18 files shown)


payu/subcommands/collate_cmd.py
```diff
@@ -86,7 +90,28 @@ def runcmd(model_type, config_path, init_run, lab_path, dir_path):
 
     pbs_config['qsub_flags'] = ' '.join(qsub_flags)
 
-    cli.submit_job('payu-collate', pbs_config, pbs_vars)
+    # Set up experiment
```
aidanheerdegen: Can you add a little detail to the comment about why it is necessary to instantiate an experiment object at this point. Thanks.

payu/status.py
```diff
@@ -184,6 +184,31 @@ def build_job_info(
...
+            
+            collate_info = {
+                "job_id": collate_data.get("scheduler_job_id"),
+                "stage": collate_data.get("stage"),
+                "exit_status": collate_data.get("payu_collate_status"),
+                "stdout_file": str(stdout) if stdout else None,
+                "stderr_file": str(stderr) if stderr else None,
+                "job_file": str(collate_file),
+            }
+            run_info["collate_info"] = collate_info
```
aidanheerdegen: I like inline statements, but in this case I think it obscures the meaning and it would be better to have a top level `if` statement and put all the rest inside that, as it simplifies the logic, e.g.

```suggestion
        if collate_dir.exists():
            collate_files = sorted(list(collate_dir.glob("*.json")), key=lambda f: f.stat().st_mtime)

            if collate_files:
                collate_data = read_job_file(Path(collate_files[-1]))

                stdout, stderr = find_scheduler_logs(
                    job_id=collate_data.get("scheduler_job_id"),
                    control_path=control_path,
                    archive_path=archive_path,
                    type=collate_data.get("scheduler_type")
                )
            
                collate_info = {
                    "job_id": collate_data.get("scheduler_job_id"),
                    "stage": collate_data.get("stage"),
                    "exit_status": collate_data.get("payu_collate_status"),
                    "stdout_file": str(stdout) if stdout else None,
                    "stderr_file": str(stderr) if stderr else None,
                    "job_file": str(collate_file),
                }
                run_info["collate_info"] = collate_info
```

I have not tested this, and it is assuming I have read the logic correctly.

payu/status.py
```diff
@@ -324,15 +365,20 @@ def display_job_info(data: dict[str, Any]) -> None:
...
-            print_line("Error Log", "stderr_file", run_info)
-            print_line("Job File", "job_file", run_info)
-    print("=" * 40)
+
+            display_status_code(run_info.get("exit_status"), "Exit Status")
+            display_status_code(run_info.get("model_exit_status"), "Model Exit Code")
+            display_log_job_files(run_info)
+
+        # display the collate job information if it exists
+        if run_info.get("collate_info"):
```
aidanheerdegen: Normally would use `in` here I think
```suggestion
        if "collate_info" in run_info:
```

Or use `collate_info = run_info.get("collate_info", None)` and do `if collate_info` and avoid evaluating inside the `if`statement.

payu/status.py
```diff
@@ -324,15 +365,20 @@ def display_job_info(data: dict[str, Any]) -> None:
...
+            print(f"  {'-' * 12} Collate Info {'-' * 12}")
+            collate_info = run_info["collate_info"]
+            print_line("Job ID", "job_id", collate_info)
+            print_line("Stage", "stage", collate_info)
+
+            collate_exit_code = collate_info.get("exit_status")
```
aidanheerdegen: For consistency with usage above. Also I tend to only use `get` if I'm providing a default. Not sure one is better than other but when in doubt consistency is usually good.
```suggestion
            collate_exit_code = collate_info["exit_status"]
```

payu/fsops.py
```diff
@@ -382,3 +383,15 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
...
+    """ Calculate the md5 hash of a file"""
+    # Make sure the file exists
+    if not os.path.isfile(file_path):
+        raise FileNotFoundError("File not found")
+
+    hash_md5 = hashlib.md5()
+    with open(file_path, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            hash_md5.update(chunk)
+    return hash_md5.hexdigest()
```
aidanheerdegen: Given this is (mostly) a copy of the `yamanifest` `_hashlib` function could that just be imported and used?

https://github.com/ACCESS-NRI/yamanifest/blob/7f9aaaddc2d31ebe1cd1b9d92ab2df349e35ba82/yamanifest/hashing.py#L67-L73

Is the file exists function logic required? Won't it just throw a `FileNotFoundError` if it doesn't exist?

payu/status.py
```diff
@@ -297,6 +322,20 @@ def print_line(label: str, key: Any, data: dict[str, Any]) -> None:
     if value is not None and value != "":
         print(f"  {f'{label}:':<{label_width}} {value}")
 
+def display_status_code(status_code: int, label: str) -> None:
+    """Display the status (success/failed) code inside the payu status output"""
+    label_width = 18
+    if status_code is not None:
+        status_str = "Success" if status_code == 0 else "Failed"
+        print(f"  {f'{label}:':<{label_width}} {status_code} ({status_str})")
```
aidanheerdegen: Rather than create a separate function it would be possible to refactor `print_line` to accept a simple string that covered both use cases.

payu/status.py
```diff
@@ -324,15 +365,20 @@ def display_job_info(data: dict[str, Any]) -> None:
...
+            print(f"  {'-' * 12} Collate Info {'-' * 12}")
+            collate_info = run_info["collate_info"]
+            print_line("Job ID", "job_id", collate_info)
+            print_line("Stage", "stage", collate_info)
+
+            collate_exit_code = collate_info.get("exit_status")
+            display_status_code(collate_exit_code, "Exit Status")
+            display_log_job_files(collate_info)
+
+    print("=" * 40)
```
aidanheerdegen: This width is hard-coded in 3 locations. Could be defined as a variable in a single location.

aidanheerdegen (commented):
Looking good. I have some questions/suggestions.

payu/models/fms.py
```diff
@@ -75,6 +77,75 @@ def get_avail_collate_flags(mppnc_path):
...
+        collate_hash = calculate_md5_hash(nc_path)
+
+        # match the collated file hash with the list of uncollated tile hashes
+        mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
+
+
+    # Write the mapping_collate_dict to job file in
+    # archive/payu_jobs/{current_run}/collate/{job_id}-gadi-pbs.json
+    job_id = os.environ.get('PBS_JOBID', '')
+    current_run = os.environ.get('PAYU_CURRENT_RUN', '')
```
jo-basevi: Are these values used anywhere?

payu/models/fms.py
```diff
@@ -75,6 +77,75 @@ def get_avail_collate_flags(mppnc_path):
...
+
+    job_file_path = get_job_file_path(
+            archive_path=Path(model.expt.archive_path),
+            run_number=model.expt.counter,
+            timings=model.expt.timings,
+            scheduler=model.expt.scheduler,
+            type='collate',
+        )
```
jo-basevi: Expt has a method for setting job file

https://github.com/payu-org/payu/blob/00f59e41cdc80e0b369f638bc641e25c7b53479d/payu/experiment.py#L168-L176

This could be re-written to get_job_file(self, type) and have a separate set_job_file for setting the run job file.

payu/models/fms.py
```diff
@@ -75,6 +77,75 @@ def get_avail_collate_flags(mppnc_path):
...
+    update_job_file(
+        file_path=job_file_path,
+        data={"collate_mapping": mapping_collate_dict}
+    )
+    return mapping_collate_dict
```
jo-basevi: Would it be cleaner to return the `mapping_collate_dict` to the experiment.collate() method and then update the telemetry file there? That then keeps the telemetry logic out of the model driver.

payu/subcommands/collate_cmd.py
```diff
@@ -86,7 +90,28 @@ def runcmd(model_type, config_path, init_run, lab_path, dir_path):
...
+            timings=expt.timings,
+            scheduler=expt.scheduler,
+            type='collate',
+        )
+
+    # Write initial job information to job file
+    update_job_file(
+        file_path=job_file_path,
+        data=get_job_info_json(job_id)
+    )
```
jo-basevi: What does this section do?

payu/subcommands/collate_cmd.py
```diff
@@ -108,5 +133,36 @@ def runscript():
...
+        job_file_path = get_job_file_path(
+            archive_path=Path(expt.archive_path),
+            run_number=expt.counter,
+            timings=expt.timings,
+            scheduler=expt.scheduler,
+            type='collate',
+        )
+
+        # Record the collation status (duration time and success/failure) in the job file
+        record_run(
```
jo-basevi, line 136: `record_run` will send telemetry once it's been deployed to `payu/dev`. I'll need to have a think on how to handle that as currently tracking-services will warn that telemetry ingestion has failed.. 

My quick thoughts is that it'll need a way to distinguish between `run` and `collate` (there is `payu_run_status` and `payu_collate_status`) if we send it to the same endpoint. It'll be worth setting up a schema for collate jobs. A copy of the payu run schema is available here:

https://github.com/payu-org/payu/blob/master/test/resources/schema/telemetry/1-0-0.json

payu/status.py
Qian-HuiChen: Sounds good! I define it as `line_width` in the latest commit

payu/status.py
Qian-HuiChen: Keeping consistent sound good. This line is merged into the print_line function in the latest commit.

payu/status.py
Qian-HuiChen: Get it. I will avoid evaluating inside `if` from now on. In the latest commit, it changed to
```
collate_info = run_info.get("collate_info", None)
    if collate_info:
```

payu/status.py
Qian-HuiChen: Sounds good. In the latest commit, `print_line` is refactored as 
```
def print_line(label: str, key: Any, data: dict[str, Any], is_status: bool = False) -> None:
    value = data.get(key)
    label_width = 18
    if value is not None and value != "":
        if is_status:
            status_str = "Success" if value == 0 else "Failed"
            print(f"  {f'{label}:':<{label_width}} {value} ({status_str})")
        else:
            print(f"  {f'{label}:':<{label_width}} {value}")
```

payu/status.py
Qian-HuiChen: Sounds good to add a top level `if` and make the logic more readable. Implemented in the latest commit.

payu/fsops.py
Qian-HuiChen: Thanks for pointing it out! In the latest commit, I removed this function and use `from yamanifest.hashing import _hashlib` instead. After testing, yes, the `FileNotFoundError` will be thrown by `_hashlib` itself so I also removed the file check.

jo-basevi (commented):
The code looks really neat! 

I've tested with `payu run` with a ACCESS-OM2 configuration with collate, and then I ran `payu collate` on it's own. Subsequent `payu status` runs would output have `No run information available.` When running `tree archive/payu_jobs` the output would look like:
```
├── 3
│   ├── collate
│   │   ├── 166832037.gadi-pbs.json
│   │   └── 166832133.gadi-pbs.json
│   └── run
│       └── 166832037.gadi-pbs.json
├── 4
│   └── collate
│       ├── 166834174.gadi-pbs.json
│       └── 20260423110714.json
└── collate
    └── 166834174.gadi-pbs.json
```

payu/subcommands/collate_cmd.py
Qian-HuiChen: Sounds good. I comment a little bit of reason here: # Initialise experiment to determine archive path and run number (which is needed to write job file)

payu/status.py
jo-basevi: Just adding that `collate_info.get("exit_status")` always has a default of None so it won't ever raise a key error. So `collate_info.get("exit_status")` is equivalent to `collate_info.get("exit_status", None)`
https://stackoverflow.com/a/11041421

Qian-HuiChen: Thanks for pointing this out! Good to know there is always a default None :)

payu/models/fms.py
Qian-HuiChen: Oops, I should have deleted this. Removed in the latest commit.

payu/models/fms.py
Qian-HuiChen: I added a parameter (`type`) in `set_job_file` so it can be used for both run and collate. I rewrote this as `job_file_path = set_job_file(type='collate')` and make `set_job_file` return the `job_file`

payu/models/fms.py
Qian-HuiChen: Sounds good! Implemented in the lates commit.

payu/subcommands/collate_cmd.py
Qian-HuiChen: I will work on schemas for run and collate. This should be built on top of Davide's changes on telemetry (#726). I have **not** implemented it yet in the latest commit.

payu/subcommands/collate_cmd.py
Qian-HuiChen: I should have removed this since `submit_job` write initial info into job file. This is removed in the latest commit.

aidanheerdegen:
I have reviewed and resolved all the conversations from my last review. Ping me again if you want me to take another look.

payu/models/fms.py
```diff
@@ -18,10 +18,14 @@
...
 import re
 import warnings
+import json
+
+from yamanifest.hashing import _hashlib
 
 from payu.models.model import Model
 from payu import envmod
 from payu.fsops import required_libs
+from payu.telemetry import get_job_file_path, update_job_file
```
jo-basevi:
```suggestion
```

Safe to remove these imports now

payu/models/fms.py
```diff
@@ -75,6 +79,58 @@ def get_avail_collate_flags(mppnc_path):
...
+        A dictionary structured as:
+        {future_collated_filename_1: [md5_hash_tile_1a, md5_hash_tile_1b, ...],
+        ...}
+    """
+    uncollate_hashes = {}
+
+    for nc_fname, tiles in mnc_tiles[prior_restart_path].items():
+        # record md5 hashes of all tiles in restart collation
+        uncollate_hashes[nc_fname] = [
+            _hashlib(os.path.join(prior_restart_path, tile), hashfn='md5') for tile in tiles
```
jo-basevi: Typically adding `_` in front of a function name means it's private - e.g. the function is only intended to be used inside of a file or internally within the library code (yamanifest). 

I think it is probably fine in this case as I don't think yamanifest code is prone to change and we can update the code here if needed. There is the [`hash`](https://github.com/ACCESS-NRI/yamanifest/blob/8d84aaefab12bc5e34972af4f4ed46908289526b/yamanifest/hashing.py#L67) method which could work?

payu/models/fms.py
```diff
@@ -75,6 +79,58 @@ def get_avail_collate_flags(mppnc_path):
...
+                   restart_path: {collated_filename: [list_of_tile_filenames]}}
+
+    Returns:
+        A dictionary structured as:
+        {future_collated_filename_1: [md5_hash_tile_1a, md5_hash_tile_1b, ...],
+        ...}
+    """
+    uncollate_hashes = {}
+
+    for nc_fname, tiles in mnc_tiles[prior_restart_path].items():
```
jo-basevi: So by default, collation doesn't collate restarts so we wouldn't have a key for restarts in `mnc_tiles`. Because mnc_tiles is initialised as `mnc_tiles = defaultdict(defaultdict(list).copy)`, it won't raise a key error here as `mnc_tiles[prior_restart_path]` would evaluate to `{}`, and the function will return None.

It might be worth being explicit and adding this above? 

```
if prior_restart_path is None:
   return None
```

payu/models/fms.py
```diff
@@ -75,6 +79,58 @@ def get_avail_collate_flags(mppnc_path):
     return set(re.findall(r"\B-[A-Za-z0-9]+\b", collate_help.stdout))
 
 
+def get_uncollate_hashes(mnc_tiles, prior_restart_path):
+    """ Map future collated filenames to the MD5 hashes of each uncollated tiles.
+
+    Args:
+        mnc_tiles: A dictionary structured as:
+                   {output_path: {collated_filename: [list_of_tile_filenames]},
+                   restart_path: {collated_filename: [list_of_tile_filenames]}}
```
jo-basevi: It'll be good to update the description as it only returns files related to restarts and add the prior_restart_path argument

payu/models/fms.py
```diff
@@ -75,6 +79,58 @@ def get_avail_collate_flags(mppnc_path):
...
+            "md5_hash_collated_filename_1": ["md5_hash_tile_1a", "md5_hash_tile_1b", ...],
+            "md5_hash_collated_filename_2": ["md5_hash_tile_2a", "md5_hash_tile_2b", ...],
+            ...
+        }
+    }
+    """
+    if model.prior_restart_path is None:
+        return None
+    else:
+        prior_restart_num = re.search(r'restart(\d+)', model.prior_restart_path).group()
```
jo-basevi: Unlikely to ever happen, but this could have a bug if restartX was somewhere in the path, e.g.

```
>>> re.search(r'restart(\d+)', 'path/to/test-with-restart0/archive/restart000').group()
'restart0'
```

You could use `$` for end of string in the regex, e.g. 
```
>>> re.search(r'restart(\d+)$', 'path/to/test-with-restart0/archive/restart000').group()
'restart000'
```

But in this case, I think it'll be simple to use the `os` or `pathlib` libraries to get the name of the directory: 

```
>>> import os
>>> os.path.basename('path/to/test-with-restart0/archive/restart000')
'restart000'
>>> from pathlib import Path
>>> Path('path/to/test-with-restart0/archive/restart000').name
'restart000'
```

payu/models/fms.py
```diff
@@ -75,6 +79,58 @@ def get_avail_collate_flags(mppnc_path):
...
+    mapping_collate_dict = {}
+    mapping_collate_dict[prior_restart_num] = {}
+
+    for nc_fname in mnc_tiles[model.prior_restart_path]:
+        # calculate md5 hash of the final collated file
+        nc_path = os.path.join(model.prior_restart_path, nc_fname)
+        collate_hash = _hashlib(nc_path, hashfn='md5')
+
+        # match the collated file hash with the list of uncollated tile hashes
+        mapping_collate_dict[prior_restart_num][collate_hash] = uncollate_hashes_dict.pop(nc_fname)
```
jo-basevi: Is there a reason to use `pop` here vs `uncollate_hashes[nc_fname]`?

To simplify it slightly, you could remove the `mnc_tiles` as an argument to this function as I think `uncollate_hashes_dict` has all the filenames anyway so the for loop could be:

`for nc_fname, uncollated_hashes in uncollate_hashes_dict.items():`

payu/subcommands/collate_cmd.py
```diff
@@ -3,13 +3,17 @@
...
+import shutil
 
 # Local
 from payu import cli
 from payu.experiment import Experiment
 from payu.laboratory import Laboratory
 import payu.subcommands.args as args
+from payu.telemetry import record_run
 from payu import fsops
+from payu.schedulers.pbs import get_job_info_json
```
jo-basevi: Safe to remove these unused imports for `shutil` and `from payu.schedulers.pbs import get_job_info_json`

payu/subcommands/profile_cmd.py
```diff
@@ -62,7 +62,16 @@ def runcmd(model_type, config_path, init_run, n_runs, lab_path):
...
 
-    cli.submit_job('payu-profile', pbs_config, pbs_vars)
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+    if init_run is None:
+        # Get the latest run number from the restart/output folder numbering
+        # and use it as the run number to write job file
+        expt.set_counters(keep_run_number=True)
+        init_run = expt.counter
```
jo-basevi: We probably don't want to have a queued job file for payu profile currently, so could pass `expt=None` to `cli.submit_job`

payu/subcommands/sync_cmd.py
```diff
@@ -58,7 +58,16 @@ def runcmd(model_type, config_path, lab_path, dir_path, sync_restarts,
...
 
-    cli.submit_job('payu-sync', pbs_config, pbs_vars)
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+    if init_run is None:
+        # Get the latest run number from the restart/output folder numbering
+        # and use it as the run number to write job file
+        expt.set_counters(keep_run_number=True)
+        init_run = expt.counter
```
jo-basevi: Similarly to `payu profile`, `expt` could be None until we have updating job files for `payu sync`.

payu/experiment.py
```diff
@@ -167,13 +167,16 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
...
+        job_file = telemetry.get_job_file_path(
             archive_path=Path(self.archive_path),
             run_number=self.counter,
             timings=self.timings,
             scheduler=self.scheduler,
             type=type,
         )
+        if type == 'run':
+            self.job_file = job_file
+        return job_file
```
jo-basevi: This is a very minor comment so feel free to ignore this one, but could break apart the get and set functionalities to separate functions, e.g.

```python
    def get_job_file(self, type='run'):
       """Returns the job file for the type of job"""
       return telemetry.get_job_file_path(
            archive_path=Path(self.archive_path),
            run_number=self.counter,
            timings=self.timings,
            scheduler=self.scheduler,
            type=type,
        )
        
    def set_run_job_file(self):
        """Set the job file for the payu run job"""
        self.job_file = get_job_file(type='run')
```

payu/subcommands/collate_cmd.py
```diff
@@ -86,7 +90,20 @@ def runcmd(model_type, config_path, init_run, lab_path, dir_path):
...
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+    if init_run is None:
+        # Get the latest run number from the restart/output folder numbering
+        # and set it as the run number to write job file
+        expt.set_counters(keep_run_number=True)
+        init_run = expt.counter
+
+    print(f"init_run: {init_run}")
```
jo-basevi:
```suggestion
```

jo-basevi: I'm assuming this is from debugging but if you want to keep this line, maybe display "Payu initial run counter: {init_run}"

payu/subcommands/run_cmd.py
```diff
@@ -173,20 +173,9 @@ def runcmd(model_type, config_path, init_run, n_runs, lab_path,
         # Return error code.
         sys.exit(1)
 
-    job_id = cli.submit_job('payu-run', pbs_config, pbs_vars)
-
     current_run = init_run if init_run is not None else expt.counter
 
-    # This could be done as part of submit_job eventually, but for now
-    # it's only used by the run command.
-    write_queued_job_file(
```
jo-basevi, line 182: Safe to remove `write_queued_job_file` from imports now

payu/experiment.py
```diff
@@ -963,12 +972,30 @@ def archive(self, force_prune_restarts=False):
...
 
+    @timeit("payu_collate_duration_seconds")
     def collate(self):
+        """ Run model collation and record the time taken in seconds to run collation"""
         # Setup modules - load user-defined modules
         self.setup_modules()
 
+        # Counters does not increase for collate runs
+        # This is used to determine the job file path
+        self.set_counters(keep_run_number=True)
```
jo-basevi: I would shift this line til after the collate methods, just before 
`job_file_path = self.set_job_file(..)`, incase collation code ever uses `expt.counter`. 

I think for most cases (when payu collate was submitted with `-i/--initial`), it'll be the same value (as 'PAYU_CURRENT_RUN' is set in the PBS job).

payu/experiment.py
```diff
@@ -1064,6 +1092,7 @@ def sweep(self, hard_sweep=False):
         default_job_name = os.path.basename(os.getcwd())
         short_job_name = str(self.config.get('jobname', default_job_name))[:15]
 
+        # find all PBS log files including payu runs, collate runs, postscript runs, and sync runs
```
jo-basevi, line 1093: Nice documentation :)

payu/status.py
```diff
@@ -3,7 +3,7 @@
 payu runs by inspecting the job files generated for telemetry,
 scheduler stdout/stderr logs, and querying the scheduler
 """
-
+import os
```
```suggestion
```
Unused import

payu/status.py
```diff
@@ -184,6 +184,31 @@ def build_job_info(
         run_num = data["payu_current_run"]
         runs.setdefault(run_num, {"run": []})["run"].append(run_info)
 
+        collate_dir = archive_path / "payu_jobs" / str(run_num) / "collate"
+        if collate_dir.exists():
+            collate_files = sorted(list(collate_dir.glob("*.json")), key=lambda f: f.stat().st_mtime)
+
+            if collate_files:
+                collate_file = collate_files[-1]
```
jo-basevi: With `payu status --all`, it includes all the job files (this includes any payu job files that exited with errors). Similarly it might good to also display collate jobs for a given run if `--all` is provided, but still just default to the latest.

jo-basevi: If you did this it'll be worth changing the structure of this dictionary to split collate into a separate section (somewhat similar to how its organised in the filesystem with collate and run subdirectories).

Currently its:
```json
{
    "experiment_uuid": "805c3908-6a45-497e-9a4f-a8e6b72d544e",
    "runs": {
        "0": {
            "run": [
                {
                    "job_id": "167689233.gadi-pbs",
                    "run_id": "02109c0e87abff3d7772de85227e26b437c4243f",
                    "stage": "archive",
                    "exit_status": 0,
                    "model_exit_status": 0,
                    "model_finish_time": "1958-04-01T00:00:00",
                    "stdout_file": "/path/to/ctrldir/1deg_jra55_iaf.o167689233",
                    "stderr_file": "/path/to/ctrldir/1deg_jra55_iaf.e167689233",
                    "job_file": "/path/to/archive/payu_jobs/0/run/167689233.gadi-pbs.json",
                    "start_time": "2026-05-04T12:38:50.336588",
                    "collate_info": {
                        "job_id": "167689349.gadi-pbs",
                        "stage": "exited",
                        "exit_status": 0,
                        "stdout_file": "/path/to/ctrldir/1deg_jra55_iaf/1deg_jra55_ia_c.o167689349",
                        "stderr_file": "/path/to/ctrldir/1deg_jra55_iaf/1deg_jra55_ia_c.e167689349",
                        "job_file": "/path/to/archive/payu_jobs/0/collate/167689349.gadi-pbs.json"
                    }
                }
            ]
        }
    }
}
```

Separate collate section:
```json
{
    "experiment_uuid": "805c3908-6a45-497e-9a4f-a8e6b72d544e",
    "runs": {
        "0": {
            "run": [
                //{...}, failed previous payu runs 
                {
                    "job_id": "167689233.gadi-pbs",
                    "run_id": "02109c0e87abff3d7772de85227e26b437c4243f",
                    "stage": "archive",
                    "exit_status": 0,
                    "model_exit_status": 0,
                    "model_finish_time": "1958-04-01T00:00:00",
                    "stdout_file": "/path/to/ctrldir/1deg_jra55_iaf.o167689233",
                    "stderr_file": "/path/to/ctrldir/1deg_jra55_iaf.e167689233",
                    "job_file": "/path/to/archive/payu_jobs/0/run/167689233.gadi-pbs.json",
                    "start_time": "2026-05-04T12:38:50.336588",
                }
            ],
            "collate": [
                //{...}, previous payu collate jobs
                {
                    "job_id": "167689349.gadi-pbs",
                    "stage": "exited",
                    "exit_status": 0,
                    "stdout_file": "/path/to/ctrldir/1deg_jra55_iaf/1deg_jra55_ia_c.o167689349",
                    "stderr_file": "/path/to/ctrldir/1deg_jra55_iaf/1deg_jra55_ia_c.e167689349",
                    "job_file": "/path/to/archive/payu_jobs/0/collate/167689349.gadi-pbs.json"
                }
            ]
        }
    }
}
```

payu/telemetry.py
```diff
@@ -562,7 +568,9 @@ def record_run(
         Path to the archive directory for the experiment
     """
     # Additional information to the run info
-    run_info = {"payu_run_status": run_status}
+    run_info = {run_info_label: run_status}
```
jo-basevi: Could use the type to get the status label, e.g. `payu_{type}_status`.

test/models/test_fms.py
```diff
@@ -289,4 +291,54 @@ def test_get_avail_collate_flags_runtimeerror(mock_run):
...
+    mock_hash.side_effect = lambda file_path, hashfn: f"md5_{os.path.basename(file_path)}"
+    
+    # Set up a temporary job file
+    job_file_path = tmpdir / "archive" / "payu_jobs" / "3" / "collate" / "test-jobid-id-3.json"
+    job_file_path.parent.mkdir(parents=True, exist_ok=True)
+    with open(job_file_path, 'w') as f:
+        json.dump({
+            "stage": "queued",
+        }, f)
+    mock_get_job_file_path.return_value = job_file_path
```
jo-basevi: The job file code is removed so I think it's ok to remove this test setup.

test/test_fsops.py
```diff
@@ -2,6 +2,7 @@
 import pytest
 import os
 from unittest.mock import patch
+from yamanifest.hashing import _hashlib
```
jo-basevi: Unused import 
```suggestion
```

jo-basevi (commented):
Thanks for reviewing everything from the last review :) I think the changes you added look good to me! Most of my comments are very minor

payu/experiment.py
Qian-HuiChen: Sounds like a good idea. Implemented in the latest commit

payu/experiment.py
Qian-HuiChen: Okie, I moved it to after the collate mapping.

payu/experiment.py
Qian-HuiChen: hehe thanks :)

payu/status.py
Qian-HuiChen: Removed in the latest commit

payu/telemetry.py
Qian-HuiChen: Good point. Implemented in the latest commit.

test/models/test_fms.py
Qian-HuiChen: Implemented in the latest commit

test/test_fsops.py
Qian-HuiChen: Removed in the latest commit

payu/models/fms.py
Qian-HuiChen: Removed redundant imports in the latest commit

payu/subcommands/collate_cmd.py
Qian-HuiChen: Removed unused imports in the latest commit

payu/subcommands/run_cmd.py
Qian-HuiChen: Removed in the latest commit

payu/models/fms.py
Qian-HuiChen: Updated in the latest commit

payu/models/fms.py
Qian-HuiChen: Sounds good. I did 
```
if prior_restart_path is None:
    return {}
```
in the latest commit.

payu/models/fms.py
Qian-HuiChen: I understand that it looks awkward to see a function starting with `_`. I also think it is a bit redundant to use the entire hash function because we only use `_hashlib`. So in the latest commit, I use `from yamanifest.hashing import _hashlib as hashlib` instead.

payu/models/fms.py
Qian-HuiChen: Sounds good. I change this to `os.path.basename` in the latest commit

payu/models/fms.py
Qian-HuiChen: There is no specific reason why I used pop. I was imagining that I pop every single key out of uncollate_hashes and there is an empty {} left.

Good idea. I update the mapping_log to
```
    for nc_fname, tile_hashes in uncollate_hashes_dict.items():
        # calculate md5 hash of the final collated file
        nc_path = os.path.join(model.prior_restart_path, nc_fname)
        collate_hash = hashlib(nc_path, hashfn='md5')

        # match the collated file hash with the list of uncollated tile hashes
        mapping_collate_dict[prior_restart_num][collate_hash] = tile_hashes
```

payu/subcommands/collate_cmd.py
Qian-HuiChen: Oops, you found my debug printing. I removed this in the latest commit.

payu/subcommands/profile_cmd.py
Qian-HuiChen: Sounds good. I changed it to
```
# Submit PBS job with expt = None so no job file is written
    cli.submit_job('payu-sync', pbs_config, pbs_vars)
```
in the latest commit

payu/subcommands/sync_cmd.py
Qian-HuiChen: Done in the latest commit.

payu/status.py
Qian-HuiChen: Sounds good. I change the runs dict to `{"0":{"run":[], "collate":[]}}` in commit [2ccbd4a](https://github.com/payu-org/payu/pull/722/commits/2ccbd4ab75b91de11d2b31cf9bd2ae56eed5549b).

payu/models/fms.py
jo-basevi: If using `_hashlib`, I would almost prefer to keep calling it `_hashlib` to make it more clear in the code that a private function is being used at that point. 

It's just to protect against the case of someone, who is working on `yamanifest`, changing or deleting private functions (as they could assume they are not used outside of the `yamanifest` code). While `hash` method is public so if it is modified, it'll be considered a breaking change.

payu/subcommands/profile_cmd.py
jo-basevi: Thanks, could remove initialising the expt too as it's unused, e.g.

```
    # Initialise experiment to determine archive path and run number (which is needed to write job file)
    lab = Laboratory(model_type, config_path, lab_path)
    expt = Experiment(lab)
```

payu/subcommands/sync_cmd.py
jo-basevi: Similarly can remove expt even though we will probably add it again soon when adding the sync job files

Qian-HuiChen: Removed in the latest commit.

payu/subcommands/profile_cmd.py
Qian-HuiChen: Good catch! I removed this in the latest commit.

payu/models/fms.py
Qian-HuiChen: I see. I didn't consider the possibility of someone changing yamanifest and deleting _hashlib. I changed payu to use hash in the latest commit.

payu/models/fms.py
```diff
@@ -75,6 +77,60 @@ def get_avail_collate_flags(mppnc_path):
...
+    else:
+        prior_restart_num = os.path.basename(model.prior_restart_path)
+
+    mapping_collate_dict = {}
+    mapping_collate_dict[prior_restart_num] = {}
+
+    for nc_fname, tile_hashes in uncollate_hashes_dict.items():
+        # calculate md5 hash of the final collated file
+        nc_path = os.path.join(model.prior_restart_path, nc_fname)
+        collate_hash = hash(nc_path, hashfn='md5')
```
aidanheerdegen: We will want to match the hashes in the manifests with this mapping, so we need to make sure we're using the same hashing algorithm.

We define this here:

https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/manifest.py#L27

We should use that I think for consistency.

aidanheerdegen (changes requested):
One small change from me.

payu/models/fms.py
jo-basevi: Technically `config.yaml` could over-ride the hash value.. But it's also under a comment of not supporting specifying hash values (its not in the documentation).
https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/manifest.py#L247-L249

payu/subcommands/collate_cmd.py
jo-basevi: Sounds good to me

payu/models/fms.py
aidanheerdegen: Yeah when I was considering what to say in the comment above I was considering suggesting querying the manifest, but the idea of changing hashes and the ramifications of that made me sick, so I didn't.

payu/models/fms.py
Qian-HuiChen: It turns out `prior_restart_path = path/to/test-with-restart0/archive/restart000/ocean`. I changed this line to 
```
rel_path = os.path.relpath(model.prior_restart_path, start=archive_path)
prior_restart_num = os.path.split(rel_path)[0]
```
in the latest commit.

payu/status.py
Qian-HuiChen: I added all previous collation into the dictionary (status_data) in commit [ecefb7e](https://github.com/payu-org/payu/pull/722/commits/ecefb7e2157e85c9739751d84e6ef8bec4624623). Also slight changed the display block so `payu status --all` will show:
```shell
Run: 11
  Job ID:            160000.gadi-pbs
  Run ID:            a0e23d
  ::
  Job ID:            160001.gadi-pbs
  Run ID:            a083
  ::
  ------------- Collate Info -------------
  Job ID:            167928.gadi-pbs
  Stage:             exited
  ::
```

payu/models/fms.py
Qian-HuiChen: Sounds good. I imported full_hashes from manifests.py in commit [ecefb7e](https://github.com/payu-org/payu/pull/722/commits/ecefb7e2157e85c9739751d84e6ef8bec4624623)

payu/models/fms.py
```diff
@@ -75,7 +81,64 @@ def get_avail_collate_flags(mppnc_path):
...
+            ...
+        }
+    }
+    """
+    if model.prior_restart_path is None:
+        return None
+    else:
+        # Get the relative path from the archive dir to the prior restart dir
+        # e.g., /restart001/ocean/, use the first part as the restart number
+        rel_path = os.path.relpath(model.prior_restart_path, start=archive_path)
```
jo-basevi: archive_path can be obtained from the `model` object (vs passing the archive_path from the experiment to the collate method). I think `model.expt.archive_path` could work?

payu/status.py
```diff
@@ -145,63 +146,85 @@ def build_job_info(
...
+            run_jobs[job_type].sort(key=lambda x: (
+                # Put queued jobs at the end (None start_time)
+                x.get("start_time") is None,
+                # Sort by start time
+                x.get("start_time") or ""
+            ))
+
+        if not all_runs:
+            # Use latest run/collate job
+            run_jobs[job_type] = [run_jobs[job_type][-1]]
```
```suggestion
            if not all_runs:
                # Use latest run/collate job
                run_jobs[job_type] = [run_jobs[job_type][-1]]
```
I think this needs to be indented so it's in the for-loop (so it runs for `run` type as well as `collate`

payu/status.py
```diff
@@ -145,63 +146,85 @@ def build_job_info(
...
+                type=data.get("scheduler_type")
+            )
+
+            run_info = {
+                "job_id": data.get("scheduler_job_id"),
+                "run_id": data.get("payu_run_id", None),
+                "stage": data.get("stage"),
+                "exit_status": data.get(f"payu_{job_type}_status"),
+                "model_exit_status": data.get("payu_model_run_status", None),
+                "model_finish_time": data.get("model_finish_time", None),
```
jo-basevi: `model_exit_status`, `model_finish_time` and `run_id` are all run job specific so could be added straight after this dict is created `if job_type == "run":`?

jo-basevi (commented):
The refactor in the job files looks good and it shows `payu status --all` as expected :)

 I haven't done a super thorough review but I thought I'll post some initial comments now.

payu/status.py
aidanheerdegen: If the tests didn't pick this up should consider adding a test that would have done so.

payu/models/fms.py
Qian-HuiChen: Thanks! I tested it and it is true that `model.expt.archive_path` would work as well. This cleans up the code a lot :D

payu/status.py
Qian-HuiChen: Okie, I implemented this in the latest commit.

payu/status.py
Qian-HuiChen: Sounds good. I add a unit test in the latest commit (`test_build_job_info_collate`) to cover this.

payu/models/fms.py
Qian-HuiChen: I tested it by doing a `payu run` with collate turned off, and then do a separate `payu collate`. 

The weird thing is: it is run number `13`, collate job file is stored under `payu_jobs/13/collate/` (which is as expected), but the `collate_mapping` dict is `restart013` (expected to be `restart012`). This means that `model.prior_restart_path` did not find the correct serial in this case.

I tested it again with `payu run` with collate is turned on in `config.yaml`. The numbering is correct in this case.

jo-basevi: We chatted about this earlier in person - the default `payu collate` behaviour seems to collate only the latest restart (and not the latest output). So to collate the latest output, need to run `payu collate -i N` where `N` is the number of the latest output, or `payu collate -d /path/to/latest/outputN`. We could change this default behaviour to collate the latest output and the previous restart (similar to the automatic collate job), but this could be a separate issue/PR as it changes existing behaviour.

aidanheerdegen (dismissed):
Wow, that turned into a pretty big PR! Thanks for wrangling that and incorporating all the feedback.

docs/source/usage.rst
```diff
@@ -413,7 +385,8 @@ collate output from run ``K``, type the following::
 This will also collate restart ``K-1`` if ``restart: true`` in the ``collate``
 section of the configuration file.
 
-Alternatively you can directly specify a directory name::
+Alternatively you can directly specify a directory name containing the uncollated files 
+(e.g., ``restart001/ocean``)::
```
jo-basevi:
```suggestion
(e.g., ``archive/restart001/ocean``)::
```
Maybe add in the archive symlink so it's a valid path when running in the control directory?

Qian-HuiChen: Sounds good :)

jo-basevi (approved):
Thanks @Qian-HuiChen, it looks good to me! The scheduler job info is now showing the files and it picks up the restart collate hashes if I use `payu collate -d` on an earlier restart directory.

PR #726, opened 2026-04-24
Update payu telemetry

### LINKED ISSUE #725: payu telemetry configuration definition
### Overview
As per discussed with @jo-basevi, and as mentioned in this [related issue](https://github.com/ACCESS-NRI/containerised-environments-infra/issues/4#issue-4228533790), we should add the telemetry configuration definition (i.e., the JSON file used within payu's [telemetry](https://github.com/payu-org/payu/blob/master/payu/telemetry.py) module) within this repository.

This will facilitate changing the configuration and payu's code at the same in case any updates are needed.

In general, the following changes will be required:

### Telemetry Configuration definition
The telemetry configuration definition currently used follows the schema:
```json
{
  "telemetry_url": <str>,
  "telemetry_proxy_url": <str>,
  "telemetry_token": <str>,
  "telemetry_service_name": <str>,
  "hostname": <str>
}
```

The proposal is to add a `"version": <str>` field, to have a way to keep track of changes in the config (which should result in the upgrade of the version number). 
Thus the definition will become:

```json
{
 "version": <str>,
  "telemetry_url": <str>,
  "telemetry_proxy_url": <str>,
  "telemetry_token": <str>,
  "telemetry_service_name": <str>,
  "hostname": <str>
}
```

Additionally, the values of these fields are considered to be sensitive, therefore will not be hardcoded in the definition. Instead, some placeholders will be set so the values can be dynamically inferred at deployment time (see the deployment paragraph below).
Solution for the placeholders might be:
- a double underscore string in the format `__<ENV_VARIABLE_NAME>__` that define the environment variable (`ENV_VARIABLE_NAME`) whose value will be used at deployment time to fill out the fields.
  Example:
  ```json
  {
  "version": "__PAYU_TELEMETRY_VERSION__",
  "telemetry_url": "__PAYU_TELEMETRY_URL__",
  "telemetry_proxy_url": "__PAYU_TELEMETRY_PROXY_URL__",
  "telemetry_token": "__PAYU_TELEMETRY_TOKEN__",
  "telemetry_service_name": "__PAYU_TELEMETRY_SERVICE_NAME__",
  "hostname": "__PAYU_TELEMETRY_HOSTNAME__"
  }
  ```
- a Jinja2 template that can be run through a jinja interpreter to produce the configuration (probably still based on env variables)
   Example:
   ```json
    {
    "version": "{{ env['PAYU_TELEMETRY_VERSION'] }}",
    "telemetry_url": "{{ env['PAYU_TELEMETRY_URL'] }}",
    "telemetry_proxy_url": "{{ env['PAYU_TELEMETRY_PROXY_URL'] }}",
    "telemetry_token": "{{ env['PAYU_TELEMETRY_TOKEN'] }}",
    "telemetry_service_name": "{{ env['PAYU_TELEMETRY_SERVICE_NAME'] }}",
    "hostname": "{{ env['PAYU_TELEMETRY_HOSTNAME'] }}"
    }
    ```
    to be used in conjunction with this [jinja2-action](https://github.com/cuchi/jinja2-action)

### Telemetry Configuration deployment
The deployment of payu's telemetry configuration to the HPC targets where `payu` is used will be handled by the [containerised-environments-infra](https://github.com/ACCESS-NRI/containerised-environments-infra/) repository (as mentioned in the related issue above).

### Payu code change
In light of the modifications above, and for the reasons mentioned in the related issue above, the proposal is to change payu's [telemetry](https://github.com/payu-org/payu/blob/master/payu/telemetry.py) code to be using a `PAYU_TELEMETRY_CONFIG_PATH` env variable to retrieve the telemetry configuration path, instead of setting the two variables:

https://github.com/payu-org/payu/blob/cbbc771c1d2dc86da93687ec01af88195ac7e4f3/payu/telemetry.py#L26-L28

where `PAYU_TELEMETRY_CONFIG` is an env variable used to retrieve the configuration:

https://github.com/payu-org/payu/blob/cbbc771c1d2dc86da93687ec01af88195ac7e4f3/payu/telemetry.py#L135-L137

With the proposed changes the code above would become something like:
```py
# Environment variable for external telemetry configuration file
TELEMETRY_CONFIG = "PAYU_TELEMETRY_CONFIG_PATH"
...
...
...
config_path = Path(os.environ[TELEMETRY_CONFIG])
```

atteggiani:
More context in #725 and [ACCESS-NRI/containerised-environments-infra/issues/4](https://github.com/ACCESS-NRI/containerised-environments-infra/issues/4).

Updated payu telemetry's module:

- [x] the env variable name where the config path is stored has been updated to `PAYU_TELEMETRY_CONFIG_PATH`
- [x] Updated logic to skip telemetry recording if the `PAYU_TELEMETRY_CONFIG_PATH` is set but empty
- [x] Updated telemetry not enabled tests to enforce testing only the specific condition to skip telemetry (before there were multiple conditions not met, so the tests would not really test what they should have been testing)
- [x] Added telemetry configuration specification as a JSON jinja template to be interpolated using the [cuchi/jinja2-action](https://github.com/cuchi/jinja2-action) action

### DIFF
--- a/payu/telemetry.py
+++ b/payu/telemetry.py
@@ -5,17 +5,14 @@ this data to telemetry services at the end of a run, if configured.
 
 import datetime
 import json
+import jsonschema
 import os
-import stat
 from pathlib import Path
 import requests
-import shutil
-import tempfile
 import threading
 from typing import Any, Optional
 import warnings
 from filelock import FileLock, Timeout
-import warnings
 
 import cftime
 
@@ -25,6 +22,7 @@ from payu.fsops import atomic_write_file
 
 # Environment variable for external telemetry configuration file
 TELEMETRY_CONFIG = "PAYU_TELEMETRY_CONFIG_PATH"
+TELEMETRY_CONFIG_SCHEMA = Path(__file__).parent / "telemetry" / "telemetry_config.schema.json"
 
 # Required telemetry configuration fields
 CONFIG_FIELDS = {
@@ -152,12 +150,15 @@ def get_external_telemetry_config(
         write_error_log(archive_path, job_file_path, error_msg)
         return None
 
-    # Check for required fields in the telemetry configuration
-    missing_fields = CONFIG_FIELDS.values() - telemetry_config.keys()
-    if missing_fields:
+    # Check that the telemetry configuration follows its schema
+    with open(TELEMETRY_CONFIG_SCHEMA, 'r') as f:
+        schema = json.load(f)
+    try:
+        jsonschema.validate(instance=telemetry_config, schema=schema)
+    except jsonschema.ValidationError as e:
         error_msg = (
-            f"Required field(s) {missing_fields} not found in configuration "
-            f"file specified by {TELEMETRY_CONFIG}: {config_path}."
+            f"The telemetry configuration file {config_path} specified by {TELEMETRY_CONFIG} "
+            f"does not follow the required schema: {e.message}"
         )
         write_error_log(archive_path, job_file_path, error_msg)
         return None
--- a/payu/telemetry.py
+++ b/payu/telemetry.py
@@ -24,8 +24,7 @@ from payu.schedulers.scheduler import Scheduler
 from payu.fsops import atomic_write_file
 
 # Environment variable for external telemetry configuration file
-TELEMETRY_CONFIG = "PAYU_TELEMETRY_CONFIG"
-TELEMETRY_CONFIG_VERSION = "1-0-0"
+TELEMETRY_CONFIG = "PAYU_TELEMETRY_CONFIG_PATH"
 
 # Required telemetry configuration fields
 CONFIG_FIELDS = {
@@ -133,11 +132,10 @@ def get_external_telemetry_config(
     If a valid file does not exist, return None
     """
     # Check path to telemetry config file exists
-    config_dir = Path(os.environ[TELEMETRY_CONFIG])
-    config_path = config_dir / f"{TELEMETRY_CONFIG_VERSION}.json"
+    config_path = Path(os.environ[TELEMETRY_CONFIG])
     if not (config_path.exists() and config_path.is_file()):
         error_msg = (
-            f"No config file found at {TELEMETRY_CONFIG}: {config_path}."
+            f"No config file found at the path specified by {TELEMETRY_CONFIG}: {config_path}."
         )
         write_error_log(archive_path, job_file_path, error_msg)
         return None
@@ -148,7 +146,7 @@ def get_external_telemetry_config(
             telemetry_config = json.load(f)
     except json.JSONDecodeError:
         error_msg = (
-            "Error parsing json in configuration file at "
+            "Error parsing json in configuration file specified by "
             f"{TELEMETRY_CONFIG}: {config_path}."
         )
         write_error_log(archive_path, job_file_path, error_msg)
@@ -159,7 +157,7 @@ def get_external_telemetry_config(
     if missing_fields:
         error_msg = (
             f"Required field(s) {missing_fields} not found in configuration "
-            f"file at {TELEMETRY_CONFIG}: {config_path}."
+            f"file specified by {TELEMETRY_CONFIG}: {config_path}."
         )
         write_error_log(archive_path, job_file_path, error_msg)
         return None
--- a/payu/telemetry.py
+++ b/payu/telemetry.py
@@ -7,6 +7,7 @@ import datetime
 import json
 import jsonschema
 import os
+from importlib.resources import files
 from pathlib import Path
 import requests
 import threading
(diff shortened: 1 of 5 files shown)


payu/telemetry/telemetry_config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "version": "1.0",
```
Qian-HuiChen: If the telemetry schema is 1-0-0.json, should we name the version as "1.0.0"?

payu/telemetry/telemetry_config.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "additionalProperties": false,
+    "properties": {
+        "version": {
+            "type": "string"
+        },
+        "telemetry_url": {
+            "type": "string",
+            "format": "uri"
+        },
```
Qian-HuiChen, line 19: It will be great if a description about "url", "proxy_url", "service_name" and "hostname" can be added.

payu/telemetry.py
```diff
@@ -272,7 +271,7 @@ def record_telemetry(run_info: dict[str, Any],
     # and whether the model was run
     if not (
         config.get("telemetry", {}).get("enable", True)
-        and TELEMETRY_CONFIG in os.environ
+        and os.environ.get(TELEMETRY_CONFIG) not in (None, "")
```
Qian-HuiChen, line 277: The logic seems to be changed between these two lines. Could you please explain a bit more to me? I might not understand our goal here.

Qian-HuiChen (commented):
Thanks for the changes! It will make my follow-up work (adapting schema to payu-run and payu-collate) much easier.

payu/telemetry/telemetry_config.json
atteggiani: Sure, updated in [4f166de](https://github.com/payu-org/payu/pull/726/commits/4f166de84ac625ce8e4fa3212ac258f14e8fbc00)

payu/telemetry.py
atteggiani: Good point, I should have explained it better. 
I mention in this PR body:
> - [x] Updated logic to skip telemetry recording if the PAYU_TELEMETRY_CONFIG_PATH is set but empty

That's exactly what the logic changes. 
With the old logic, if the env variable identified by the python variable `TELEMETRY_CONFIG` (i.e., `PAYU_TELEMETRY_CONFIG`) was empty, payu would have thrown an error for "file not found" ([related code](https://github.com/payu-org/payu/blob/cbbc771c1d2dc86da93687ec01af88195ac7e4f3/payu/telemetry.py#L138-L141)). With the new logic, the telemetry recording is skipped if the env variable identified by the python variable `TELEMETRY_CONFIG` (that has been changed to `PAYU_TELEMETRY_CONFIG_PATH`) is set but empty.

This will simplify the integration with the containerised infrastructure, in which is easier to set this variable to an empty value rather than not setting it.

payu/telemetry/telemetry_config.schema.json
atteggiani: Good point!
Updated in [87a5359](https://github.com/payu-org/payu/pull/726/commits/87a53597b9cd6e597a34e5133f42fb626b0158db)

I also hardcoded `telemetry_service_name` to `"payu"` as it's always the case.

atteggiani:
The features in this PR can be tested on Gadi by loading:
```
module use /g/data/vk83/testing/containerised_environments/admin/staging/prerelease/modules
module load payu/dev-20260425T153126-76bce66-pr32
```

jo-basevi (commented):
Thanks @atteggiani, the changes look good to me. The schema validation check is neat. 

I've tested that the `payu/dev-20260425T153126-76bce66-pr32` sends telemetry without issue. 

I would hold off merging this until we have the new `payu/dev` module in production so we still have telemetry running with `payu/dev`. (Alternatively we could add `PAYU_TELEMETRY_CONFIG_PATH` environment variable to the current infra)

payu/telemetry.py
Qian-HuiChen: Thanks for explaining this!

Qian-HuiChen:
Thanks @atteggiani for the changes! `payu/dev-20260425T153126-76bce66-pr32` also runs fine on my side

jo-basevi (approved):
LGTM - will be good to get this change into `payu/dev` soon

atteggiani:
> LGTM - will be good to get this change into `payu/dev` soon

The [related PR](https://github.com/ACCESS-NRI/containerised-environments-infra/pull/32) in the containerised environents infra is ready. 
If you'd like, we can merge this and merge that at the same time, and deploy a new `payu/dev` environment then (with the new infra).

jo-basevi:
Yes, we can do a quick test with the latest staging environment in the related PR, then update the PR to point to the `main` branch and merge at the same time?

atteggiani:
> Yes, we can do a quick test with the latest staging environment in the related PR, then update the PR to point to the `main` branch and merge at the same time?

Sounds good!

Qian-HuiChen:
Is it ok that I delete the branch davide/725-payu-telemetry-updates?

atteggiani:
Fine for me

PR #728, opened 2026-05-01
673/jasmeen jo/handling errors payu

KAUR1984:
Fixes issue #673. Related to draft PR - https://github.com/payu-org/payu/pull/709. Thanks so much @jo-basevi @Qian-HuiChen for your input on this and my apologies for not being able to create this pr earlier. **This is a draft PR for all the proposed changes, as it is not ready to be merged yet.** 

## **PR Description:** 
Currently, we have `sys.exit()` and `print()` statements for handling errors in payu. As mentioned in the #673 issue description, `sys.exit()` statements are hard to handle as they exit the program straight away. This PR is aiming to introduce a "bubbling up" approach for handling exception, as described in [this article](https://blog.miguelgrinberg.com/post/the-ultimate-guide-to-error-handling-in-python).

## **Key changes:**
- [x] **Centralised Error Hierarchy**: Added `payu/errors.py` which defines base `PayuError` class (for catching user errors) and other error classes are children from this parent class.  

- [x] **Bubbling up exceptions** - Replaced `sys.exit()` calls with `raise` statements. This avoids sudden exit from the program, and allows the exception to "bubble up" all the way to the main exception handler.
- [x] **Testing** - Not sure if this would be useful for testing purposes (very much down to be corrected) - perhaps with this we can write tests that verify specific errors are raised without crashing the whole program.
- [x] **Catch and recover** - For recoverable exceptions, in case we have any, we can specify what to do with exceptions handling, instead of `sys.exit()` calls.
- [x] **Standard vs Custom Python Exceptions** - Very much open to feedback on this. `errors.py` has all the custom Python exceptions, ex. `PayuConfigError`, `PayuGitError` etc. We can route the errors instantly based on the type of exception instead of having just one exception `PayuError` for catching all exceptions. These are intended to be not too specific to avoid overly many subclasses.

## **Questions:** 
1. Currently, a bit unclear if it is better to use `logging` module or print statements for the catch-all exception handler - `try-except block` in `cli.py`. Would appreciate any thoughts on this. I remember Jo said something about this.

2. All the custom exceptions come with `exit_code`. This could be useful to handle outputs based on just these codes, however, not sure if they are that useful in our case. Very happy to remove them otherwise.

> Reference: `gh` CLI

> Since this is a draft PR, @Qian-HuiChen pls feel free to check out to this branch in case that's easier for review, or adding more improvements. 

> @jo-basevi Very happy to create another pull request for the rest of the changes in PR - #709. Was a bit unsure if this PR is already too big.

> Opening up for suggestions. :)

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -27,6 +27,7 @@ from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 import payu.subcommands
 from payu.logger import setup_logger
 import payu.subcommands.args as arg_templates
+import payu.errors as errors
 
 # Default configuration
 DEFAULT_CONFIG = 'config.yaml'
@@ -59,8 +60,23 @@ def parse():
         warnings.formatwarning = (
             lambda message, category, filename, lineno, line=None: f"{message}"
         )
-        
-    run_cmd(**args)
+    try:
+        run_cmd(**args)
+
+    except errors.PayuError as e:
+        # CLEAN EXIT for expected errors
+        if stacktrace:
+            # Shows the full stacktrace
+            logging.exception(str(e), exc_info=True)
+        else:
+            print(f'payu: error: {e}', file=sys.stderr)
+        sys.exit(1)
+
+    except Exception as e:
+        # CRASH EXIT for unexpected bugs (ValueError, TypeError etc.)
+        print('payu: An unexpected internal error occurred!', file=sys.stderr)
+        logging.exception(str(e)) # Always show stacktrace for unknown bugs
+        sys.exit(1)
 
 
 def generate_parser(is_interactive=False):
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -65,18 +65,14 @@ def parse():
     run_cmd(**args)
 
     except errors.PayuError as e:
-        # CLEAN EXIT for expected errors
-        if stacktrace:
-            # Shows the full stacktrace
-            logging.exception(str(e), exc_info=True)
-        else:
-            print(f'payu: error: {e}', file=sys.stderr)
+        # Logs a clean message, and attaches stacktrace 'only' if requested.
+        logging.error(e, exc_info=stacktrace)
         sys.exit(1)
 
     except Exception as e:
         # CRASH EXIT for unexpected bugs (ValueError, TypeError etc.)
         print('payu: error: An unexpected internal error occurred!', file=sys.stderr)
-        logging.exception(str(e)) # Always show stacktrace for unknown bugs
+        logging.exception(e) # Always show stacktrace for unknown bugs
         sys.exit(1)
 
 
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -64,6 +64,21 @@ def parse():
         
     run_cmd(**args)
 
+    except errors.PayuError as e:
+        # CLEAN EXIT for expected errors
+        if stacktrace:
+            # Shows the full stacktrace
+            logging.exception(str(e), exc_info=True)
+        else:
+            print(f'payu: error: {e}', file=sys.stderr)
+        sys.exit(1)
+
+    except Exception as e:
+        # CRASH EXIT for unexpected bugs (ValueError, TypeError etc.)
+        print('payu: error: An unexpected internal error occurred!', file=sys.stderr)
+        logging.exception(str(e)) # Always show stacktrace for unknown bugs
+        sys.exit(1)
+
 
 def generate_parser(is_interactive=False):
     """Parse the command line inputs generate and return parser."""
--- a/payu/errors.py
+++ b/payu/errors.py
@@ -1,4 +1,12 @@
-'''Custom Payu Exceptions - for catching user errors.
+'''Custom Payu Exceptions. For catching user errors.
+ - Use them when user inputs are not recognised. 
+ - They intend to provide a polite, actionable message to users.
+
+Standard Exceptions (ex. ValueError, TypeError) are for the Developer.
+ - Use them to catch exceptions from another piece of code or 
+   function calls.
+ - We want the program to crash with a stack trace so developer 
+   could fix the bug.
 '''
 
 class PayuError(Exception):
@@ -36,4 +44,3 @@ class PayuRunError(PayuError):
     Raised when an active model run fails unexpectedly.
     '''
     exit_code = 5
-    
--- a/payu/errors.py
+++ b/payu/errors.py
@@ -1,0 +1,39 @@
+'''Custom Payu Exceptions - for catching user errors.
+'''
+
+class PayuError(Exception):
+    '''
+    Base class for all Payu Exceptions. 
+    Any error that inherits from this class - represents an expected failure 
+    mode (ex. bad user input, missing files) rather than a code bug.
+    '''
+    exit_code = 1
+
+
+class PayuConfigError(PayuError):
+    '''
+    Raised when there is an error reading the config.yaml file.
+    '''
+    exit_code = 2
+
+
+class PayuFileNotFoundError(PayuError):
+    '''
+    Raised when a required file or directory for a model is missing.
+    '''
+    exit_code = 3
+
+
+class PayuBranchError(PayuError):
+    '''
+    Raised when there are issues with Git branches.
+    '''
+    exit_code = 4
+
+
+class PayuRunError(PayuError):
+    '''
+    Raised when an active model run fails unexpectedly.
+    '''
+    exit_code = 5
+    
--- a/payu/errors.py
+++ b/payu/errors.py
@@ -1,19 +1,20 @@
-'''Custom Payu Exceptions. For catching user errors.
- - Use them when user inputs are not recognised. 
+'''Custom Payu Exceptions for handling user-facing errors.
+ - Use these when user input is invalid or not recognised. 
  - They intend to provide a polite, actionable message to users.
 
-Standard Exceptions (ex. ValueError, TypeError) are for the Developer.
- - Use them to catch exceptions from another piece of code or 
-   function calls.
- - We want the program to crash with a stack trace so developer 
-   could fix the bug.
+---
+
+Note: 
+Standard Exceptions (e.g. ValueError, TypeError) are intended for
+developer-facing errors. Use them when catching issues from internal
+code or function calls.
 '''
 
 class PayuError(Exception):
     '''
     Base class for all Payu Exceptions. 
     Any error that inherits from this class - represents an expected failure 
-    mode (ex. bad user input, missing files) rather than a code bug.
+    mode (e.g. bad user input, missing files) rather than a code bug.
     '''
     exit_code = 1
 
@@ -44,3 +45,4 @@ class PayuRunError(PayuError):
     Raised when an active model run fails unexpectedly.
     '''
     exit_code = 5
+ 
--- a/payu/errors.py
+++ b/payu/errors.py
@@ -35,7 +35,7 @@ class PayuFileNotFoundError(PayuError):
 
 class PayuRunError(PayuError):
     '''
-    Raised when an active model run fails unexpectedly.
+    Raised when an active model run fails.
     '''
     exit_code = 4
 
--- a/payu/errors.py
+++ b/payu/errors.py
@@ -33,16 +33,23 @@ class PayuFileNotFoundError(PayuError):
     exit_code = 3
 
 
-class PayuBranchError(PayuError):
-    '''
-    Raised when there are issues with Git branches.
-    '''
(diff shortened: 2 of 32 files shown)


payu/models/cice.py
```diff
@@ -97,26 +98,38 @@ def set_model_pathnames(self):
...
+                        ({path}) does not match input path 
+                        ({input_path})
+                        ''')
+                    # print('payu: error: Grid file path in {nmlfile} '
+                    #       '({path}) does not match input path '
+                    #       '({inputpath})'.format(
+                    #         nmlfile=self.ice_nml_fname,
+                    #         path=path,
+                    #         inputpath=input_path))
+                    # sys.exit(1)
```
KAUR1984, line 111: This commented code instances would need to be removed before merging. Have included them for now for reference only.

payu/errors.py
```diff
@@ -0,0 +1,55 @@
...
+
+Note: 
+Standard Exceptions (e.g. ValueError, TypeError) are intended for
+developer-facing errors. Use them when catching issues from internal
+code or function calls.
+'''
+
+class PayuError(Exception):
+    '''
+    Base class for all Payu Exceptions.
```
Qian-HuiChen:
```suggestion
    Base class for Payu Exceptions from user behavior. 
```

payu/errors.py
```diff
@@ -0,0 +1,55 @@
...
+class PayuRunError(PayuError):
+    '''
+    Raised when an active model run fails.
+    '''
+    exit_code = 4
+
+
+class PayuGitError(PayuError):
+    '''
+    Raised when there are Git related issues .
```
Qian-HuiChen:
```suggestion
    Raised when there are Git related issues.
```

payu/errors.py
```diff
@@ -0,0 +1,55 @@
...
+    '''
+    exit_code = 5
+
+
+class PayuBranchError(PayuError):
+    '''
+    Custom exception for payu branch operations
```
Qian-HuiChen: Could we add a bit more explanation on this branch error?

payu/models/mitgcm.py
```diff
@@ -156,11 +160,15 @@ def setup(self):
...
-                        '{niter}\nThis would overwrite previous '
-                        'pickups'.format(dt=dt, niter=(n_iter0 + n_timesteps)))
-                sys.exit(mesg)
+                niter = n_iter0 + n_timesteps
+                msg = (
+                    f'''
+                    payu : error: Timestep changed to {dt}. 
+                    Timestep at end identical to previous pickups: {niter}
+                    This would overwrite previous pickups
+                    ''')
```
Qian-HuiChen: I think this will result in printing a few spaces in front of the error message. Might be good to change it to
```
                msg =(f'''
payu: error 
::
''')
```

payu/models/mitgcm.py
```diff
@@ -174,10 +182,15 @@ def setup(self):
...
         if pchkpt_freq != dt * n_timesteps:
-            print('payu : error : time inconsistencies, '
-                  'pchkptfreq ({}) != experiment length ({})'
-                  ''.format(pchkpt_freq, dt * n_timesteps))
-            sys.exit(1)
+            raise errors.PayuRunError(
+                f'''
+                payu: error: time inconsistences, 
+                pchkptfreq ({pchkpt_freq}) != experiment length ({dt * n_timesteps})
+                ''')
```
Qian-HuiChen: Same here and a few other spots below. But please correct if you have tested it and don't see redundant space in front of the error message

payu/subcommands/clone_cmd.py
```diff
@@ -186,8 +188,9 @@ def fetch_tags(url):
         tags = [line.split('\t')[1].replace("refs/tags/", "") for line in result.stdout.splitlines()]
         return tags
     except subprocess.CalledProcessError as e:
-        print(f"Error fetching tags: {e}")
-        sys.exit(1)
+        raise errors.PayuBranchError(f'Error fetching branches: {e}') from e
```
Qian-HuiChen, line 219: Re line 174 and 191: I am not sure if this is PayuGitError or PayuBranchError. The error comes from running `git branch` and it will error when the repo not exist or internet not working. So maybe a PayuGitError in my head. Meanwhile, I don't have enough info to confirm what is `PayuBranchError` based on the comments in `error.py`.

payu/cli.py
```diff
@@ -59,8 +60,23 @@ def parse():
...
+        run_cmd(**args)
+
+    except errors.PayuError as e:
+        # CLEAN EXIT for expected errors
+        if stacktrace:
+            # Shows the full stacktrace
+            logging.exception(str(e), exc_info=True)
+        else:
+            print(f'payu: error: {e}', file=sys.stderr)
+        sys.exit(1)
```
Qian-HuiChen, line 70: Do we still need this sys.exit, since we already raise errors.PayuError?

payu/experiment.py
```diff
@@ -475,8 +476,10 @@ def setup(self, force_archive=False):
 
         # Confirm that no output path already exists
         if os.path.exists(self.output_path):
-            sys.exit('payu: error: Output path already exists: '
-                     '{path}.'.format(path=self.output_path))
+            raise errors.PayuError(f'payu: error: output path already exists:\
+                 {self.output_path}')
```
Qian-HuiChen, line 489: Should this be PayuError or PayuRunError?

payu/experiment.py
```diff
@@ -485,9 +488,15 @@ def setup(self, force_archive=False):
...
                 self.sweep()
             else:
-                sys.exit('payu: error: work path already exists: {path}.\n'
-                         '             payu sweep and then payu run'
-                         .format(path=self.work_path))
+                raise errors.PayuError(
+                    f'''
+                    payu: error: work path already exists: {self.work_path}.
+                             payu sweep and then payu run
+                    ''')
```
Qian-HuiChen: We classify this as PayuRunError in run_cmd.py so maybe good to keep it consistent as PayuRunError.

payu/experiment.py
```diff
@@ -902,7 +914,8 @@ def archive(self, force_prune_restarts=False):
 
         # Double-check that the run path does not exist
         if os.path.exists(self.output_path):
-            sys.exit('payu: error: Output path already exists.')
+            raise errors.PayuError('payu: error: output path already exists')
```
Qian-HuiChen: Maybe PayuRunError?

payu/manifest.py
```diff
@@ -122,14 +125,22 @@ def check_reproduce(self, previous_manifest):
...
-            )
-            print(f"Manifest path: stored hash != calculated hash")
-            for row in differences:
-                print(row)
-
-            sys.exit(1)
+            diff_text = '\n'.join(str(row) for row in differences)
+            raise errors.PayuRunError(
+                f'''
+                Run cannot reproduce: manifest {self.path} is not correct.
```
Qian-HuiChen:
```suggestion
payu: error: Run cannot reproduce: manifest {self.path} is not correct.
```
Just to keep consistent use of "payu: error" in the message for classified error type.

Qian-HuiChen (changes requested):
Wow that is a lot of work! Thank you for going through all the error in so many files :)
I left a few comments to help keep consistency across different errors. So excited to see the new error messages 🤩

jo-basevi:
Thanks @KAUR1984 for adding all of these updates!

> Currently, a bit unclear if it is better to use logging module or print statements for the catch-all exception handler - try-except block in cli.py. Would appreciate any thoughts on this. I remember Jo said something about this.

I think we should use the logging module for consistent formatting for errors and warnings. I think we had set up coloured formatting for logging if running in a interactive shell a few months ago. 

> All the custom exceptions come with exit_code. This could be useful to handle outputs based on just these codes, however, not sure if they are that useful in our case. Very happy to remove them otherwise

I'm not sure how useful they will be for handling outputs based on just these codes either. But it's also not adding any complexity leaving them in.

I noticed a few errors have format
```
                    raise errors.PayuFileNotFoundError(
                        f'''
                        payu: error: Input directory {rel_path} not found;
                        aborting.
                        '''
                     )
```
Using ''' preserves new lines and tabs/spaces 

The error above presents in my terminal as:
```
payu: error: 
                        payu: error: Input directory /scratch/$project/$user/mom6/input/not-a-real-file not found;
                        aborting.
```

For very long messages, I add a global variable at the top of the file, otherwise I use multiple strings, e.g. 
```
    print(
        "line 1\n"
         "line 2"
    )
```

I would also avoid the 'payu: error:' prefix if we use logging as it'll prefix an `ERROR:` already, and it'll avoid multiple 'payu: error' being added if errors are caught and re-raised. 

> Very happy to create another pull request for the rest of the changes in PR - https://github.com/payu-org/payu/pull/709. Was a bit unsure if this PR is already too big.

So the current cli.py changes only impact the commands that are typically run on the login node, e.g. `payu setup`, `payu clone`. It won't have any impact on the logs from payu runs on PBS jobs. My take is that it's OK if you want to keep this PR short but it would be good to have the other PR added not too long after so the changes are in the same payu release. 

To get the tests to run, you will need a small import fix to `branch.py`:
```
-from payu.git_utils import GitRepository, git_clone, PayuBranchError
+from payu.git_utils import GitRepository, git_clone
+from payu.errors import PayuBranchError
```

A few tests fail because the error type it is expecting is no longer raised, but that should be a small fix.

payu/subcommands/clone_cmd.py
KAUR1984: Thanks for raising this @Qian-HuiChen ! tbh, the distinction between the `PayuGitError` and `PayuBranchError` is a bit blurry and could overlap perhaps. `PayuBranchError` already existed previously, and to catch non-branch related errors, `PayuGitError` class could be useful (although very much down to keep only one in case the other one is not needed as much).

For this particular code line, not sure if `PayuGitError` would be ideal here as it is raised when there is an error in fetching branches - which sounds more like `PayuBranchError`. I have added a bit more explanation in doctstring of these classes. Happy to add more explanations in case that's still a bit unclear.

payu/cli.py
KAUR1984: Yes that was my understanding. Would love to have thoughts on this, `sys.exit()` in the main error handler would terminate the current terminal command. Without this, the command would look like it is successful although it is not.

payu/experiment.py
KAUR1984: Yes. I wasn't entirely sure if `PayuRunError` would fit here. That said, am happy to change it to `PayuRunError` is that's more suited.

payu/experiment.py
KAUR1984: Good point. Changed to `PayuRunError`.

payu/experiment.py
KAUR1984: Done

payu/models/mitgcm.py
KAUR1984: Great catch @Qian-HuiChen ! Fixed now. Didn't know these triple quote f-strings would add tab spaces as well. Have just replaced it with the following format to keep the indentation as is: 

```python
print(f'payu: error: This is my description  '
      f'and this is a continued...\n'
)
```

payu/manifest.py
KAUR1984: Thanks. Implemented.

payu/cli.py
Qian-HuiChen: I see, then it might be safer to have `sys.exit()` to make sure it exits the program.

Qian-HuiChen (commented):
The changes look good to me. Thanks for correcting all the string in the print out.

payu/models/model.py
```diff
@@ -139,8 +140,11 @@ def set_input_paths(self):
 
                     self.input_paths.append(rel_path)
                 else:
-                    sys.exit('payu: error: Input directory {0} not found; '
-                             'aborting.'.format(rel_path))
+                    raise errors.PayuFileNotFoundError(
+                        f'payu: error: Input directory {rel_path} not found; '
+                        'aborting.')
```
Qian-HuiChen, line 145: According to #749, we should probably update the error message to 
```
payu: error: Input {real_path} not found;
aborting
```

KAUR1984: Question - what did we mean by `real_path` here?

KAUR1984: This [comment](https://github.com/payu-org/payu/issues/749#issue-4536802330) also mentions raising an `OSError`, but I’m not sure what the best implementation approach would be. Any ideas? We could perhaps address this in a separate PR.

Qian-HuiChen: Apologies for my typo, I mean `{rel_path}`. We just need to remove the `directory` from the current error message.
I think keeping it as PayuFileNotFoundError is fine.

PR #729, opened 2026-05-01
ACCESS-OM3: Handling optional input files

### LINKED ISSUE #727: Handling optional input files in text format
In MOM6 (and therefore ACCESS-OM3), optional input files in text format can be configured in MOM_input. They then needed to be placed inside the payu `work/INPUT` directory.

There are [lines](https://github.com/ACCESS-NRI/access-om3-configs/blob/a030de3c7cd70cd16efb6db209f62ed087b3fdf4/docs/MOM_parameter_doc.all) to configure the folder name where the model expects to find the files:

```
INPUTDIR = "./INPUT/"           ! default = "."
                                ! The directory in which input files are found.
```

and then several possible input files, including:

```
DEPTH_LIST_FILE ...
```
and
```
CHANNEL_LIST_FILE = "MOM_channel_list" ! default = "MOM_channel_list"
                                ! The file from which the list of narrowed channels is read.
```

These files are only read when turned on in other parameters ( e.g `READ_DEPTH_LIST` and `CHANNEL_CONFIG`) . They can be required and not set in `MOM_input` if the default value for the filename is used.

In other cases input files are in a netcdf format, and so they are added to [`input:` section of config.yaml](https://github.com/ACCESS-NRI/access-om3-configs/blob/6b1a241c5f11a80af8dffaa826debac7deaba095/config.yaml#L34). However the `..._LINES` examples are text files, and so should be added to the git repository for a config. There are other optional config text files in the git repository, listed in [optional_config_files](https://github.com/payu-org/payu/blob/cbbc771c1d2dc86da93687ec01af88195ac7e4f3/payu/models/cesm_cmeps.py#L49) individually. However the current files in optional_config_files go into the `work` directory, rather than `work/INPUT`.

So the TL;DR is - we need a way of adding additional optional input files in text format to the ACCESS-OM3 driver. Some options:

1. add a step during payu setup to copy _optional input files_ into `work/<INPUTDIR>` where INPUTDIR is read from MOM_input, and the files are as many as we can find which might be in MOM_input (from reading the value of <CHANNEL_LIST_FILE>, <DEPTH_LIST_FILE> etc in MOM_input). This requires PAYU to know both what optional files can exist, and what their default values are. This is a bit problematic as defaults change in time, and inevitably, new optional input files will be added.
2. create an folder `INPUT` folder in the configuration, and copy the whole contents of this folder during setup. _payu_ then adds the netcdf input files (from `input:`) to this
3. treat these files as binary inputs and store them outside the repo

I think option 2 is the best one ?

anton-seaice:
Closes #727 

This allows configurations to have files in an `tables_lists` directory in their git repository, which are added to the work directory as extra config files

### DIFF
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -24,6 +24,7 @@ from payu.models.mom6 import mom6_add_parameter_files, mom6_save_docs_files
 
 NUOPC_CONFIG = "nuopc.runconfig"
 NUOPC_RUNSEQ = "nuopc.runseq"
+INPUT_DIR = 'INPUT'
 
 # mapping of runconfig to cftime calendars:
 # these match the supported calendars in CMEPS
@@ -132,8 +133,8 @@ class CesmCmeps(Model):
 
         super().set_model_pathnames()
 
-        self.work_input_path = os.path.join(self.work_path, 'INPUT')
-        
+        self.work_input_path = os.path.join(self.work_path, INPUT_DIR)
+
         # MOM restarts are dealt with via pointer files (see below). Use work_restart_path
         # for additional restarts (e.g. generic tracer flux restarts)
         self.get_runconfig(self.control_path)
@@ -144,6 +145,24 @@ class CesmCmeps(Model):
 
     def setup(self):
         super().setup()
+
+        # Special handling to also copy INPUT folder for optional input files
+        # e.g. mask_table, channel_list etc
+        src = os.path.join(self.control_path, INPUT_DIR)
+        if os.path.exists(src):
+            # Check for conflicting files before copying
+            for root, dirs, files in os.walk(src):
+                rel_path = os.path.relpath(root, src)
+                dst_path = os.path.join(self.work_input_path, rel_path)
+
+                for fname in files:
+                    dst_file = os.path.join(dst_path, fname)
+                    if os.path.exists(dst_file):
+                        raise FileExistsError(f"File already exists in INPUT dir: {dst_file}")
+
+            # Merge directories
+            shutil.copytree(src, self.work_input_path,
+                            symlinks=True, dirs_exist_ok=True)
 
         # Read components from nuopc.runconfig
         self.get_components()
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -24,7 +24,6 @@ from payu.models.mom6 import mom6_add_parameter_files, mom6_save_docs_files
 
 NUOPC_CONFIG = "nuopc.runconfig"
 NUOPC_RUNSEQ = "nuopc.runseq"
-INPUT_DIR = 'INPUT'
 
 # mapping of runconfig to cftime calendars:
 # these match the supported calendars in CMEPS
@@ -97,6 +96,9 @@ class CesmCmeps(Model):
             NUOPC_RUNSEQ
         ]
 
+        self.input_dir = 'INPUT'
+        self.extra_config_dir = 'masks_lists' #directory with any optional config files
+
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
         self.runconfig = None # nuopc.runconfig. Can't read this yet as paths haven't necessarily been set
         self.components = {} # To be read from nuopc.runconfig
@@ -133,7 +135,7 @@ class CesmCmeps(Model):
 
         super().set_model_pathnames()
 
-        self.work_input_path = os.path.join(self.work_path, INPUT_DIR)
+        self.work_input_path = os.path.join(self.work_path, self.input_dir)
 
         # MOM restarts are dealt with via pointer files (see below). Use work_restart_path
         # for additional restarts (e.g. generic tracer flux restarts)
@@ -146,23 +148,7 @@ class CesmCmeps(Model):
     def setup(self):
         super().setup()
 
-        # Special handling to also copy INPUT folder for optional input files
-        # e.g. mask_table, channel_list etc
-        src = os.path.join(self.control_path, INPUT_DIR)
-        if os.path.exists(src):
-            # Check for conflicting files before copying
-            for root, dirs, files in os.walk(src):
-                rel_path = os.path.relpath(root, src)
-                dst_path = os.path.join(self.work_input_path, rel_path)
-
-                for fname in files:
-                    dst_file = os.path.join(dst_path, fname)
-                    if os.path.exists(dst_file):
-                        raise FileExistsError(f"File already exists in INPUT dir: {dst_file}")
-
-            # Merge directories
-            shutil.copytree(src, self.work_input_path,
-                            symlinks=True, dirs_exist_ok=True)
+        self._setup_extra_config_files()
 
         # Read components from nuopc.runconfig
         self.get_components()
@@ -213,6 +199,14 @@ class CesmCmeps(Model):
             else:
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
+
+    def  _setup_extra_config_files(self):
+        # Special handling to also copy an extra folder for optional input files
+        # e.g. mask_table, channel_list etc
+        src = os.path.join(self.control_path, self.extra_config_dir)
+        if os.path.exists(src):
+            dest = os.path.join(self.work_path, self.extra_config_dir)
+            shutil.copytree(src, dest, symlinks=True)
 
     def _setup_checks(self):
         # check pelayout fits within requested cpucount
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -97,7 +97,7 @@ class CesmCmeps(Model):
         ]
 
         self.input_dir = 'INPUT'
-        self.extra_config_dir = 'masks_lists' #directory with any optional config files
+        self.extra_config_dir = 'tables_lists' #directory with any optional config files
 
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
         self.runconfig = None # nuopc.runconfig. Can't read this yet as paths haven't necessarily been set
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -97,7 +97,7 @@ class CesmCmeps(Model):
         ]
 
         self.input_dir = 'INPUT'
-        self.extra_config_dir = 'tables_lists' # directory with any optional config files
+        self.ancillary_input_dir = 'ancillary_input' # directory with extra input files in text format
 
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
         self.runconfig = None # nuopc.runconfig. Can't read this yet as paths haven't necessarily been set
@@ -148,7 +148,7 @@ class CesmCmeps(Model):
     def setup(self):
         super().setup()
 
-        self._setup_extra_config_files()
+        self._setup_ancillary_input_dir()
 
         # Read components from nuopc.runconfig
         self.get_components()
@@ -200,12 +200,12 @@ class CesmCmeps(Model):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
-    def _setup_extra_config_files(self):
+    def _setup_ancillary_input_dir(self):
         # Special handling to also copy an extra folder for optional input files
         # e.g. mask_table, channel_list etc
-        src = os.path.join(self.control_path, self.extra_config_dir)
+        src = os.path.join(self.control_path, self.ancillary_input_dir)
         if os.path.exists(src):
-            dest = os.path.join(self.work_path, self.extra_config_dir)
+            dest = os.path.join(self.work_path, self.ancillary_input_dir)
             shutil.copytree(src, dest, symlinks=True)
 
     def _setup_checks(self):
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -97,7 +97,7 @@ class CesmCmeps(Model):
         ]
 
         self.input_dir = 'INPUT'
-        self.extra_config_dir = 'tables_lists' #directory with any optional config files
+        self.extra_config_dir = 'tables_lists' # directory with any optional config files
 
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
         self.runconfig = None # nuopc.runconfig. Can't read this yet as paths haven't necessarily been set
@@ -200,7 +200,7 @@ class CesmCmeps(Model):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
-    def  _setup_extra_config_files(self):
+    def _setup_extra_config_files(self):
         # Special handling to also copy an extra folder for optional input files
         # e.g. mask_table, channel_list etc
         src = os.path.join(self.control_path, self.extra_config_dir)
--- a/payu/models/cesm_cmeps.py
+++ b/payu/models/cesm_cmeps.py
@@ -97,7 +97,7 @@ class CesmCmeps(Model):
         ]
 
         self.input_dir = 'INPUT'
-        self.ancillary_input_dir = 'ancillary_input' # directory with extra input files in text format
+        self.auxilary_input_dir = 'auxilary_input' # directory with extra input files in text format
 
         self.realms = ["ocn", "ice", "wav", "atm", "rof", "cpl"]
         self.runconfig = None # nuopc.runconfig. Can't read this yet as paths haven't necessarily been set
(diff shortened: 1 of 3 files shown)


payu/models/cesm_cmeps.py
```diff
@@ -145,6 +146,24 @@ def set_model_pathnames(self):
...
+
+                for fname in files:
+                    dst_file = os.path.join(dst_path, fname)
+                    if os.path.exists(dst_file):
+                        raise FileExistsError(f"File already exists in INPUT dir: {dst_file}")
+
+            # Merge directories
+            shutil.copytree(src, self.work_input_path,
+                            symlinks=True, dirs_exist_ok=True)
+
```
aidanheerdegen, line 152: This would result in these files being copied to the `outputXXX` directory on `archive`.

If you *had* to do this it would be better to symlink them from their default location in the model root dir.

jo-basevi:
Btw if you are wondering, the conda build failure is un-related to this PR change - I've created an issue here in the meantime: https://github.com/ACCESS-NRI/action-build-and-upload-conda-packages/issues/9

payu/models/cesm_cmeps.py
anton-seaice: This would mean there is zero tracking/provenance of these files in the output. We'd only want to symlink to files with a permanent location

anton-seaice:
I've confirmed that OM3 can be configured to us this - see 

https://github.com/ACCESS-NRI/access-om3-configs/pull/1325

and then the contents of the tables_lists folder are archved

anton-seaice:
ready for review @dougiesquire and someone from release team @aidanheerdegen

payu/models/cesm_cmeps.py
```diff
@@ -195,6 +200,14 @@ def setup(self):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
+    def  _setup_extra_config_files(self):
```
dougiesquire:
```suggestion
    def _setup_extra_config_files(self):
```

payu/models/cesm_cmeps.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, expt, name, config):
             NUOPC_RUNSEQ
         ]
 
+        self.input_dir = 'INPUT'
+        self.extra_config_dir = 'tables_lists' #directory with any optional config files
```
dougiesquire: For PEP8

```suggestion
        self.extra_config_dir = 'tables_lists'  # directory with any optional config files
```

test/models/access-om3/test_access_om3.py
```diff
@@ -351,6 +351,45 @@ def test__setup_checks_bad_io(pio_numiotasks, pio_stride):
...
+    cmeps_config(1)
+
+    with cd(ctrldir):
+        os.makedirs('tables_lists')
+
+        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
+        expt = payu.experiment.Experiment(lab, reproduce=False)
+        model = expt.models[0]
+
+
```
dougiesquire, line 373:
```suggestion

```

test/models/access-om3/test_access_om3.py
```diff
@@ -351,6 +351,45 @@ def test__setup_checks_bad_io(pio_numiotasks, pio_stride):
...
+        model._setup_extra_config_files()
+
+        assert os.path.isdir(os.path.join(model.work_path,model.extra_config_dir))
+
+        shutil.rmtree('tables_lists')
+        shutil.rmtree(os.path.join(model.work_path,model.extra_config_dir))
```
dougiesquire: If this `assert` fails, the `shutil.rmtree` calls won't be made. Could pollute other tests (e.g. the next one)?

dougiesquire (commented):
Thanks @anton-seaice.

I'm not a huge fan of the directory name `./tables_lists`. I don't have great alternatives, but if we wanted something specific, maybe:
- `input_tables`
- `mom_tables`

or more general:
- `input_extra`

test/models/access-om3/test_access_om3.py
anton-seaice: What the suggested way of handling this ?

anton-seaice:
> Thanks @anton-seaice.
> 
> I'm not a huge fan of the directory name `./tables_lists`. I don't have great alternatives, but if we wanted something specific, maybe:
> 
>     * `input_tables`
> 
>     * `mom_tables`
> 
> 
> or more general:
> 
>     * `input_extra`

I thought just a name with _extra_ wasn't adding much, like what makes it extra vs normal ?

The folder contains _text_input_files_ or _optional_configs_ but both names seemed a bit meaningless 

Maybe _optional_inputs_ is ok? But it gives the impression that they aren't essential, which is sort of true but deleting them would need other configuration changes and change results

anton-seaice:
we have a _datm_in_ , _ice_in_ etc files so how about _config_in_ directory

dougiesquire:
> I thought just a name with _extra_ wasn't adding much, like what makes it extra vs normal ?
> 
> The folder contains _text_input_files_ or _optional_configs_ but both names seemed a bit meaningless

I think all these names are meaningless because the subdirectory only exits for a Payu hack.

I don't like the word optional because they aren't really optional. At least no more so than some other inputs

> we have a _datm_in_ , _ice_in_ etc files so how about _config_in_ directory

I don't understand what config means here. The whole thing is a config.

> I thought just a name with _extra_ wasn't adding much, like what makes it extra vs normal ?

Ok, `input` and a README describing why it exists.

anton-seaice:
> Ok, `input` and a README describing why it exists.

`input` is already taken :)

test/models/access-om3/test_access_om3.py
aidanheerdegen: Tests shouldn't depend on other tests. Usual approach is to use a fixture to create and then remove test directories.

dougiesquire:
> > Ok, `input` and a README describing why it exists.
> 
> `input` is already taken :)

Well `INPUT` is already taken. But yes I guess that makes the work directory quite confusing. That's why I suggested `input_extra`

anton-seaice:
_input_tables_ is possibly the most descriptive we've come up with, except some of the files are called _lists_. Their contents could be described as [tables](https://github.com/ACCESS-NRI/access-om3-configs/blob/5f2339a39d69ffe80cd2c5f9367a259ca6406bea/tables_lists/MOM_channel_list) ?

aidanheerdegen:
> I think all these names are meaningless because the subdirectory only exits for a Payu hack.

You say hack, I say work-around for weirdly structured model.

dougiesquire:
> except some of the files are called _lists_

I don't see that as a big issue. For the files in question they're more or less synonyms I think. But it does become a little messy if we were to put other files in here. E.g. for OM3 single column config it would actually be quite convenient to include (very small) netcdf files in here.

aidanheerdegen:
Ping me for a review when you've finished squabbling about what to call stuff.

aidanheerdegen:
> OM3 single column config it would actually be quite convenient to include (very small) netcdf files in here

Oh boy. You give them an inch ... do we *really* want to do this?

anton-seaice:
I guess the non-hacky way would be for the `optional_config_files` to be configured in `config.yaml`, but would that be clear to user's they need to add the additional files there? (It's also not very within the payu paradigm). (And possibly no less clear than they need to put them in this arbitrary folder)

test/models/access-om3/test_access_om3.py
anton-seaice: Oh yes, this rings bells

dougiesquire:
> do we _really_ want to do this?

I do. There are reasons why it makes sense in this particular (unsupported) configuration. I just wanted to make the point that we may want to handle another files in the same way in the future so a name that is somewhat general could be good.

test/models/access-om3/test_access_om3.py
anton-seaice: I had a bit of scope creep trying to address this. I have fixed this for cases where the test setup doesn't rely on the module wide setup of directories. But not for all tests - I made an issue to explain:

https://github.com/payu-org/payu/issues/736

test/models/access-om3/test_access_om3.py
```diff
@@ -161,22 +216,16 @@ def test__setup_checks_npes(ncpu, moc_ntasks, moc_nthreads, moc_pestride, moc_ro
...
         "moc_rootpe": moc_rootpe
         })
+    model, expt = cmeps_model
 
-    with cd(ctrldir):
-        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
-        expt = payu.experiment.Experiment(lab, reproduce=False)
-        model = expt.models[0]
-
-        model.realms = ["moc"]
```
anton-seaice, line 170: the `cmeps_model` fixture replaces all the times that 

```
with cd(ctrldir):
        lab = payu.laboratory.Laboratory(lab_path=str(labdir))
        expt = payu.experiment.Experiment(lab, reproduce=False)
        model = expt.models[0]
```

was being used

anton-seaice:
Ready to go @dougiesquire @aidanheerdegen 

Sorry the test change is quite hard to follow due to whitespace changes, but should tidy things up a fair bit

test/models/access-om3/test_access_om3.py
```diff
@@ -49,25 +49,82 @@ def teardown_module(module):
...
+
+    yield cmeps_model
+
+    with cd(ctrldir):
+        shutil.rmtree(model.ancillary_input_dir)
+
+    try:
+        # delete the extra config dir from this modules work directory
+        shutil.rmtree(os.path.join(model.work_path,model.ancillary_input_dir))
+    except:
```
dougiesquire:
```suggestion
    except FileNotFoundError:
```

test/models/access-om3/test_access_om3.py
```diff
@@ -49,25 +49,82 @@ def teardown_module(module):
...
+        pass
+
+@pytest.fixture
+def cmeps_model_rest_dir(request, cmeps_model):
+    # Create archive/restart directory with rpointer.cpl file as well as normal cmeps_model
+    restart_path = make_expt_archive_dir(type='restart')
+    print(restart_path)
```
dougiesquire:
```suggestion
```

test/models/access-om3/test_access_om3.py
```diff
@@ -275,98 +307,91 @@ def test__setup_checks_bad_io(ncpu, pio_numiotasks, pio_stride, pio_root, pio_ty
...
-        ):
-            model._setup_checks()
+    with pytest.warns(
+        Warning, match="using model default"
+    ):
+        model._setup_checks()
 
-    teardown_cmeps_config()
 
+# test copy extra config files directory is copied
```
dougiesquire:
```suggestion
# test ancillary_input directory is copied
```

test/models/access-om3/test_access_om3.py
```diff
@@ -49,25 +49,82 @@ def teardown_module(module):
...
+    #Create a work directory with a med.log file as well as the normal cmeps_model
+
+    model, expt = cmeps_model
+
+    log_path = os.path.join(model.work_path, "log", "med.log")
+    os.makedirs(os.path.dirname(log_path), exist_ok=True)
+
+    # write param as contents of the file
+    with open(log_path, "w") as f:
+            f.write(request.param)
```
dougiesquire:
```suggestion
        f.write(request.param) 
```

test/models/access-om3/test_access_om3.py
dougiesquire, line 301: `ncpu` is no longer an input. This isn't throwing an error because I think this function is never run - there's a second `test__setup_checks_bad_io` function defined further down... (see [this comment](https://github.com/payu-org/payu/pull/729/changes#r3215678261))

test/models/access-om3/test_access_om3.py
```diff
@@ -275,98 +307,91 @@ def test__setup_checks_bad_io(ncpu, pio_numiotasks, pio_stride, pio_root, pio_ty
...
+        model._setup_checks()
 
+@pytest.mark.parametrize("cmeps_model",[1], indirect=['cmeps_model'])
 @pytest.mark.parametrize("pio_numiotasks, pio_stride", [
                          (1, -99),
                          (-99, 1),
                          ])
-def test__setup_checks_bad_io(pio_numiotasks, pio_stride):
-    cmeps_config(1)
+def test__setup_checks_bad_io(cmeps_model, pio_numiotasks, pio_stride):
```
dougiesquire: There are two `test__setup_checks_bad_io` test functions. I think the first definition never runs. We should either remove or rename one of them (see [this comment](https://github.com/payu-org/payu/pull/729/changes#r3215676964)).

dougiesquire (changes requested):
Thanks @anton-seaice, looking close to me

anton-seaice:
Try again @dougiesquire

dougiesquire (dismissed):
Thanks @anton-seaice LGTM

anton-seaice:
> Ping me for a review when you've finished squabbling about what to call stuff.

Ready to go @aidanheerdegen

payu/models/cesm_cmeps.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, expt, name, config):
             NUOPC_RUNSEQ
         ]
 
+        self.input_dir = 'INPUT'
+        self.ancillary_input_dir = 'ancillary_input' # directory with extra input files in text format
```
aidanheerdegen: Honestly so sorry, but ancillary has a very specific meaning in the UKMO/UM world, and I think would be confusing for that reason. 

How about `auxiliary_input`? 

Or could be more specific and used `tracked_inputs`, as they are inputs that we track with git.

payu/models/cesm_cmeps.py
```diff
@@ -195,6 +200,14 @@ def setup(self):
                 # TODO: copied this from other models. Surely we want to exit here or something
                 print('payu: error: Unable to find mod_def.ww3 file in input directory')
 
+    def _setup_ancillary_input_dir(self):
```
aidanheerdegen: I'd change the name of the function to be consistent with changes above.

payu/models/cesm_cmeps.py
aidanheerdegen: Is it the case that any OM3 configuration that wants to make use of this feature would need to have the matching name in the relevant model configuration files?

Does the driver code need to check for this?

dougiesquire: Oh I like `tracked_inputs`

anton-seaice: `tracked` could equally mean tracked by git or tracked by manifest, non ?

auxiliary is ok

anton-seaice: > Is it the case that any OM3 configuration that wants to make use of this feature would need to have the matching name in the relevant model configuration files?
> 
> Does the driver code need to check for this?

Is the suggestion to enforce that any text files referenced in the OM3 config with configurable filenames are in the `auxiliary_input` folder ? (payu doesn't do that with regular inputs)

aidanheerdegen: > Is the suggestion to enforce that any text files referenced in the OM3 config with configurable filenames are in the `auxiliary_input` folder ? (payu doesn't do that with regular inputs)

It's really just a consistency check. 

Some models have hard-coded `INPUT` and `RESTART` directories so a consistency check isn't required.

 [Others are configurable](https://github.com/ACCESS-NRI/cice5/blob/bd291fb3877ae68ed0de7742c6f087c85fc74c16/drivers/access/cpl_parameters.F90#L45), but in practice never change. 

You're introducing a new de-facto standard. As it's new I think it makes sense to check so users don't make a mistake, as it might not be something they are used to.

I actually think it would be a good thing to do consistency checks for all models that have configurable input directories.

aidanheerdegen: > `tracked` could equally mean tracked by git or tracked by manifest, non ?

`git_tracked_inputs` isn't as snappy, but it is arguably more precise.

anton-seaice: > > Is the suggestion to enforce that any text files referenced in the OM3 config with configurable filenames are in the `auxiliary_input` folder ? (payu doesn't do that with regular inputs)
> 
> It's really just a consistency check.
> 
> Some models have hard-coded `INPUT` and `RESTART` directories so a consistency check isn't required.
> 
> [Others are configurable](https://github.com/ACCESS-NRI/cice5/blob/bd291fb3877ae68ed0de7742c6f087c85fc74c16/drivers/access/cpl_parameters.F90#L45), but in practice never change.
> 
> You're introducing a new de-facto standard. As it's new I think it makes sense to check so users don't make a mistake, as it might not be something they are used to.
> 
> I actually think it would be a good thing to do consistency checks for all models that have configurable input directories.

Oh got it - there's no single place in the model config to check against.

aidanheerdegen: > Oh got it - there's no single place in the model config to check against.

Shame. Oh well. Forget that then.

aidanheerdegen:
I think we're waiting for a change to the `ancillary_input` dir and then all should be fine.

anton-seaice:
Ready to go @aidanheerdegen

aidanheerdegen (approved):
Normally I'd ask for the documentation to be updated, but we have no driver docs (yet) so you're off the hook!

I'm approving, but just to make sure, you have tested this works in a working config?

PR #734, opened 2026-05-07
Update usage of payu status in doc.

### LINKED ISSUE #733: Details on getting a log from payu status
I think the [docs](https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/docs/source/usage.rst#monitoring-payu-jobs) could cover using `watch` and a loop with `payu status`

e.g.

folks can use `watch` to get it to update automatically `watch -n 60 payu status`

or put it in a loop for a running log `while true ; do payu status ; sleep 60; done`

Qian-HuiChen:
Closes #733.

This PR updates the documentation of `payu status` , including 
- its usage with `watch`
- displayed queue time
- displayed current experiment time.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -298,6 +298,14 @@ To monitor the status of running and finished payu run jobs, run::
 
    payu status
 
+To keep updated with the latest status, use `watch`::
+
+   watch -n {refresh_interval} payu status
+
+e.g., to refresh every 30 seconds::
+
+   watch -n 30 payu status
+
 By default, this displays information about the latest run number.
 This includes:
 
@@ -315,6 +323,19 @@ This includes:
    * ``model-run`` - Model is running
 
    * ``archive`` - Model run has finished, and payu is archiving the output
+
+* Total queue time for running or completed jobs. 
+  To update the current queue time for a queuing job, use::
+
+   payu status --update
+
+* Current model time for running experiments or finished model time for completed experiments.
+  This feature is implemented in models: 
+   * ACCESS-OM2
+   * ACCESS-OM3
+   * ACCESS-ESM1.5
+   * ACCESS-ESM1.6
+   * MOM6
 
 * Exit status of the payu run (if available). This is set at the end of a payu
   run so may not reflect the exit status of the scheduler job, e.g., if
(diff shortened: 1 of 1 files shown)


docs/source/usage.rst
```diff
@@ -298,6 +298,14 @@ To monitor the status of running and finished payu run jobs, run::
 
    payu status
 
+To keep updated with the latest status, use `watch`::
+
+   watch -n {refresh_interval} payu status
+
+e.g., to refresh every 30 seconds::
```
anton-seaice: In https://opus.nci.org.au/spaces/Help/pages/236880322/Job+monitoring...

they request only querying job status every 10 minutes !

Which is kind of impractical, but we should make some note about not setting the interval too small (is 60 seconds a good minimum ?)

anton-seaice:
I think we should also inlcude

while true ; do payu status ; sleep 60; done

I think having a log is useful to see that something has changed since I last looked ?

docs/source/usage.rst
jo-basevi: `payu status` is all good to query more often as it's relying on payu updating files.
While `payu status --update` is getting updated scheduler information from `qstat`. 

So would be good to have a note of 60 seconds minimum when using `payu status --update`.

Qian-HuiChen: I think `payu status` [reads out the information in the job file](https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/status.py#L150-L157), rather than calling `qstat` each time. The job file gets updated when the payu status changes (e.g., [entering run](https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/experiment.py#L713-L719) or [entering archive](https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/experiment.py#L872-L882))

One note is that `payu status --update` does call `qstat` each time in order to refresh the job file.

---Edit---
Please also see Jo's comment above

anton-seaice (dismissed):
Thanks @Qian-HuiChen 

Looks good to me

The formatting is a bit unclear maybe ? But that's mroe the template than what's written

Image

Qian-HuiChen:
@anton-seaice I tried to twist the formatting a little bit to make it easier to read.
image

jo-basevi (approved):
Thanks @Qian-HuiChen, the changes look good to me

anton-seaice (approved):

PR #738, opened 2026-05-13
Pytest fixture to create and clean tmp directory in unit test

Qian-HuiChen:
Use a fixture (`setup_test_dir`) to create tmp directories and ensure clean up after test for model drivers. This guarantees test isolation and ensures that all temporary directories are removed regardless of whether tests pass or fail.

Closes #736.

Issue #232, opened 2020-02-26
use more than 3 digits for outputnnn and restartnnn?


aekiss:
The output and restart directories are distinguished by a 3-digit run number, but we are getting uncomfortably close to running out of digits in the long 0.1 deg ACCESS-OM2 runs (we would run out of digits after 250 years at 3mo/submit). So I'm wondering how hard-coded that is, and whether it could easily be increased?

aekiss:
e.g. @andyhoggANU's run `/g/data/ik11/outputs/access-om2-01/01deg_jra55v13_ryf9091` is up to `output688` so far...

marshallward:
Has anyone actually tried this before?  I recall it will output more digits if needed, it will just lack the padding.
```
>>> counter = 1000
>>> restart_dir = 'restart{0:03}'.format(counter)
>>> print(restart_dir)
restart1000
```

aekiss:
Some workflows I've created rely on alphabetic sorting, so the lack of zero-padding would be a problem for that.

But if payu is adjusted to use 4 digits it would need to initially be optional so it doesn't mess up the naming in existing runs. 

Maybe it's not worth worrying about, since we are yet to even exceed this 3-digit limit?

marshallward:
It should be a quick job to replace the hard-coded 3 with a variable whose default value is 3.

marshallward:
Another solution is to allow the user to define their own format string, with `0:03` being the default (or even `restart{0:03}` depending on how ambitious we are feeling...)

aidanheerdegen:
I noticed Adele had a run with run numbers > 999 and it seemed to work fine apart from `ls` not sorting correctly without using an option like `-v` or `-t`
```
$ls -v /scratch/v45/akm157/access-om2/archive/01deg_jra55v13_ryf9091_weddell_up1
error_logs  restart998   restart1015  restart1032  restart1049
output995   restart999   restart1016  restart1033  restart1050
output1055  restart1000  restart1017  restart1034  restart1051
output1056  restart1001  restart1018  restart1035  restart1052
output1057  restart1002  restart1019  restart1036  restart1053
output1058  restart1003  restart1020  restart1037  restart1054
output1059  restart1004  restart1021  restart1038  restart1055
output1060  restart1005  restart1022  restart1039  restart1056
output1061  restart1006  restart1023  restart1040  restart1057
output1062  restart1007  restart1024  restart1041  restart1058
output1063  restart1008  restart1025  restart1042  restart1059
[...]

PR #740, opened 2026-05-15
Pass current run number to payu sync command.

### LINKED ISSUE #732: Automated payu sync command should pass along current run number
Currently payu sync is invoked here:

https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/experiment.py#L998-L1001

For reference, collate passes along the current run number: 

https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/experiment.py#L947-L951

I think the payu sync job currently syncs all outputs which could be an issue for jobs run in series (e.g. if a later payu run archives before an earlier sync job runs).

I think if the current run number is provided, it shouldn't sync any outputs from runs greater than current run number.

Qian-HuiChen:
This PR pass the current run number to `payu sync` command. In this case, restart and output greater than the current run will not be sync.

Close #732.

### DIFF
--- a/payu/sync.py
+++ b/payu/sync.py
@@ -16,7 +16,7 @@ import time
 
 
 # Local
-from payu.fsops import list_archive_dirs
+from payu.fsops import list_sorted_archive_dirs
 from payu.metadata import METADATA_FILENAME, UUID_FIELD
 
 DEST_NOT_CONFIGURED_MSG ="""
@@ -55,15 +55,12 @@ def filter_previous_runs(all_dir, prefix):
               "Syncing all runs in archive.")
         return all_dir
 
-    # Sort the dicectories from lowest to highest based on the suffix number
-    sorted_dir = sorted(all_dir, key=lambda d: int(d.removeprefix(prefix)))
-
-    for i in range(len(sorted_dir)-1, -1, -1):
-        suffix = sorted_dir[i].removeprefix(prefix)
+    for i in range(len(all_dir)-1, -1, -1):
+        suffix = all_dir[i].removeprefix(prefix)
 
         # Only include suffix that is less than or equal to the current run
         if int(suffix) <= int(current_run):
-            return sorted_dir[0:i+1]
+            return all_dir[0:i+1]
 
     # Return empty list if no directories are <= current run
     return []
@@ -87,7 +84,7 @@ class SyncToRemoteArchive():
     def add_outputs_to_sync(self):
         """Add paths of outputs in archive to sync. The last output is
         protected"""
-        all_outputs = list_archive_dirs(archive_path=self.expt.archive_path,
+        all_outputs = list_sorted_archive_dirs(archive_path=self.expt.archive_path,
                                     dir_type='output')
         outputs = filter_previous_runs(all_outputs, prefix='output')
         outputs = [os.path.join(self.expt.archive_path, output)
@@ -111,7 +108,7 @@ class SyncToRemoteArchive():
             return
 
         # Get sorted list of restarts in archive
-        all_restarts = list_archive_dirs(archive_path=self.expt.archive_path,
+        all_restarts = list_sorted_archive_dirs(archive_path=self.expt.archive_path,
                                      dir_type='restart')
         restarts = filter_previous_runs(all_restarts, prefix='restart')
         restarts = [os.path.join(self.expt.archive_path, restart)
--- a/payu/sync.py
+++ b/payu/sync.py
@@ -5,7 +5,7 @@
 """
 
 # Standard
-import getpass
+import warnings
 import glob
 import os
 import shutil
@@ -42,6 +42,25 @@ class SourcePath():
         self.path = path
         self.is_log_file = is_log_file
 
+def filter_previous_runs(all_dir, prefix=None):
+    """Given a list of directories of all runs (e.g., ['output001', 'output002']), 
+    filter to only include dir that is less than or equal to current run."""
+    if all_dir == []:
+        return []
+
+    current_run = os.environ.get('PAYU_CURRENT_RUN')
+    if current_run is None:
+        warnings.warn("PAYU_CURRENT_RUN environment variable not set. "
+              "Syncing all runs in archive.")
+        return all_dir
+
+    filter_dir = []
+    for directory in all_dir:
+        suffix = directory.replace(prefix, '') if prefix else directory
+        if int(suffix) <= int(current_run):
+            filter_dir.append(directory)
+
+    return filter_dir
 
 class SyncToRemoteArchive():
     """Class used for archiving experiment outputs to a remote directory"""
@@ -62,8 +81,9 @@ class SyncToRemoteArchive():
     def add_outputs_to_sync(self):
         """Add paths of outputs in archive to sync. The last output is
         protected"""
-        outputs = list_archive_dirs(archive_path=self.expt.archive_path,
+        all_outputs = list_archive_dirs(archive_path=self.expt.archive_path,
                                     dir_type='output')
+        outputs = filter_previous_runs(all_outputs, prefix='output')
         outputs = [os.path.join(self.expt.archive_path, output)
                    for output in outputs]
         if len(outputs) > 0:
@@ -85,8 +105,9 @@ class SyncToRemoteArchive():
             return
 
         # Get sorted list of restarts in archive
-        restarts = list_archive_dirs(archive_path=self.expt.archive_path,
+        all_restarts = list_archive_dirs(archive_path=self.expt.archive_path,
                                      dir_type='restart')
+        restarts = filter_previous_runs(all_restarts, prefix='restart')
         restarts = [os.path.join(self.expt.archive_path, restart)
                     for restart in restarts]
         if restarts == []:
--- a/payu/sync.py
+++ b/payu/sync.py
@@ -12,6 +12,7 @@ import shutil
 import subprocess
 from ruamel.yaml import YAML
 from pathlib import Path
+import time
 
 
 # Local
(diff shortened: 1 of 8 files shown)


Qian-HuiChen:
I’m not sure what is the best way to handle the payu_jobs directory gracefully.
At the moment, the entire payu_jobs directory is synced without filtering out jobs greater than the current run. 
However, if I do something like:
```
    def add_payu_jobs_to_sync(self):
        """Add paths of payu jobs in archive to sync. The last job is
        protected"""
        log_path = os.path.join(self.expt.archive_path, "payu_jobs")
        all_jobs = sorted(os.listdir(log_path)) if os.path.isdir(log_path) else []
        jobs = filter_previous_runs(all_jobs, prefix=None)
        jobs = [os.path.join(log_path, job) 
                for job in jobs]
        if len(jobs) > 0:
            last_job = jobs.pop()
            if not self.ignore_last:
                # Protect the last job
                self.source_paths.append(SourcePath(path=last_job,
                                                    protected=True))
        self.source_paths.extend([SourcePath(path, is_log_file=True) for path in jobs])
```
then the jobs get synced as `sync_path/0`, `sync_path/1`, etc., instead of `sync_path/payu_jobs/0`.

I can think of two ways:
1. changing the destination path for `payu_jobs`
2. sync the entire `payu_jobs` dir and set `--exclude` flag to exclude jobs greater than current run.

The tests in test_sync.py are not independent for now. I will improve them after #738 is merged.

aidanheerdegen:
> I’m not sure what is the best way to handle the payu_jobs directory gracefully. At the moment, the entire payu_jobs directory is synced without filtering out jobs greater than the current run. However, if I do something like:

Do we need to handle `payu_jobs` in this way? Isn't it ok to just sync over the current state of `payu_jobs`?

payu/sync.py
```diff
@@ -42,6 +42,25 @@ def __init__(self, path, protected=False, is_log_file=False):
         self.path = path
         self.is_log_file = is_log_file
 
+def filter_previous_runs(all_dir, prefix=None):
+    """Given a list of directories of all runs (e.g., ['output001', 'output002']), 
+    filter to only include dir that is less than or equal to current run."""
```
aidanheerdegen, line 48: I don't know if it's better/faster, but you could sort the directories and work backwards from the end (most recent/highest number) until you get a match that is `<= current_run`.

Sort touches all the elements, but it is fast. This way the `replace` and `int` comparison is done on far fewer elements. 

Just a thought.

Qian-HuiChen:
I see. It should be fine to sync all `payu_jobs` at the current state. `restart` and `output` are the focus here.

payu/sync.py
Qian-HuiChen: Sounds good. Implemented in [b304178](https://github.com/payu-org/payu/pull/740/commits/b304178b2b25bf4986f239317c7a78a1ddc628d5). I tried to test the time needed in each case, but both of them returns 0.00s 😆

test/test_sync.py
```diff
@@ -83,6 +86,32 @@ def assert_expected_archive_paths(source_paths,
     assert protected_dirs == expected_protected_dirs
 
 
+def test_filter_previous_runs(monkeypatch):
+    """Test filter_previous_runs pick up runs <= the current run."""
+
+    # Set current run to 3
+    monkeypatch.setenv("PAYU_CURRENT_RUN", "3")
+
+    all_dirs = ['output001', 'output002', 'output003', 'output004', 'output005']
```
aidanheerdegen: We [need to support the case for run numbers > 999](https://github.com/payu-org/payu/issues/232). Can you include examples with 4 digits on the end and unordered to make sure this still sorts correctly?

e.g.
```suggestion
    all_dirs = ['output1001', 'output997', 'output999', 'output1002', 'output998']
```

Which would require changing `PAYU_CURRENT_RUN` above.

test/test_sync.py
```diff
@@ -83,6 +86,32 @@ def assert_expected_archive_paths(source_paths,
...
+    
+    assert result == expected
+
+
+def test_filter_previous_runs_no_current_run(monkeypatch):
+    """Test filter_previous_runs returns all dirs when PAYU_CURRENT_RUN is not set."""
+    
+    monkeypatch.delenv("PAYU_CURRENT_RUN", raising=False)
+
+    all_dirs = ['output001', 'output002', 'output003', 'output004', 'output005']
```
aidanheerdegen: If changed above might make sense to do the same here.

payu/sync.py
aidanheerdegen: You can use [`timeit`](https://docs.python.org/3.13/library/timeit.html) to get reliable stats on relatively fast functions. Also it would be very dependent on the number of elements.

aidanheerdegen (commented):
Just a suggestion about testing.

payu/sync.py
```diff
@@ -42,6 +43,30 @@ def __init__(self, path, protected=False, is_log_file=False):
...
+
+    # Sort the dicectories from highest to lowest
+    sorted_dir = sorted(all_dir)[::-1]
+    for i in range(len(sorted_dir)):
+        directory = sorted_dir[i]
+        suffix = directory.replace(prefix, '') if prefix else directory
+
+        # Only include directories that are less than or equal to the current run
+        if int(suffix) <= int(current_run):
+            return sorted_dir[i:][::-1]
```
jo-basevi: So `list_archive_dirs` already sorts the directories based on the numbers:

https://github.com/payu-org/payu/blob/27624d9d3ad07b04099068c1f40590a6d639e771/payu/fsops.py#L242-L257

A very optional suggestion (feel free to ignore this!), is to loop backwards rather than using list slices to reverse order. E.g. 

```python
for i in range(len(all_dir) - 1, -1, -1):
    directory = all_dir[i]
    suffix = directory.replace(prefix, '') if prefix else directory
    if int(suffix) <= int(current_run):
        return all_dir[:i + 1]
```

I don't think prefix ever would be None - so I would make it a required argument. 

Could use `int(d.removeprefix(prefix)` too. Today I learnt that the `list_archive_dirs` should use `removeprefix` rather than `lstrip` as `lstrip` uses removes any letters matching the characters in the string.. E.g. 

```
>>> x = "outputput111"
>>> x.removeprefix("output")
'put111'
>>> x = "outputput111"
>>> x.lstrip("output")
'111'
```

jo-basevi (commented):
Thanks for adding this fix in! Changes look good to me

payu/sync.py
Qian-HuiChen: Thanks for pointing out `removeprefix`! It looks very convenient. Implemented in the latest commit. Also change `list_archive_dir` with `removeprefix` and add a unit test for it.

test/test_sync.py
Qian-HuiChen: Thanks for pointing this out! It actually finds a bug:
```
sorted_dir = sorted(all_dir)
```
thinks ['output1001', 'output1002', 'output997', 'output998', 'output999'] is sorted.
I updated this line to
```
sorted_dir = sorted(all_dir, key=lambda d: int(d.removeprefix(prefix)))
```
then it knows to sort it out as ['output997', 'output998', 'output999', 'output1001', 'output1002']

test/test_sync.py
Qian-HuiChen: Implemented in the commit [a78b669](https://github.com/payu-org/payu/pull/740/commits/a78b669ac0f4e10b1ce302498d0431ccecb1d1ea).

payu/sync.py
jo-basevi: There's a unit test for `test_list_archive_dirs` in test_payu.py so might be worth removing that one and keep the new one on test_fsops.py - but maybe check it wasn't testing any other edge cases?

jo-basevi: Sorry this is code I wrote a while ago but `list_archive_dirs` -> `list_sorted_archive_dirs` is probably a more appropriate name since the sorted is essential to how its used in other code in payu..

Qian-HuiChen: Oh easy, I will move the `test_list_archive_dirs` from `test_payu.py` to `test_fsops.py`. I will add a few dirs into the test for edge cases. 
I will also change the name to `list_sorted_archive_dirs`.

See commit [9d73ddb](https://github.com/payu-org/payu/pull/740/commits/9d73ddb0825b1f13be79f230446b82ed3f6f5df4)

jo-basevi: Thanks!

test/test_sync.py
Qian-HuiChen: Because @jo-basevi pointed it out that `list_archive_dir` already sorts out the directories, this should not be a problem for `filter_previous_runs`.  In the `test_filter_previous_runs`, I only put in sorted dirs for testing. The sorting process is tested in `test_list_sorted_archive_dirs` separately.

aidanheerdegen (approved):
Looks good. Thanks!

payu/sync.py
```diff
@@ -42,6 +43,27 @@ def __init__(self, path, protected=False, is_log_file=False):
...
 
+def filter_previous_runs(all_dir, prefix):
+    """Given a list of directories of all runs (e.g., ['output001', 'output002']), 
+    filter to only include dir that is less than or equal to current run."""
+    if all_dir == []:
+        return []
+
+    current_run = os.environ.get('PAYU_CURRENT_RUN')
+    if current_run is None:
+        warnings.warn("PAYU_CURRENT_RUN environment variable not set. "
```
jo-basevi, line 54: This warning is not being displayed for me.. I've verified that that this code is reached in my test so I don't think it's an issue with this PR. 

Are warnings being displayed for you?

jo-basevi (approved):
Thanks @Qian-HuiChen! The changes in this PR look good to me

payu/sync.py
jo-basevi: Ok I've confirmed that this issue is unrelated to this PR (see comment: https://github.com/payu-org/payu/pull/713#discussion_r3285098967) so shouldn't block this PR from being merged.

PR #480, opened 2024-08-08
441 inconsistency between yaml libs bugfix

jo-basevi:
As suggested in this issue #441,
- Add warnings when reading config.yaml files with duplicate keys (as PyYAML library silently over-writes the value with the subsequent value) 
- When adding restarts to config.yaml when creating a new git branch, allow duplicate keys but still warn that they will get deleted when the config.yaml file gets modified (note the subsequent key will get deleted). Ruamel yaml library is used for modifying the config files as it can preserve multi-line strings and comments.
- Added a test for testing the warnings above

Closes #441

By default warnings are formatted with a line of the source code, e.g. the line `warnings.warn(` in
```
/home/189/jb4202/payu_fork/payu/fsops.py:83: UserWarning: Duplicate key found in config.yaml: key 'model' with value 'mom6' (original value: '')
  warnings.warn(
```
In fb45911983e399befc52e03feb9394a74c796f93, I have just patched the warnings used globally it doesn't include the line of source code. This is to avoid logs with like `warnings.warn(` with unclosed brackets.

Issue #441, opened 2024-05-10
Inconsistency between yaml libraries


aidanheerdegen:
It seems ruamel yaml is pickier than PyYAML, because I go this error

```
$ payu checkout -r archive/restart100 -b 1xCO2 e12bda105f3
Created and checked out new branch: 1xCO2                                                                                                    
laboratory path:  /scratch/lg87/aph502/access-esm
binary path:  /scratch/lg87/aph502/access-esm/bin
input path:  /scratch/lg87/aph502/access-esm/input
work path:  /scratch/lg87/aph502/access-esm/work
archive path:  /scratch/lg87/aph502/access-esm/archive
Updated metadata. Experiment UUID: cc4197a5-9df7-410c-9aa9-a60c5593c5d2
Traceback (most recent call last):
  File "/g/data/vk83/apps/payu/1.1.3/bin/payu", line 10, in <module>
    sys.exit(parse())
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/payu/cli.py", line 41, in parse
    run_cmd(**args)
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/payu/subcommands/checkout_cmd.py", line 34, in runcmd
    checkout_branch(is_new_branch=new_branch,
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/payu/branch.py", line 175, in checkout_branch
    add_restart_to_config(prior_restart_path, config_path=config_path)
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/payu/branch.py", line 71, in add_restart_to_config
    config = yaml.load(config_path)
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/ruamel/yaml/main.py", line 446, in load
    return self.load(fp)
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/ruamel/yaml/main.py", line 451, in load
    return constructor.get_single_data()
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/ruamel/yaml/constructor.py", line 116, in get_single_data
    return self.construct_document(node)
  File "/g/data/vk83/apps/payu/1.1.3/lib/python3.9/site-packages/ruamel/yaml/constructor.py", line 125, in construct_document
    for _dummy in generator:
[...]

PR #742, opened 2026-05-20
Replace `PyYAML` with `Ruamel.yaml` across payu

Qian-HuiChen:
This PR adapts `ruamel.yaml` across payu, replacing `PyYAML`.

`ruamel.yaml` reserves the formatting and comments in yaml files. By default, it does not allow duplicate keys. To handle duplicate keys in config files, payu use `allow_duplicate_keys == True` when reading configs.

Ref #441 and #480.

### DIFF
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -23,7 +23,8 @@ import stat
 import warnings
 
 # Extensions
-import yaml
+from ruamel.yaml import YAML
+from ruamel.yaml.constructor import DuplicateKeyError
 
 DEFAULT_CONFIG_FNAME = 'config.yaml'
 
@@ -36,6 +37,11 @@ EXTENSION_TO_INTERPRETER = {'.py': sys.executable,
                             '.sh': '/bin/bash',
                             '.csh': '/bin/tcsh'}
 
+duplicate_key_warning = """
+    ---------------------------
+    Duplicate key found in config file.
+    Payu will keep the first value for this key.
+    --------------------------- """
 
 def movetree(src, dst, symlinks=False):
     """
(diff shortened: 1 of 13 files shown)


payu/models/cable.py
```diff
@@ -113,7 +115,7 @@ def setup(self):
         self.cable_nml = f90nml.read(self.cable_nml_path)
         if self.prior_restart_path:
             with open(self.restart_calendar_path, 'r') as restart_file:
-                self.restart_info = yaml.safe_load(restart_file)
+                self.restart_info = yaml.load(restart_file)
```
aidanheerdegen, line 118: To save anyone else looking, ruamel defaults to `typ=rt`, which is round-trip, so comment preserving, and doesn’t execute code by default.

So `safe_load` is not required.

payu/fsops.py
```diff
@@ -85,7 +71,22 @@ def read_config(config_fname=None):
...
+                yaml.allow_duplicate_keys = False
+                config = yaml.load(config_file)
+
+            except DuplicateKeyError as e:
+                # Warn the user about duplicated keys,
+                # Second attempt to load config with duplicated key allowed.
+                warnings.warn("Details: " + str(e) + duplicate_key_warning, UserWarning)
+                yaml = YAML(typ='safe')
+                config_file.seek(0)
+                yaml.allow_duplicate_keys = True
```
aidanheerdegen, line 87: Does this configuration setting persist in subsequent invocation? 

Are the settings like this localised to the file where it is imported?

aidanheerdegen (dismissed):
Looks good to me.

payu/fsops.py
Qian-HuiChen: Good question. `yaml.allow_duplicate_keys = True` is applied locally to line 85 `yaml = YAML(typ='safe')` inside function `read_config()`. Each call of `read_config()` will create a new `yaml = YAML(typ='safe')` freshly.

aidanheerdegen: > Each call of `read_config()` will create a new `yaml = YAML(typ='safe')` freshly.

Ah of course, it is only set on the specific yaml object. Sorry, should have noticed that.

aidanheerdegen (approved):

PR #743, opened 2026-05-20
When payu git commit fail, print a more informative message

### LINKED ISSUE #737: Payu git commits fail when control directory is repository sub-directory
Currently payu commits fail if control directory is a sub-directory in a repository. 

Places where they fail is when its using the `GitRepository` class in `git_utils.py` (so when committing new metadata.yaml files or [mom6 parameters files](https://github.com/payu-org/payu/blob/6ee5504974ff581d23ad82f601734900aa5e9ed3/payu/models/mom6.py#L65-L67)).

https://github.com/payu-org/payu/blob/6ee5504974ff581d23ad82f601734900aa5e9ed3/payu/git_utils.py#L47-L48

The git python repository is initialised here: 

https://github.com/payu-org/payu/blob/6ee5504974ff581d23ad82f601734900aa5e9ed3/payu/git_utils.py#L30

From a very quick test, I think it can work if we change `repo = git.Repo(repo_path)` -> `repo = git.Repo(repo_path, search_parent_directories=True)`, and set the base repository path value in `GitRepository` class (or replace it) to `self.repo_path = repo.working_tree_dir`.

If there is multiple control directories as sub-directories in one repository branch, the run log will have multiple commits related to different experiments. So it's not the typical payu workflow.

Qian-HuiChen:
This PR checks whether the current directory is a subdirectory of a Git repository. Payu now prints a more informative message when payu git commit fails.

Closes #737 .


payu/models/mom6.py
```diff
@@ -64,7 +64,7 @@ def mom6_save_docs_files(model):
 
     if model.expt.runlog.enabled: #if runlog true, default to true
         # commit new files to the control dir
-        repo = GitRepository(repo_path = model.control_path)
+        repo = GitRepository(repo_path = model.control_path, catch_error = True)
```
aidanheerdegen: With `catch_error = True` won't you need to check the return status and deal with it if `repo is None`?

payu/branch.py
```diff
@@ -185,7 +185,7 @@ def checkout_branch(branch_name: str,
         control_path = get_control_path(config_path)
 
     # Checkout branch
-    repo = GitRepository(control_path)
+    repo = GitRepository(control_path, catch_error=True)
```
aidanheerdegen, line 188: With `catch_error = True` won't you need to check the return status and deal with it if `repo is None`?

payu/experiment.py
```diff
@@ -565,6 +566,12 @@ def setup(self, force_archive=False):
 
     @timeit("payu_run_duration_seconds")
     def run(self, *user_flags):
+        # Error out if runlog is enables and running from subdirectory of git root repository
```
```suggestion
        # Error out if runlog is enabled and running from subdirectory of git root repository
```
Typo

payu/git_utils.py
```diff
@@ -39,6 +50,16 @@ def get_git_repository(repo_path: Union[Path, str],
...
         )
+
+        parent_repo = check_git_parent(repo_path)
+        if parent_repo:
+            warnings.warn(
+                f"Payu expects to run from repository root to ensure metadata tracking.\n"
+                f"Current situation leads to an InvalidGitRepositoryError.\n"
+                f"Suggest fix:\n    cd {parent_repo.working_tree_dir}\n    and run your command again.",
+                PayuGitWarning
+            )
```
aidanheerdegen, line 52:
```suggestion
        parent_repo = check_git_parent(repo_path)
        if parent_repo:
            warnings.warn(
                f"Payu expects to run from repository root to ensure metadata tracking.\n"
                f"Current situation leads to an InvalidGitRepositoryError.\n"
                f"Suggest fix:\n    cd {parent_repo.working_tree_dir}\n    and run your command again.",
                PayuGitWarning
            )
        else:
            if initialise:
                repo = git.Repo.init(repo_path)
                print(f"Initialised new git repository at: {repo_path}")
                return repo

            warnings.warn(
                f"Path is not a valid git repository: {repo_path}",
                PayuGitWarning
            )   
```
I suggest moving the initialise code block so that it only runs if there is no parent git repo.

I don't know if the suggested fix will work in this case: there is a valid parent git repo, but aren't we saying we don't support having `runlog` turned on in this case? So I think the suggested fix is to either move the control directory into a separate directory with it's own git repo, or set `runlog: False`.

payu/runlog.py
```diff
@@ -71,47 +73,31 @@ def create_manifest(self):
...
+                git_repo.repo.index.commit(m=commit_msg, no_gpg_sign=True)
+                warnings.warn("Runlog commit without gpg signing.")
+            except git.exc.GitCommandError as e:
+                print(f"payu: error: Failed to commit runlog changes to git repository: {e}.")
 
         # Save the commit hash
-        self.expt.run_id = commit_hash(self.expt.control_path)
-
-        f_null.close()
+        self.expt.run_id = git_repo.repo.head.object.hexsha
```
aidanheerdegen: This whole refactoring is very nice and clean!

payu/runlog.py
```diff
@@ -139,12 +125,18 @@ def push(self):
         if not os.path.isfile(ssh_key_path):
             print('payu: error: Github SSH key {key} not found.'
                   ''.format(key=ssh_key_path))
-            print('payu: error: Run `payu ghsetup` to generate a new key.')
+            print('payu: error: Please generate a new key, e.g., by running `gh auth login`.')
             sys.exit(-1)
```
aidanheerdegen: If a user has set up authentication it will work. If they haven't, it won't. I don't think we should be bothering with trying to guess ssh keys, especially as `gh auth` uses tokens, not ssh keys.

payu/runlog.py
```diff
@@ -139,12 +125,18 @@ def push(self):
...
-               ''.format(key=ssh_key_path))
-        subprocess.check_call(shlex.split(cmd), cwd=self.expt.control_path)
+        ssh_cmd = f'ssh -i {ssh_key_path}'
+        git_repo = GitRepository(self.expt.control_path, catch_error=True)
+        with git_repo.repo.git.custom_environment(GIT_SSH_COMMAND=ssh_cmd):
+            try:
+                remote = git_repo.repo.remotes.payu
+                remote.push(all=True)
+            except Exception as e:
+                print(f"payu: error: Failed to push runlog changes to remote repository: {e}.")
```
aidanheerdegen: If you agree we shouldn't be guessing keys, can we simplify this to use default `.push`?

payu/runlog.py
```diff
@@ -285,29 +272,3 @@ def authenticate(self):
 
         github_auth = (github_username, github_password)
         return github_auth
```
aidanheerdegen: These are redundant now the `ghsetup` cmd has been removed. Should be deleted.

aidanheerdegen (changes requested):
I have some suggestions

payu/branch.py
Qian-HuiChen: Thanks for raising this issue.
In commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f), an error message will be print out if `repo==None`.

payu/git_utils.py
Qian-HuiChen: That makes sense. I updated the parent check to be in front of init, as well as updating the suggest fix in commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f).

payu/models/mom6.py
Qian-HuiChen: Thanks for raising this issue. I add an error message if `repo==None` in commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f)

payu/experiment.py
Qian-HuiChen: Resolved in commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f)

payu/runlog.py
Qian-HuiChen: This is removed in commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f) since we don't guess ssh key anymore

payu/runlog.py
Qian-HuiChen: Commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f) simplifies this to use `git.push`

payu/runlog.py
Qian-HuiChen: I see. 
This function is removed in commit [af729cc](https://github.com/payu-org/payu/pull/743/commits/af729cc5e5443f46b082fc7d58636adc0a62963f) since `ghsetup` command is cleaned up.

aidanheerdegen (dismissed):
Looks great. Really appreciate the testing too.

What seemed like a fairly simple change has led to fairly extensive changes, and much cleaner code, so thanks.

payu/experiment.py
```diff
@@ -565,6 +566,12 @@ def setup(self, force_archive=False):
 
     @timeit("payu_run_duration_seconds")
     def run(self, *user_flags):
+        # Error out if runlog is enabled and running from subdirectory of git root repository
+        if self.runlog.enabled and get_git_repository(self.control_path, catch_error=True) is None:
+            raise PayuGitWarning(
+                "payu: error: Runlog is enabled, but current directory is not a git repository.\n"
+            )
+
```
jo-basevi: Good check. Could add this to setup() method instead so this check can run on the login node if someone runs `payu setup` first. 

Should raise an exception rather than an Warning here

payu/models/mom6.py
```diff
@@ -64,7 +64,10 @@ def mom6_save_docs_files(model):
 
     if model.expt.runlog.enabled: #if runlog true, default to true
         # commit new files to the control dir
-        repo = GitRepository(repo_path = model.control_path)
+        repo = GitRepository(repo_path = model.control_path, catch_error=True)
+        if repo.repo is None:
+            print("payu: warning: Invalid repository, could not commit new files to the control directory.")
```
jo-basevi: Probably best to use warnings.warn, so it can be formatted consistently in the future. E.g.

```suggestion
            warnings.warn("Failed to commit MOM6 run-time configuration to the control directory.")
```

payu/branch.py
```diff
@@ -185,7 +185,10 @@ def checkout_branch(branch_name: str,
...
+    if repo.repo is None:
+        print("payu: error: Invalid repository, could not checkout branch.")
```
jo-basevi: To avoid a checkout being left in a incomplete state when using `payu clone` could raise a `PayuBranchError("Invalid repository, could not checkout branch.")` here. There's logic in `payu clone` to remove the control directory when `PayuBranchError` is raised.

payu/git_utils.py
```diff
@@ -30,15 +41,28 @@ def get_git_repository(repo_path: Union[Path, str],
...
-            PayuGitWarning
-        )
+        parent_repo = check_git_parent(repo_path)
+        if parent_repo:
+            warnings.warn(
+                f"Payu expects to run from repository root to ensure metadata tracking.\n"
+                f"Current situation leads to an InvalidGitRepositoryError.\n"
+                f"Suggest fix:\n"
+                f"    Move the control directory to the a separate directory with its own git repository\n"
+                f"    Or set `runlog: false`\n",
```
jo-basevi: This warning will be displayed when running `payu checkout`/`payu clone` too - so setting `runlog: false` will not work.

payu/runlog.py
```diff
@@ -71,47 +73,31 @@ def create_manifest(self):
...
+        except git.exc.GitCommandError as e:
             try:
-                subprocess.check_call(shlex.split(cmd),
-                                      stdout=f_null,
-                                      cwd=self.expt.control_path)
-                warnings.warn("Runlog commit was commited without git signing")
-            except subprocess.CalledProcessError:
-                warnings.warn("Error occured when attempting to commit runlog")
+                git_repo.repo.index.add(paths_to_add)
+                git_repo.repo.index.commit(m=commit_msg, no_gpg_sign=True)
```
jo-basevi: Could extend `git_repo.commit` to take an optional `no_gpg_sign` argument? So this can be 

```
git_repo.commit(commit_msg, paths_to_add, no_gpg_sign=True)
```

A benefit is that it keeps any GitPython logic within git_utils, and it runs the same logic as the previous commit attempt, but just with signing disabled.

payu/runlog.py
```diff
@@ -71,47 +73,31 @@ def create_manifest(self):
...
+                warnings.warn("Runlog commit without gpg signing.")
+            except git.exc.GitCommandError as e:
+                print(f"payu: error: Failed to commit runlog changes to git repository: {e}.")
```
jo-basevi: Could be `warnings.warn` as above

payu/runlog.py
```diff
@@ -71,47 +73,31 @@ def create_manifest(self):
...
 
         # Save the commit hash
-        self.expt.run_id = commit_hash(self.expt.control_path)
-
-        f_null.close()
+        self.expt.run_id = git_repo.repo.head.object.hexsha
 
     def push(self):
         """Push the changes to the remote repository.
```
jo-basevi, line 83: Nice! This function looks very simple now :) 

It'll be good to update the doc-string to remove the SSH key info.

I am wondering whether it's worth keeping this `payu push` command as `git push --all payu` can be run instead?

jo-basevi (commented):
Nice refactor with the runlog :)

payu/runlog.py
jo-basevi: I'm not sure what this code does, but I don't think it's needed anymore?

payu/models/mom6.py
Qian-HuiChen: Sounds good. Implemented in the latest commit.

payu/branch.py
Qian-HuiChen: I see. I change this to `raise PayuBranchError` so `payu clone` will clean up after the `try::except`.

payu/experiment.py
Qian-HuiChen: Right, I moved this to setup() and added a unit test.

payu/git_utils.py
Qian-HuiChen: I removed the "Or set `runlog: false`" printing.

payu/runlog.py
Qian-HuiChen: Ha, this must be the residual from the previous code. I removed this in the latest commit.

payu/runlog.py
Qian-HuiChen: Good point. This is so much cleaner. Implemented in the latest commit.

payu/runlog.py
Qian-HuiChen: Implemented in the latest commit.

payu/runlog.py
Qian-HuiChen: I am not sure if there are users using `payu push`. If we remove this feature, they might get angry (?)

payu/runlog.py
```diff
@@ -71,47 +58,30 @@ def create_manifest(self):
...
+                git_repo.commit(commit_msg, paths_to_add, gpg_sign=False)
+                warnings.warn("Runlog commit without gpg signing.")
+            except git.exc.GitCommandError as e:
+                warnings.warn(f"Failed to commit runlog changes to git repository: {e}.")
 
         # Save the commit hash
-        self.expt.run_id = commit_hash(self.expt.control_path)
-
-        f_null.close()
+        self.expt.run_id = git_repo.repo.head.object.hexsha
```
jo-basevi: Could reuse the get_hash function here, e.g.
`self.expt.run_id = git_repo.get_hash()`

payu/git_utils.py
```diff
@@ -131,6 +155,10 @@ def commit(self,
 
         # Run commit if there's changes
         if changes:
+            # Set gpg signing for this commit
+            with self.repo.config_writer() as writer:
+                writer.set_value("commit", "gpgsign", str(gpg_sign).lower())
+                
             self.repo.index.commit(commit_message)
             print(commit_message)
```
jo-basevi, line 160: It looks like `index.commit()` does not support git signing (which I did not realise!), so that means currently new payu UUID commits and mom6 commits are not signed..

We can call the git commit directly with `git.commit()` and that seems to work. 

```suggestion
        if changes:
            try:
                self.repo.git.commit("-m", commit_message)
            except git.exc.GitCommandError:
                self.repo.git.commit("--no-gpg-sign", "-m", commit_message)
                warnings.warn(
                    "Git commit was added without GPG signing",
                    PayuGitWarning
                )
            print(commit_message)
            
```

payu/runlog.py
jo-basevi: Thanks!

payu/runlog.py
jo-basevi: Yeah true, though this PR is changing how it works already. We can add this change in, and if no one complains, then can remove it another day maybe.

jo-basevi (commented):
Thanks for adding in those changes! Just a comment on a bug that was already in the code..

payu/git_utils.py
Qian-HuiChen: In the latest commit, I updated `GitRepository.commit()` to use `repo.git.commit`. I leave the `try::except` logic and warning to `runlog.py` when `git_repo.commit` is called.

payu/runlog.py
Qian-HuiChen: Sounds good! Implemented in the latest commit.

payu/git_utils.py
jo-basevi: The try and except logic could be moved into git_utils so it can be re-used by the new metadata UUID commits  and mom6 files. As my signing keys are forwarded from my laptop, signing on login nodes doesn't cause errors but signing commits on compute nodes does (so will now impact that mom6 commit too).

Qian-HuiChen: I see. I moved the try except logic into git_util: commit() function in [9a586f0](https://github.com/payu-org/payu/pull/743/commits/9a586f062db4bda01be91107bf73abbe1a63d3d9)

jo-basevi: Thanks! Runs without errors

jo-basevi (approved):
Thanks @Qian-HuiChen! Looks good to me - no comments :)

aidanheerdegen (approved):
Looks great!

PR #748, opened 2026-05-27
Fix missing warning in PBS.

Qian-HuiChen:
Configure logging in PBS so warning appears in `std.err`.
Ref [comment](https://github.com/payu-org/payu/pull/713#discussion_r3285098967) .

The formatting of warning messages in PBS will be addressed in another PR.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -33,11 +33,10 @@ from payu.telemetry import write_queued_job_file
 # Default configuration
 DEFAULT_CONFIG = 'config.yaml'
 
-# Pass the warning through the logger
-logging.captureWarnings(True)
-
 def parse():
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
I added some custom warnings to trigger and confirm that when running `payu run`, the warning was still formatted:

```
[WARNING] Test payu run warning!!
```

And similarly with `payu-run`, the following was in my PBS standard output file:

```
/home/189/user-id/payu_development/payu_fork/payu/subcommands/run_cmd.py:184: UserWarning: Test payu-run Warning!!
  warnings.warn("Test payu-run Warning!!")
```

I don't know if re-adding the global warning formatting would mess with the warning formatting in `parse()`, but for now, some warnings are better than no warnings

PR #750, opened 2026-05-29
Configure warning formatting and log-level

Qian-HuiChen:
This PR configures the warning formatting and log level for both PBS jobs and direct  payu run scripts.

Changes:
- Standardised warning handling:
Added `set_stacktrace_runscript` to control warning formatting based on the `--stacktrace` flag or the `PAYU_STACKTRACE` environment variable in PBS job submission.
- Log-level configuration:
Added `set_logger_runscript` to ensure log levels are configured consistently, with correct priority between command-line arguments and environment variables.
- CLI argument support:
Ensures both `payu cmd` and `payu-cmd` correctly accept and apply the `--stacktrace` and `--log-level` arguments.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -199,8 +199,10 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
             payu_env_vars[var] = os.environ[var]
 
     # Pass on stacktrace and log level to PBS as environment variables
-    payu_env_vars['PAYU_STACKTRACE'] = os.environ.get('PAYU_STACKTRACE', 'False')
-    payu_env_vars['PAYU_LOG_LEVEL'] = os.environ.get('PAYU_LOG_LEVEL', 'INFO')
+    if os.environ.get('PAYU_STACKTRACE'):
+        payu_env_vars['PAYU_STACKTRACE'] = os.environ.get('PAYU_STACKTRACE')
+    if os.environ.get('PAYU_LOG_LEVEL'):
+        payu_env_vars['PAYU_LOG_LEVEL'] = os.environ.get('PAYU_LOG_LEVEL')
 
     return payu_env_vars
 
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -63,11 +63,14 @@ def parse():
 
     # We pop --stacktrace and --log_level here so they will not be propagated to runcmd() in subcommands
     stacktrace = args.pop('stacktrace', False)
-    log_level = args.pop('log_level', 'INFO')
+    log_level = args.pop('log_level', None)
 
     # Override the STACKTRACE and LOG_LEVEL environment variables if flags are provided
-    os.environ['PAYU_STACKTRACE'] = str(stacktrace)
-    os.environ['PAYU_LOG_LEVEL'] = str(log_level)
+    if log_level:
+        os.environ['PAYU_LOG_LEVEL'] = str(log_level)
+    if stacktrace:
+        os.environ['PAYU_STACKTRACE'] = str(stacktrace)
+    
         
     _execute_command(run_cmd, stacktrace=stacktrace, log_level=log_level, **args)
 
@@ -250,7 +253,7 @@ def set_logger_runscript(log_level=None):
 
     # If log_level is changed from default, update setup_logger
     # Priority: command line argument > environment variable > 'INFO' default
-    if log_level and log_level != 'INFO':
+    if log_level is not None:
         active_level = log_level
     elif log_level_env:
         active_level = str(log_level_env).upper()
@@ -260,15 +263,19 @@ def set_logger_runscript(log_level=None):
     setup_logger(active_level)
 
 def set_stacktrace_runscript(stacktrace=None):
-    """Configure stacktrace settings based on arguments and environment variables."""
+    """
+    Configure stacktrace settings based on arguments and environment variables.
+    Return True if stacktrace is enabled, False otherwise.
+    """
 
     if stacktrace is True or str(os.environ.get('PAYU_STACKTRACE', 'False')).lower() == 'true':
-        pass
+        return True
     else:
         # Force warnings.warn() to omit the source code line in the message
         warnings.formatwarning = (
             lambda message, category, filename, lineno, line=None: f"{message}"
         )
+        return False
 
 def _execute_command(func, stacktrace=None, log_level=None, **args):
     """Execute a payu command with error handling and logging.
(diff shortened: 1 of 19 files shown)


payu/cli.py
```diff
@@ -186,6 +198,10 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
         if var in os.environ:
             payu_env_vars[var] = os.environ[var]
 
+    # Pass on stacktrace and log level to PBS as environment variables
+    payu_env_vars['PAYU_STACKTRACE'] = os.environ.get('PAYU_STACKTRACE', 'False')
+    payu_env_vars['PAYU_LOG_LEVEL'] = os.environ.get('PAYU_LOG_LEVEL', 'INFO')
```
jo-basevi: Maybe only set these if defined? E.g.
```
if os.environ.get('PAYU_STACKTRACE'):
    payu_env_vars['PAYU_STACKTRACE'] = os.environ.get('PAYU_STACKTRACE')
```

jo-basevi (commented):
Thanks @Qian-HuiChen for these changes! The setting and handling of environment variables is very neat.

While the approach of adding functions to format the warnings to each runscript works nicely for warnings, when we need to catch and format exceptions with and without stacktraces, I think we will need a common runscript function. Happy to chat more about this in person

payu/cli.py
Qian-HuiChen: Updated in commit [fbc5bf2](https://github.com/payu-org/payu/pull/750/commits/fbc5bf2eb3a63fcc5ed53e966144f98cdd40af2f)

Qian-HuiChen: Commit [721f3a2](https://github.com/payu-org/payu/pull/750/commits/721f3a2ae0b060e3cb427f8ae5cea565144f88e9) updated with a common runscript function to configure warnings and log-level for all runscripts.

pyproject.toml
```diff
@@ -53,13 +53,13 @@ mitgcm = ["mnctools>=0.2"]
...
-payu-branch = "payu.subcommands.branch_cmd:runscript"
-payu-clone = "payu.subcommands.clone_cmd:runscript"
-payu-checkout = "payu.subcommands.checkout_cmd:runscript"
+payu-run = "payu.cli:parse_run"
+payu-collate = "payu.cli:parse_collate"
+payu-profile = "payu.cli:parse_profile"
+payu-sync = "payu.cli:parse_sync"
+payu-branch = "payu.cli:parse_branch"
+payu-clone = "payu.cli:parse_clone"
+payu-checkout = "payu.cli:parse_checkout"
```
jo-basevi: Safe to delete `payu-branch`, `payu-clone`, `payu-checkout` methods. I must have added the runscripts as a mistake in the past.

Qian-HuiChen: I see. Removed in [86cdc6f](https://github.com/payu-org/payu/pull/750/commits/86cdc6fed70dc562e35060f31d447abad91c8e8e). Confirm that these command runs fine after removal.

jo-basevi: Thanks!

payu/cli.py
```diff
@@ -53,15 +61,15 @@ def parse():
...
-        warnings.formatwarning = (
-            lambda message, category, filename, lineno, line=None: f"{message}"
-        )
+    # We pop --stacktrace and --log_level here so they will not be propagated to runcmd() in subcommands
+    stacktrace = args.pop('stacktrace', False)
+    log_level = args.pop('log_level', 'INFO')
+
+    # Override the STACKTRACE and LOG_LEVEL environment variables if flags are provided
+    os.environ['PAYU_STACKTRACE'] = str(stacktrace)
+    os.environ['PAYU_LOG_LEVEL'] = str(log_level)
```
jo-basevi: I'm not sure how much this matters - so feel free to push back on this. 

Currently this will always set environment variables for PAYU_LOG_LEVEL and PAYU_STACKTRACE. 

An option could be to have `log_level` default to None, and then here only set environment variables if the user has explicitly passed an flag to the command, e.g.:

```
if log_level is not None:
    os.environ['PAYU_LOG_LEVEL'] = log_level
if stacktrace:
   os.environ['PAYU_LOG_LEVEL'] = stacktrace
```

It'll make the PBS script less verbose as these arguments aren't specified if not needed (which isn't that important), but it does make clear when a user has passed a command-line flag.

payu/cli.py
```diff
@@ -225,3 +242,81 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
...
+    log_level_env = os.environ.get('PAYU_LOG_LEVEL')
+
+    # If log_level is changed from default, update setup_logger
+    # Priority: command line argument > environment variable > 'INFO' default
+    if log_level and log_level != 'INFO':
+        active_level = log_level
+    elif log_level_env:
+        active_level = str(log_level_env).upper()
+    else:
+        active_level = 'INFO'
```
jo-basevi: If using None when there's no `log_level` cmd-line flag then this could be:
```
if log_level is not None:
    active_level = log_level
elif os.environ.get('PAYU_LOG_LEVEL'):
    active_level = os.environ['PAYU_LOG_LEVEL']
else:
    active_level = 'INFO'  
```

payu/cli.py
```diff
@@ -225,3 +242,81 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
...
+
+    setup_logger(active_level)
+
+def set_stacktrace_runscript(stacktrace=None):
+    """Configure stacktrace settings based on arguments and environment variables."""
+
+    if stacktrace is True or str(os.environ.get('PAYU_STACKTRACE', 'False')).lower() == 'true':
```
jo-basevi, line 271: Just thinking about the error-handling PR that this function could return True/False whether to show stacktrace or not?

jo-basevi (dismissed):
Thanks for adding all these changes in!! I ran a few tests and it seems to work for me.

payu/cli.py
Qian-HuiChen: Updated in commit [df5ea62](https://github.com/payu-org/payu/pull/750/commits/df5ea62ca158a0c3015e2f46e420ec068612cefb)

payu/cli.py
Qian-HuiChen: Good idea. Implemented in commit [df5ea62](https://github.com/payu-org/payu/pull/750/commits/df5ea62ca158a0c3015e2f46e420ec068612cefb)

payu/cli.py
Qian-HuiChen: I think it is good to make PBS less verbose by default. I changed the default log-level to `None` in commit [df5ea62](https://github.com/payu-org/payu/pull/750/commits/df5ea62ca158a0c3015e2f46e420ec068612cefb)

jo-basevi (approved):
Added changes look good to me :)

PR #751, opened 2026-05-29
Add option to specify experiment name prefix.

### LINKED ISSUE #746: Add option to specify experiment name prefix
It is possible to override the automatic experiment naming which concatenates the control directory + branch name + UUID (branch-UUID aware),  with a simple string by specifying `experiment:` in the `config.yaml`.

However there is no way to specify just the "prefix", so override the control directory name, and still have a branch-UUID aware name.

The use case for this is running multiple branches from the same repo simultaneously (see https://github.com/payu-org/payu/pull/744#issuecomment-4537936390). To achieve consistent naming of the experiments, as if they were run sequentially from the same repository, requires cloning the repository multiple times into uniquely named sub-directories but with the control directory name consistent. Because all the experiments have to have the same control directory name. e.g.

Given a repo with name `mom5-bowl1` with 3 branches, `ctrl`, `up` and `down`.

To run all three branches simultaneously with consistently named experiments (`mom5-bowl`) would require checking out the repo in three separate uniquely named sub-directories, e.g.

```
mom5-bowl-experiments/
├── 1/
│   └── mom5-bowl/
│       ├── .git
│       ├── config.yaml
│       ├── input.nml
│       └── ...
├── 2/
│   └── mom5-bowl/
│       ├── .git
│       ├── config.yaml
│       ├── input.nml
│       └── ...
└── 3/
    └── mom5-bowl/
        ├── .git
        ├── config.yaml
        ├── input.nml
        └── ...
```

| Control dir name | Branch | Experiment Name |
|--------|--------|--------|
| 1/mom5-bowl | ctrl | mom5-bowl-ctrl-UUID |
| 2/mom5-bowl | up | mom5-bowl-up-UUID | 
| 3/mom5-bowl | down | mom5-bowl-up-UUID |

~~Note: the unique names of the first level of sub-directory above is the same as the branch name for convenience and to make the purpose obvious, but it isn't required, and has no effect on the experiment name.~~
Arbitrary names have been used for the first level of sub-directory above to show these have no role in how the experiment name is generated.

If there was an option to specify just the prefix part of the experiment name, e.g. `experiment_prefix: mom5-bowl` then the structure above could be made simpler because it wouldn't matter what the control directory was named:
```
mom5-bowl-experiments/
├── 1/
│   ├── .git
│   ├── config.yaml
│   ├── input.nml
│   └── ...
├── 2/
│   ├── .git
│   ├── config.yaml
│   ├── input.nml
│   └── ...
└── 3/
    ├── .git
    ├── config.yaml
    ├── input.nml
    └── ...
```

| Control dir name | Branch | Experiment Name |
|--------|--------|--------|
| 1 | ctrl | mom5-bowl-ctrl-UUID |
| 2 | up | mom5-bowl-up-UUID | 
| 3 | down | mom5-bowl-up-UUID |

Edit: Apparently this is still confusing because I used the branch names in the subdirectories above, so I have edited this to take away that confusion and added tables of the resulting experiment names.

Qian-HuiChen:
This PR adds an option for user to specify `experiment_prefix` in the `config.yaml`.

- When `experiment_prefix` is configured and not empty, the experiment name is set to `${prefix}-{branch}-{UUID}`. In this case, control directory name is not included in the experiment name.
- When `experiment_prefix` is not configured, the experiment name remains as `${ctrl_dir_name}-{branch}-{UUID}`. 

The logic of prioritising `experiment` in the `config.yaml` remains unchanged, i.e., once `experiment` is configured and not empty, the experiment name will be the exact same as the configuration.

Closes #746

### DIFF
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -143,9 +143,6 @@ class Metadata:
 
     def new_experiment_name(self) -> str:
         """Generate a new experiment name"""
-        # Get expt prefix from config
-        expt_name_prefix = self.config.get("experiment_name_prefix", None)
-
         if self.branch is None:
             self.branch = self.repo.get_branch_name()
 
@@ -157,8 +154,8 @@ class Metadata:
         suffix += f'-{truncated_uuid}'
 
         # If experiment name prefix is configured and not empty, expt name is prefix-branch-UUID
-        if expt_name_prefix:
-            return expt_name_prefix + suffix
+        if self.expt_name_prefix and self.expt_name_prefix.strip():
+            return self.expt_name_prefix + suffix
         # Otherwise, expt name is ctrl_dir_name-branch-UUID
         else:
             return self.control_path.name + suffix
@@ -170,7 +167,14 @@ class Metadata:
         sub-directories in the Laboratory"""
         # Experiment name configuration - this overrides experiment name
         self.experiment_name = self.config.get("experiment", None)
-        if self.experiment_name is not None:
+        self.expt_name_prefix = self.config.get("experiment_prefix", None)
+
+        if self.experiment_name is not None and self.experiment_name.strip():
+            # Check if experiment prefix is also configured and warn that it will be ignored
+            if self.expt_name_prefix and self.expt_name_prefix.strip():
+                warnings.warn("Both experiment name and prefix are configured in config.yaml.\n"
+                            "Only the experiment name will be used, while the experiment prefix is ignored.",
+                            UserWarning)
             print(f"Experiment name is configured in config.yaml: ",
                   self.experiment_name)
             return
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -143,6 +143,9 @@ class Metadata:
 
     def new_experiment_name(self) -> str:
         """Generate a new experiment name"""
+        # Get expt prefix from config
+        expt_name_prefix = self.config.get("experiment_name_prefix", None)
+
         if self.branch is None:
             self.branch = self.repo.get_branch_name()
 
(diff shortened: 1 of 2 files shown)


payu/metadata.py
```diff
@@ -143,6 +143,9 @@ def setup(self,
 
     def new_experiment_name(self) -> str:
         """Generate a new experiment name"""
+        # Get expt prefix from config
+        expt_name_prefix = self.config.get("experiment_name_prefix", None)
```
aidanheerdegen: I like how explicit `experiment_name_prefix` is, but it's not consistent with the related option, which is just `experiment`.

So unless we change `experiment` to `experiment_name`, I think it would be better to just use `experiment_prefix`.

payu/metadata.py
```diff
@@ -153,7 +156,12 @@ def new_experiment_name(self) -> str:
         truncated_uuid = self.uuid[:TRUNCATED_UUID_LENGTH]
         suffix += f'-{truncated_uuid}'
 
-        return self.control_path.name + suffix
+        # If experiment name prefix is configured and not empty, expt name is prefix-branch-UUID
+        if expt_name_prefix:
```
aidanheerdegen: Note that this isn't checking explicitly for an empty string:
```yaml
experiment_name_prefix: ""
```

Admittedly it is quite difficult to do accidentally, but nonetheless we should check that there is something meaningful, e.g. `not experiment_name_prefix.strip()`. 

A check like this is probably better done when the value is set so the user can be warned it is being ignored.

payu/metadata.py
aidanheerdegen: Should check `experiment` isn't empty/blank as well.

Also do we need some logic here to check if `experiment_name_prefix` is also set that setting `experiment` overrides it?

aidanheerdegen (changes requested):

payu/metadata.py
Qian-HuiChen: Sounds good, I update the key as `experiment_prefix` when reading in the `config.yaml`.

payu/metadata.py
Qian-HuiChen: Yeah, it is good to check for an empty string. I update this `if` logic to:
```
if self.expt_name_prefix and self.expt_name_prefix.strip():
```

payu/metadata.py
Qian-HuiChen: I updated this `if` logic to:
```
if self.experiment_name is not None and self.experiment_name.strip():
```
In the latest commit, I also updated this block so that when both `experiment` and `experiment_prefix` are configured, 
- payu emits a warning `Only the experiment name will be used, while the experiment prefix is ignored`, 
- sets the experiment name to the value of `experiment` in `config.yaml`, 
- returns immediately. As a result, new_experiment_name() is not called  at all.

aidanheerdegen (approved):
Looks good, thanks!

PR #759, opened 2026-06-05
Set parent experiment to source branch UUID when `-p BASE` is set in `payu checkout`

### LINKED ISSUE #752: Get `payu checkout -p` option to accept git references in addition to UUID
While discussing a workflow to setup experiments for ESM1.6 (see https://github.com/ACCESS-Community-Hub/access-esm1.6-dev-experiments/pull/16), we came up with a workflow like:

```bash
# Checkout the initial ref for the new experiment
git checkout <git-ref>
# Get the UUID for that experiment to use as parent UUID
payu branch -v | grep "experiment_uuid"
# Create new branch with correct parent UUID
payu checkout -b esm-flat10-zec-2026-05-28 -p <paste-UUID-from-previous-step>
```

I think this workflow could be neatly collapsed into one command if the `-p` option accepted git references in addition to UUID. What I'm thinking about is doing something like:

```bash
payu checkout -b esm-flat10-zec-2026-05-28 <git-ref> -p HEAD
# or
payu checkout -b esm-flat10-zec-2026-05-28 <git-ref> -p .
```

Obviously, the `HEAD` should refer to the state of the repository once we are on the `<git-ref>` given to `payu checkout` and not wherever we are now on the repository. Then, payu would do the work internally to extract the correct UUID.

It could be dangerous for `-p` to accept all git references. For example, we probably don't want to start a new branch from an existing experiment and use the UUID from another experiment as parent_experiment. I can't see use cases for this but I'm happy to be told they exist. So maybe we only want `-p` to accept `HEAD` or `.` (I prefer HEAD) or something like that.

@blimlim FYI, in case you think of anything else.

Qian-HuiChen:
This PR sets the parent experiment to the UUID of the source branch, when `payu checkout -b new_branch <git-ref_of_source_branch> -p BASE` is used. 
This saves the user from looking for UUID of the source branch.

Closes #752

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -530,18 +530,20 @@ for example::
 
       payu checkout -b perturb --restart path/to/restart
 
+This automatically sets the parent experiment UUID in the new branch's metadata to the
+experiment UUID of the restart directory, if one exists.
 Note: This can also be achieved by configuring ``restart`` (see :ref:`config`).
 
 Alternatively, you can base a new experiment on an existing experiment, using the ``--parent-experiment`` or ``-p`` flag, for example::
 
       payu checkout -b perturb2 perturb1 -p ${PARENT_EXPERIMENT_UUID}
 
-To simplify workflows, you can use ``BASE`` to refer to the current experiment UUID of the source branch::
+To simplify workflows, you can use ``BASE`` to refer to the experiment UUID of the source branch::
 
       payu checkout -b perturb2 perturb1 -p BASE
 
-In this case, payu extracts the current experiment UUID from the metadata of the source branch, and 
-sets it as the parent experiment UUID for the newly created branch.
+In this example, payu extracts the experiment UUID from the metadata of the source branch ``perturb1``, 
+and sets it as the parent experiment UUID for the newly created branch ``perturb2``.
 To checkout and switch to an existing branch and experiment, omit the ``-b`` flag. 
 For example, the following checks out the ``perturb1`` branch:: 
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -532,6 +532,16 @@ for example::
 
 Note: This can also be achieved by configuring ``restart`` (see :ref:`config`).
 
+Alternatively, you can base a new experiment on an existing experiment, using the ``--parent-experiment`` or ``-p`` flag, for example::
+
+      payu checkout -b perturb2 perturb1 -p ${PARENT_EXPERIMENT_UUID}
+
+To simplify workflows, you can use ``BASE`` to refer to the current experiment UUID of the source branch::
+
+      payu checkout -b perturb2 perturb1 -p BASE
+
+In this case, payu extracts the current experiment UUID from the metadata of the source branch, and 
+sets it as the parent experiment UUID for the newly created branch.
 To checkout and switch to an existing branch and experiment, omit the ``-b`` flag. 
 For example, the following checks out the ``perturb1`` branch:: 
 
--- a/payu/branch.py
+++ b/payu/branch.py
@@ -44,7 +44,7 @@ To checkout an existing branch, run:
     payu checkout BRANCH_NAME
 Where BRANCH_NAME is the name of the branch"""
 
-SET_PARENT_TO_CURRENT_STRING = "CURRENT"
+DEFAULT_PARENT_STRING = "PARENT"
 
 def remove_traceback_hook(kind, message, traceback):
     """Remove traceback for only PayuBranchError"""
@@ -185,14 +185,16 @@ def checkout_branch(branch_name: str,
     if control_path is None:
         control_path = get_control_path(config_path)
     
+    # Checkout branch
     repo = GitRepository(control_path)
-    
-    # If parent_experiment is set to `CURRENT`, set to current branch's experiment UUID
-    if parent_experiment == SET_PARENT_TO_CURRENT_STRING:
-        print("Setting parent experiment to current branch's experiment UUID")
-        # Read experiment UUID from metadata on current branch
-        current_branch = repo.get_branch()
-        metadata = get_branch_metadata(current_branch)
+    repo.checkout_branch(branch_name, is_new_branch, start_point)
+
+     # If parent_experiment is set to DEFAULT_PARENT_STRING, set to start_point's experiment UUID
+    if parent_experiment == DEFAULT_PARENT_STRING:
+        print("Setting parent experiment to start point's experiment UUID")
+        # Read experiment UUID from metadata on start point's branch
+        start_branch = repo.get_branch()
+        metadata = get_branch_metadata(start_branch)
         uuid = metadata.get(UUID_FIELD, None)
         
         if uuid is None:
--- a/payu/branch.py
+++ b/payu/branch.py
@@ -44,6 +44,7 @@ To checkout an existing branch, run:
     payu checkout BRANCH_NAME
 Where BRANCH_NAME is the name of the branch"""
 
+SET_PARENT_TO_CURRENT_STRING = "CURRENT"
 
 def remove_traceback_hook(kind, message, traceback):
     """Remove traceback for only PayuBranchError"""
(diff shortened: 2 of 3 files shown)


test/test_branch.py
```diff
@@ -516,6 +516,54 @@ def test_checkout_branch_with_restart_path(mock_uuid):
...
+        checkout_branch(is_new_branch=True,
+                        branch_name="Branch1",
+                        lab_path=labdir)
+
+    # Check metadata
+    experiment1_name = f"{ctrldir_basename}-Branch1-df050eaf"
+    check_branch_metadata(repo,
+                          expected_current_branch='Branch1',
+                          expected_uuid=uuid1,
+                          expected_experiment=experiment1_name)
```
aidanheerdegen:
```suggestion
    branch_name='Branch1'
    with cd(ctrldir):
        # Test checkout with no metadata
        checkout_branch(is_new_branch=True,
                        branch_name="branch_name,
                        lab_path=labdir)

    # Check metadata
    experiment1_name = f"{ctrldir_basename}-{branch_name}-{uuid1:.8}"
    check_branch_metadata(repo,
                          expected_current_branch=branch_name,
                          expected_uuid=uuid1,
                          expected_experiment=experiment1_name)
```
I think it's a good idea to use and reuse variables as much as possible.

test/test_branch.py
aidanheerdegen, line 1: There is no test for the error condition where there is no parent UUID.

aidanheerdegen (changes requested):
Looks nice and clean. Thanks.

A couple of suggestions around testing, but we also need a docs update I assume?

test/test_branch.py
aidanheerdegen: I didn't make a recommendation for the second part of this test, but that could also do something similar.

ccarouge:
I put a comment in the issue but I saw the discussion late so I'm putting it here too in case it is more visible.

I'm not keen on `-p CURRENT` because that's confusing. If we do `payu checkout -b my-branch old-branch`, we want to pick the UUID from `old-branch` and not the UUID of the branch currently checked out.

payu/branch.py
```diff
@@ -183,9 +184,25 @@ def checkout_branch(branch_name: str,
...
+        current_branch = repo.get_branch()
+        metadata = get_branch_metadata(current_branch)
+        uuid = metadata.get(UUID_FIELD, None)
+        
+        if uuid is None:
+            raise PayuBranchError(
+                "No UUID in metadata file. Cannot set parent experiment to current experiment."
+            )
+
+        parent_experiment = uuid
```
ccarouge, line 207: This needs to happen after `repo.checkout_branch(branch_name, is_new_branch, start_point)`!

That's very important. Because we want the UUID of `start_point` not the one where we run the command.

ccarouge (changes requested):

test/test_branch.py
Qian-HuiChen: Sounds good. I reuse the branch_names and uuid[0:8] variable in the latest commit.

payu/branch.py
Qian-HuiChen: Thanks for clarifying the goal. 
In the latest commit, I move this to after 
```
repo.checkout_branch(branch_name, is_new_branch, start_point)
```
Also updated the unit tests for its purpose.

test/test_branch.py
Qian-HuiChen: Good point, I add a test about the error in `test_checkout_branch_with_parent_experiment` in commit [fa92938](https://github.com/payu-org/payu/pull/759/commits/fa929389c2f0d6654b3fb9c9b925bf1dfa7e6f4b)

docs/source/usage.rst
```diff
@@ -523,6 +523,14 @@ for example::
...
 Note: This can also be achieved by configuring ``restart`` (see :ref:`config`).
 
+Alternatively, you can base a new experiment on an existing one, using the ``--parent-experiment`` or ``-p`` flag, for example::
+
+      payu checkout -b perturb2 perturb1 -p ${PARENT_EXPERIMENT_UUID}
+
+To simplify workflows, you can use ``BASE`` to refer to the parent experiment UUID that will be branching from::
+
+      payu checkout -b perturb2 perturb1 -p BASE
+
```
aidanheerdegen, line 544: Would be good to explain *exactly* what this does, i.e. extract UUID from `metadata.yaml` file in `perturb1` branch/experiment.

aidanheerdegen (changes requested):
Sorry, just one small change to the docs please.

docs/source/usage.rst
Qian-HuiChen: Good idea. Added in commit [f702991](https://github.com/payu-org/payu/pull/759/commits/f7029915a214ecb8c98cfc89f62b185ec8c89d65)

aidanheerdegen (dismissed):

docs/source/usage.rst
```diff
@@ -523,6 +523,16 @@ for example::
...
+Alternatively, you can base a new experiment on an existing experiment, using the ``--parent-experiment`` or ``-p`` flag, for example::
+
+      payu checkout -b perturb2 perturb1 -p ${PARENT_EXPERIMENT_UUID}
+
+To simplify workflows, you can use ``BASE`` to refer to the current experiment UUID of the source branch::
+
+      payu checkout -b perturb2 perturb1 -p BASE
+
+In this case, payu extracts the current experiment UUID from the metadata of the source branch, and 
+sets it as the parent experiment UUID for the newly created branch.
```
ccarouge:
```suggestion
To simplify workflows, you can use ``BASE`` to refer to the experiment UUID of the source branch::

      payu checkout -b perturb2 perturb1 -p BASE

In this example, payu extracts the experiment UUID from the metadata of the source branch, `perturb1`, and sets it as the parent experiment UUID for the newly created branch, `perturb2`.
```

ccarouge (dismissed):
I've made a small suggestion for the documentation. I am also wondering if it's missing a test: are you testing if someone gives a UUID to option `-p` but somehow the UUID is wrong? I guess that's not necessarily an easy test to define. There could be a (relatively) easy check on the format of the string. This might already be in payu but I'm not sure I see a test for it. 

What is more complex is ensuring that the UUID provided is a UUID used by an existing experiment. So that's probably its own problem and not part of this PR.

Anyway, I'm happy to approve, just have a look at the suggestion for the documentation. I let you decide on the suggested test on UUID formatting, if that's useful or would need additional development in payu etc.

docs/source/usage.rst
Qian-HuiChen: Thanks for raising this issue! After testing, I can confirm that payu accepts the parent uuid without any check, even when the provided parent uuid is invalid.
I created a new issue #782 for it so I won't forget to add a UUID check :)

I updated the document in commit [4710da8](https://github.com/payu-org/payu/pull/759/commits/4710da80bffde7e85c91281aed5dc0be39c93866)

docs/source/usage.rst
jo-basevi, line 531: Do we want to mention that setting restart will set the parent experiment UUID to the experiment UUID in the restart directory?

jo-basevi (dismissed):
Had a very quick review as I was curious. 

So payu checks the parent directory of the `restart: ` field to check if metadata.yaml and there's a UUID if parent_experiment is not defined. This might not be the case if the restart directory has been moved out of a payu archive structure. 

https://github.com/payu-org/payu/blob/616b54e66cd3a76a1321d9af40ddbfbf36697947/payu/metadata.py#L295-L298

I wondered if it is worth checking, if the provided parent uuid matches the value in the `restart:` if it exists. Maybe this ties into the issue linked above?

docs/source/usage.rst
Qian-HuiChen: I think it is worth mentioning it. I don't know that until you point it out, so I assume other user might not know this either. Added in [76cbd41](https://github.com/payu-org/payu/commit/76cbd4127ce80dfd6bd8e674057179205b76a3a8)

ccarouge: How does it work in conjunction with the `-p` option? Is `-p` value ignored?

ccarouge (approved):
LGTM. Thanks for the work.

docs/source/usage.rst
jo-basevi: `-p` overrides it currently

PR #763, opened 2026-06-10
Add informative error message to cli.submit_job and lab.init

Qian-HuiChen:
This PR updates the error message to be more informative in the following two cases:

- Job submission failure (qsub): Captures exceptions from the subprocess and prints the error output to the user.
- Model type detection failure: This usually happens when payu is run outside a control directory and the configuration file is missing.

Thanks @chrisb13 for raising these issues.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -219,7 +219,20 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
     cmd = sched.submit(script, config, vars)
     print(cmd)
 
-    result = subprocess.run(shlex.split(cmd), capture_output=True, check=True)
+    try:
+        result = subprocess.run(shlex.split(cmd), capture_output=True, check=True)
+
+    except subprocess.CalledProcessError as e:
+        error_msg = ("Error occurred while submitting job.\n")
+
+        if e.returncode:
+            error_msg += f"Exit code: {e.returncode}\n"
+        if e.stdout:
+            error_msg += f"STDOUT: {e.stdout.decode('utf8')}\n"
+        if e.stderr:
+            error_msg += f"STDERR: {e.stderr.decode('utf8')}"
+
+        raise RuntimeError(error_msg)
 
     # Decode stdout and extract the job ID which is last for both PBS and Slurm
     result = result.stdout.decode("utf8").strip()
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -220,7 +220,7 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
     print(cmd)
 
     try:
-        result = subprocess.run(shlex.split(cmd), capture_output=True, check=True)
+        result = subprocess.run(shlex.split(cmd), capture_output=True, check=True, text=True)
 
     except subprocess.CalledProcessError as e:
         error_msg = ("Error occurred while submitting job.\n")
@@ -228,14 +228,14 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
         if e.returncode:
             error_msg += f"Exit code: {e.returncode}\n"
         if e.stdout:
-            error_msg += f"STDOUT: {e.stdout.decode('utf8')}\n"
+            error_msg += f"STDOUT: {e.stdout}\n"
         if e.stderr:
-            error_msg += f"STDERR: {e.stderr.decode('utf8')}"
+            error_msg += f"STDERR: {e.stderr}"
 
         raise RuntimeError(error_msg)
 
     # Decode stdout and extract the job ID which is last for both PBS and Slurm
-    result = result.stdout.decode("utf8").strip()
+    result = result.stdout.strip()
     print(result)
     job_id = result.split()[-1]
 
--- a/payu/laboratory.py
+++ b/payu/laboratory.py
@@ -38,7 +38,8 @@ class Laboratory(object):
             model_type = config.get('model')
 
         if not model_type:
-            raise ValueError('Cannot determine model type.')
+            raise ValueError("Cannot determine model type.\n"
+                             "Please ensure payu is running from an experiment control directory containing a config file.")
 
         self.model_type = model_type
 
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -5,6 +5,7 @@
 """
 
 # Standard library
+import math
 import os
 from pathlib import Path
 import re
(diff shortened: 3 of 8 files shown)


payu/cli.py
```diff
@@ -199,7 +199,20 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
...
+
+    except subprocess.CalledProcessError as e:
+        error_msg = ("Error occurred while submitting job.\n")
+
+        if e.returncode:
+            error_msg += f"Exit code: {e.returncode}\n"
+        if e.stdout:
+            error_msg += f"STDOUT: {e.stdout.decode('utf8')}\n"
+        if e.stderr:
+            error_msg += f"STDERR: {e.stderr.decode('utf8')}"
```
jo-basevi: Nice I did not know that these exceptions had the stdout and stderr populated when capturing output.

To avoid the `.decode('utf8')`, could pass `text=True` in subprocess.run() so it's stored in text form rather than bytes.

payu/laboratory.py
```diff
@@ -38,7 +38,8 @@ def __init__(self, model_type=None, config_path=None, lab_path=None):
             model_type = config.get('model')
 
         if not model_type:
-            raise ValueError('Cannot determine model type.')
+            raise ValueError("Cannot determine model type.\n"
+                             "Please ensure payu is running from an experiment control directory containing a config file.")
```
chrisb13, line 42: Yee-hah, that would have really helped me out! :)

jo-basevi, line 42: Initially I thought the exception could be raised if a configuration file is not found - so earlier in the process. 

However.. when reading the configuration, it seems to just display a warning when config is not found:

https://github.com/payu-org/payu/blob/afda4b946bfaed5541dd0038986d2eb077024a9b/payu/fsops.py#L94-L98

And it also seems to support empty config files? @aidanheerdegen do you know of example which uses no configuration? Like does it determine the model from the command line (`--model`) then gets the inputs and executables from the input and bin paths in the payu laboratory?

jo-basevi (dismissed):
Thanks @Qian-HuiChen, looks good to me

payu/laboratory.py
Qian-HuiChen: Issue created in #764 and will be addressed in another PR.

payu/cli.py
Qian-HuiChen: Good idea. Changed in commit [5404c73](https://github.com/payu-org/payu/pull/763/commits/5404c730c5c63627fa611c54ac4e6a2613cee793)

Qian-HuiChen:
In commit [5404c73](https://github.com/payu-org/payu/pull/763/commits/5404c730c5c63627fa611c54ac4e6a2613cee793), I added a validation of requested memory as suggested in our in-person discussion.

payu/schedulers/pbs.py
```diff
@@ -279,6 +279,38 @@ def get_queue_node_shape(cls, queue: str) -> tuple[int, int]:
...
+        queue: the queue name
+        n_cpus: the number of CPUs requested.
+        """
+        # Change the unit of memory to GB
+        if pbs_mem.lower().endswith("tb"):
+            req_mem_gb = float(pbs_mem[:-2]) * 1024
+        elif pbs_mem.lower().endswith("gb"):
+            req_mem_gb = float(pbs_mem[:-2])
+        elif pbs_mem.lower().endswith("mb"):
+            req_mem_gb = float(pbs_mem[:-2]) / 1024
```
jo-basevi: I think PBS can handle "kb", "b" or no units (defaults to bytes). 

Here's an example of a bytes conversion in tracking-services:
https://github.com/ACCESS-NRI/tracking-services/blob/3c9590e9c2946e83014ce4b8d83c1800c2743f6a/server/payu/data_ingestion/telemetry.py#L145-L168

jo-basevi: Though that tracking-services function should use `.lower()` like you do above

payu/laboratory.py
jo-basevi: Thanks @Qian-HuiChen

test/test_pbs.py
```diff
@@ -592,3 +592,40 @@ def test_check_storage_access(storages, user_groups, expected_denied):
...
+
+
+@pytest.mark.parametrize(
+    "pbs_mem, queue, n_cpus, raise_error",
+    [   
+        ("1TB", "normalsr", 48, True),  # Exceeds max memory for normalsr
+        ("1300GB", "normalsr", 48, True),  # Exceeds max memory for normalsr
+        ("100GB", "normalsr", 48, False),  # Valid memory
+        ("100MB", "normalsr", 48, False),  # Valid memory
+        ("0.1TB", "normalsr", 48, False),  # Valid memory
```
jo-basevi, line 604: It'll be good to add test cases when pbs_mem == total memory on the node (e.g. 192GB), and also ncpus > cpus_per_node.

jo-basevi (commented):
Nice test coverage! Just a couple of very minor comments

test/test_pbs.py
Qian-HuiChen: I would never say no to more unit tests. Added in the latest commit.

payu/schedulers/pbs.py
Qian-HuiChen: I see. I mimic the `_to_byte()` in tracking-service and add this function into `pbs.py`. Now it can handle no unit, "b" and all the way up to "PB".

payu/schedulers/pbs.py
```diff
@@ -279,6 +311,33 @@ def get_queue_node_shape(cls, queue: str) -> tuple[int, int]:
...
+        try:
+            req_mem_gb = cls.mem_string_to_gb(pbs_mem)
+        except ValueError:
+            raise ValueError(f"Memory string '{pbs_mem}' has invalid format, must end with PB, TB, GB, MB, KB, B, or no unit.")
+        
+        # Get the node shape for the queue
+        cpu_per_node, mem_per_node = cls.get_queue_node_shape(queue)
+        
+        # Calculate the requested memory per node, node number is rounded up
+        req_mem_per_node = req_mem_gb / math.ceil(n_cpus / cpu_per_node)
```
Qian-HuiChen, line 335: In the latest commit, I change this line to use math.ceil. In case someone asks for 100GB with cpu =4, which is rare but just in case.

payu/schedulers/pbs.py
```diff
@@ -250,6 +251,37 @@ def validate_walltime_with_queue_limits(cls, walltime: int | str, queue: str, nc
...
+                # Remove the suffix and convert to float
+                size_str = size_str[: -len(suffix)]
+                return float(size_str) * multiplier
+
+        # If no suffix is found, assume it's in bytes
+        warnings.warn(
+            f"Memory string '{size_str}' has no unit suffix, assuming bytes.\n "
+            "It is recommended to specify units explicitly (e.g. '100GB')."
+        )
+        return float(size_str)/(1024**3)
```
aidanheerdegen: Maybe this is too late, but there is a very nice python package called [pint](https://pint.readthedocs.io/en/stable/getting/tutorial.html) that deals with units of all kinds [including binary units](https://github.com/hgrecco/pint/blob/7a927b4515b29917a76ee14617608531c59dd1d5/pint/default_en.txt#L92-L100).

Resources: a [useful run-through](https://towardsdatascience.com/leveraging-python-pint-units-handler-package-part-1-716a13e96b59/), and a [small example](https://github.com/tisnik/most-popular-python-libs/blob/master/pint/19_byte_to_bites.py)

It's not clear to me that NCI are using the correct [binary unit specifiers](https://en.wikipedia.org/wiki/Binary_prefix).

jo-basevi (dismissed):
Added changes look good to me, thanks @Qian-HuiChen

payu/schedulers/pbs.py
Qian-HuiChen: I don't think it's too late :D I haven't merged it. I implemented `pint` in commit [cc0fec6](https://github.com/payu-org/payu/pull/763/commits/cc0fec638c98633382ea776ee4b945229753c700)

This reminds me of the `astropy.unit` [package](https://docs.astropy.org/en/stable/units/index.html). It was so useful to convert between physical units.

aidanheerdegen (approved):
Nice improvement in user feedback.

chrisb13:
Thanks @Qian-HuiChen, @jo-basevi and @aidanheerdegen!

PR #768, opened 2026-06-15
Create archive symlink in payu setup; show experiment paths in payu status

### LINKED ISSUE #765: `payu setup` should create symlink to archive directory
Currently, `payu setup` only creates a symlink for the work directory. While the archive directory is created, it is not symlinked in the control directory.
This makes it harder for users to use the archive path in a script that's run from the control directory.

Qian-HuiChen:
This PR creates the symlink to archive directory during `payu setup`.

This PR also adds a flag of `--experiment-paths` (`--ep`) to `payu status`. This function will print out the experiment paths to users in human-readable way, or in json format (with `--json` flag). An example of print out is attached below:
```
Experiment Paths:
  Experiment UUID:   oo11000
  Experiment Name:   mom6-777eeee
  Control Directory: /home/$USER/mom6
                      (Where model configuration is stored)
  Work Directory:    /scratch/$PROJECT/$USER/mom6/work/mom6-777eeee
                      (Temporary directory for experiment runs)
  Archive Directory: /scratch/$PROJECT/$USER/mom6/archive/mom6-777eeee
                      (Where all experiment outputs are stored)
  Sync Destination:  /path/to/sync
                      (Remote directory to sync outputs to)
```
```
{
    "experiment_uuid": "777eeee-faaa-cccc-8005-ff6ffffff",
    "experiment_name": "mom6-777eeee",
    "control_path": "/home/$USER/mom6",
    "lab_path": "/path/to/lab",
    "work_path": "/scratch/$PROJECT/$USER/mom6/work/mom6-777eeee",
    "archive_path": "/scratch/$PROJECT/$USER/mom6/archive/mom6-777eeee",
    "sync_path": "path/to/sync"
}
```
Closes #765.


Qian-HuiChen:
Documentation needs to be updated with `--ep` flag at the final stage of this PR.

payu/subcommands/args.py
```diff
@@ -60,13 +60,13 @@
     }
 }
 
-
-force_archive = {
-    'flags': ('--archive',),
+# payu directory paths
+show_expt_paths = {
+    'flags': ('--experiment-paths', '--ep'),
```
aidanheerdegen: Could just use `--paths` as a singular option. It's not used already.

payu/status.py
```diff
@@ -380,4 +383,48 @@ def display_job_info(data: dict[str, Any]) -> None:
...
+
+def collect_expt_paths(expt, lab_path):
+    """Find the experiment paths (control, lab, work, archive, sync) and return them in a dictionary"""
+    expt_paths = {}
+    try:
+        expt_paths = {
+            "experiment_uuid": expt.metadata.uuid,
+            "experiment_name": expt.name,
+            "control_path": expt.control_path,
+            "lab_path": lab_path,
```
aidanheerdegen: I don't know if it's a good idea ... but there is a reference to the `Laboratory` object in each `Experiment` object, so something like
```
lab_path = expt.lab.base_path
```
should be sufficient to recover the laboratory path rather than having to pass it in.

aidanheerdegen (commented):
A couple of initial thoughts.

payu/status.py
Qian-HuiChen: Sounds good to me. Implemented in commit [79b4077](https://github.com/payu-org/payu/pull/768/commits/79b4077f93a5105aafa7827f6f23c14044c6494d)

payu/subcommands/args.py
Qian-HuiChen: Confirm that `--paths` is not used. Updated in commit [79b4077](https://github.com/payu-org/payu/pull/768/commits/79b4077f93a5105aafa7827f6f23c14044c6494d)

aidanheerdegen (approved):
Nice work, thanks.

Issue #544, opened 2024-12-22
payu profiler status


manodeep:
I see that there is a `profiler` sub-module within payu - the code however looks sparse and I could not find any relevant documentation about the profiler - neither for further development nor for how to use it. The only relevant issue I found is [this one](https://github.com/payu-org/payu/issues/27) - which seems to imply that the profiler code might be unused/dead code.

aidanheerdegen:
The profiler code was mostly added by @marshallward when he needed it. It isn't actively used or tested as far as I know.

That said the machinery is there to do things like preload libraries and wrap `mpirun` commands.

Can you give some details of what you require and we can see if it fits in the current design?

manodeep:
Thanks @aidanheerdegen. 

I am hoping to use the Intel Vtune profiler - https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2025-0/mpi-code-analysis.html

The summary is that something like this needs to run for the codes to be profiled:
`mpirun [mpirun-options] vtune [vtune-options] <program> [<args>]` (with possibly a `--` demarcating vtune options and the exe)

For example,
`mpirun -n 16 –ppn 4 –l vtune -collect hotspots -k sampling-mode=hw -trace-mpi -result-dir my_result -- my_app.exe`

aidanheerdegen:
What happens with multiple executables, e.g.

```bash
mpirun --mca io ompio --mca io_ompio_num_aggregators 1 -wdir work/1deg_jra55_ryf-expt-acfa7666/atmosphere -np 1  work/1deg_jra55_ryf-expt-acfa7666/atmosphere/yatm.exe : -wdir work/1deg_jra55_ryf-expt-acfa7666/ocean -np 216  work/1deg_jra55_ryf-expt-acfa7666/ocean/fms_ACCESS-OM.x : -wdir work/1deg_jra55_ryf-expt-acfa7666/ice -np 24  work/1deg_jra55_ryf-expt-acfa7666/ice/cice_auscom_360x300_24x1_24p.exe
```

Does `vtune -collect hotspots -k sampling-mode=hw -trace-mpi -result-dir my_result --` have to be inserted before every executable (after the `-np` options)?

manodeep:
I tried earlier today adding the vtune specific bit in front of every exe (and a new results directory) - but that failed with a seemingly unrelated error about too many files open (but the suggested fix didn't work either). Will try again tomorrow.

manodeep:
[...]

PR #783, opened 2026-06-22
Remove profiler code

KAUR1984:
Profiler code is no longer used, so this PR removes the unused `profiler` code. Fixes issue #544. 

**Changes made:**
- [x] Removed the `profiler` directory and its subfiles. 
- [x] Cleans up the related profiler logic from `experiment.py`.


payu/experiment.py
Qian-HuiChen, line 30:
```suggestion
# Local
import payu
```

Qian-HuiChen: This is to fix the `payu not defined in payu.__version__` error

Qian-HuiChen: Do we need to remove the "profilers" here: https://github.com/payu-org/payu/blob/026abd5dba2ee89654758f042181dbbcf8a23521/docs/source/config.rst?plain=1#L175-L179

jo-basevi: >  This can be useful for configuring profilers or valgrind. 

This is un-related to the payu profiler code as it's a way to manually run a profiler. E.g. If `exe_prefix: aps -r aps-output-dir`, it'll run `mpirun <mpi-params> aps -r aps-output-dir <model-exe>`

KAUR1984: Thanks @Qian-HuiChen @jo-basevi, I guess we are leaning towards leaving this text as is. Resolving this thread if that's ok.

Qian-HuiChen (approved):
Looks good to me. Might need to push a commit with empty changes to re-build the readthedoc.

PR #784, opened 2026-06-24
Remove killed job in payu status --update.

### LINKED ISSUE #777: payu status --update no longer removes deleted queued jobs
After submitting a PBS job, I immediately deleted it with `qdel <JOBID>`. Then running `payu status --update` still shows a queued job. 

I think it's because we are now querying historical job info here:

https://github.com/payu-org/payu/blob/8fbcbfc1ac622ee9f8612b84f28521e6d42f4d93/payu/schedulers/pbs.py#L493

So this logic that removes queued files when there's no job info doesn't work:

https://github.com/payu-org/payu/blob/8fbcbfc1ac622ee9f8612b84f28521e6d42f4d93/payu/status.py#L315-L316

There's a job state field in the info e.g. `"job_state":"F"` to indicate it's finished? 

Though I imagine we will be using HPCpy in future to determine if a job has completed and to fetch job information.

Qian-HuiChen:
This PR removes job files of killed jobs when `payu status --update` is used.

Closes #777

### DIFF
--- a/payu/status.py
+++ b/payu/status.py
@@ -297,13 +297,19 @@ def update_all_job_files(
 
         # Get job info from the scheduler data
         job_info = all_jobs.get(job_id)
-        exit_status = job_info.get("Jobs", {}).get(job_id, {}).get("Exit_status") if job_info else None
-        
-        if job_info and exit_status and stage == "queued":
+        if job_info:
+            job_info_dict = job_info.get("Jobs", {}).get(job_id, {})
+            exit_status = job_info_dict.get("Exit_status")
+            job_state = job_info_dict.get("job_state")
+        else:
+            exit_status = None
+            job_state = None
+
+        if exit_status and stage == "queued":
             # Job has exited, but is still marked as queued in the job file
             remove_job_file(file_path=job_file)
 
-        elif job_info and job_info.get("Jobs", {}).get(job_id, {}).get("job_state") == "F":
+        elif job_state == "F" and stage == "queued":
             # Job is killed or deleted but still exists in the job file
             remove_job_file(file_path=job_file)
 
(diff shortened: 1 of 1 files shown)


payu/status.py
```diff
@@ -296,6 +296,11 @@ def update_all_job_files(
         if job_info and exit_status and stage == "queued":
             # Job has exited, but is still marked as queued in the job file
             remove_job_file(file_path=job_file)
+
+        elif job_info and job_info.get("Jobs", {}).get(job_id, {}).get("job_state") == "F":
+            # Job is killed or deleted but still exists in the job file
+            remove_job_file(file_path=job_file)
+
```
```suggestion
        if job_info:
            job_dict = job_info.get("Jobs", {}).get(job_id,{})
            exit_status = job_dict.get("Exit_status")
            job_state = job_dict.get("job_state")
        else: 
            exit_status = None
            job_state = None

        if job_state == "F" and stage == "queued":
            # Job has exited, but is still marked as queued in the job file
            remove_job_file(file_path=job_file)
```

My suggestion is formatted a bit weird sorry, but I think we want to check the job has exited and the stage we have in the payu job file is still queued.

We don't want to delete job files where the stage is run/archive and job has exited.

payu/status.py
```diff
@@ -310,17 +315,13 @@ def update_all_job_files(
...
-                    file_path=job_file,
-                    data={f"payu_{job_type}_status": 1}
-                )
+                
+        elif run_status is None and stage != "queued":
+            # Run status isn't set, so job must have exited earlier
+            update_job_file(
+                file_path=job_file,
+                data={f"payu_{job_type}_status": 1}
+            )
```
jo-basevi: I think the previous `else:` can remain as PBS keeps logs only for about a day after the job has finished. For example, if I delete a job file on the queue, then after a couple days I run `payu status --update`,  then `job_info` will be None as the job logs got deleted.

jo-basevi (commented):
Thanks for adding this fix in! Let me know if you have questions about my comments

payu/status.py
Qian-HuiChen: I reverse it back in [6a790d9](https://github.com/payu-org/payu/pull/784/commits/6a790d9af1582de5d2a563165680f2ddf08ea8c8) :)

payu/status.py
Qian-HuiChen: I see. Logic is updated in [6a790d9](https://github.com/payu-org/payu/pull/784/commits/6a790d9af1582de5d2a563165680f2ddf08ea8c8)

jo-basevi (approved):
LGTM :)

PR #786, opened 2026-06-25
`payu status` handles PAYU_CURRENT_RUN when it is a string.

### LINKED ISSUE #785: payu status failing
I'm getting this
```
$ payu status
  File "/g/data/vk83/apps/base_conda/envs/payu-1.3.2/bin/payu", line 10, in <module>
    sys.exit(parse())
  File "/g/data/vk83/apps/base_conda/envs/payu-1.3.2/lib/python3.10/site-packages/payu/cli.py", line 64, in parse
    run_cmd(**args)
  File "/g/data/vk83/apps/base_conda/envs/payu-1.3.2/lib/python3.10/site-packages/payu/subcommands/status_cmd.py", line 48, in runcmd
    data = build_job_info(
  File "/g/data/vk83/apps/base_conda/envs/payu-1.3.2/lib/python3.10/site-packages/payu/status.py", line 232, in build_job_info
    latest_run_num = max(status_data["runs"].keys())
TypeError: '>' not supported between instances of 'str' and 'int'
```
for an ACCESS-OM3 job that's running fine if I look at `work/log/med.log`.

Control dir: `/home/156/aek156/payu/MC_25km_jra_iaf+wombatlite`

Qian-HuiChen:
`payu status` errors out when the PAYU_RUN_NUMBER is a string(e.g., "39") in a job file. This is generated by [run_userscript -> set_userscript_env_vars](https://github.com/payu-org/payu/blob/254d68ff4fd4b13c19aaebf20ecb1891902ca80e/payu/experiment.py#L1073-L1091).

This PR fixes the issue by ignoring string run_number entries, and only constructing `status_data` from runs with integer keys.
As a result, `payu status` will only display job information of `payu run`.

Closes #785


payu/status.py
```suggestion
            run_num = int(data["payu_current_run"])
```

Could convert it to an integer here? 

It'll be worth investigating why the queued files are saving payu_current_run as a string?

Qian-HuiChen: Implemented in [ed353d4](https://github.com/payu-org/payu/pull/786/commits/ed353d44a8cc08325385678ab50f3f6b84652146)

jo-basevi (approved):
Thanks, looks good to me.

I think slurm job IDs might also use increasing integers but would be something to bear in mind when we port to pawsey. 
An alternative option in future is to save a `queue_time` when job submitted and that can be compared with `start_time` when ordering jobs of the same run number.

PR #788, opened 2026-06-26
Telemetry: add file volume into job_file.

### LINKED ISSUE #780: Add file output volume calculation to payu telemetry
It would be useful to add the amount the size of the model diagnostic output to the telemetry as the amount of data produced by a run is a critical resource limitation.

@aekiss's run_summary tool calls out to the unix command `du` for this purpose:

https://github.com/aekiss/run_summary/blob/master/run_summary.py#L675-L680

Even though this is undoubtedly fast and simple there are some reasons to use a pure python approach

1. Platform independence
2. `du` calculates actual disk usage, so the [answer can depend on the physical block size](https://www.reddit.com/r/Python/comments/311ri9/are_there_times_when_using_subprocess_is/). This is undesirable
3. Security
4. Calling a subprocess might not be faster as there are overheads

There are python-only solutions in [this SO thread](https://stackoverflow.com/a/1392549), in which this comment

https://stackoverflow.com/questions/1392413/calculating-a-directorys-size-using-python#comment138317408_25574638

Suggests `du` is *a lot* faster, and using `du -xs --block-size=1` avoids the block size issue.

As this would need to be done at the end of the `archive` step, so still during the main PBS submission speed is a priority.

Qian-HuiChen:
This PR calculates the file volume of work directory right before archiving.
It also adds the file volume into job file for telemetry. 

Closes #780.

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -869,7 +869,7 @@ class Experiment(object):
             stage='archive',
             timings=self.timings
         )
-        # Check there is a work directory, otherwise bail
+        # Check there is a work directory, otherwise fail
         if not os.path.exists(self.work_sym_path):
             raise errors.PayuFileNotFoundError('No work directory to archive.')
 
@@ -927,7 +927,8 @@ class Experiment(object):
         telemetry.update_run_job_file(
             file_path=self.job_file,
             model_restart_datetimes=self.get_model_restart_datetimes(),
-            output_volume=get_size(self.output_path)
+            output_volume_gb=get_size(self.output_path),
+            restart_volume_gb=get_size(self.restart_path),
         )
 
         collate_config = self.config.get('collate', {})
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -876,9 +876,6 @@ class Experiment(object):
         os.makedirs(self.archive_path, exist_ok=True)
         make_symlink(self.archive_path, self.archive_sym_path)
 
-        # Calculate the file volume of the work directory and record in telemetry
-        work_dir_volume = get_size(self.work_path)
-
         # Remove work symlink
         if os.path.islink(self.work_sym_path):
             os.remove(self.work_sym_path)
@@ -897,13 +894,6 @@ class Experiment(object):
             raise errors.PayuRuntimeError('output path already exists')
 
         movetree(self.work_path, self.output_path)
-
-        # Record model restart datetimes in telemetry
-        telemetry.update_run_job_file(
-            file_path=self.job_file,
-            model_restart_datetimes=self.get_model_restart_datetimes(),
-            file_volume=work_dir_volume,
-        )
 
         # Remove any outdated restart files
         try:
@@ -932,6 +922,13 @@ class Experiment(object):
         archive_script = self.userscripts.get('archive')
         if archive_script:
             self.run_userscript(archive_script, 'archive')
+    
+        # Record model restart datetimes and output volume in telemetry
+        telemetry.update_run_job_file(
+            file_path=self.job_file,
+            model_restart_datetimes=self.get_model_restart_datetimes(),
+            output_volume=get_size(self.output_path)
+        )
 
         collate_config = self.config.get('collate', {})
         collating = collate_config.get('enable', True)
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -35,6 +35,7 @@ from payu.fsops import make_symlink, read_config, movetree
 from payu.fsops import list_sorted_archive_dirs
 from payu.fsops import run_script_command
 from payu.fsops import needs_subprocess_shell
+from payu.fsops import get_size
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 from payu.models import index as model_index
 from payu.runlog import Runlog
@@ -875,6 +876,9 @@ class Experiment(object):
         os.makedirs(self.archive_path, exist_ok=True)
         make_symlink(self.archive_path, self.archive_sym_path)
 
+        # Calculate the file volume of the work directory and record in telemetry
+        work_dir_volume = get_size(self.work_path)
+
         # Remove work symlink
         if os.path.islink(self.work_sym_path):
             os.remove(self.work_sym_path)
@@ -897,7 +901,8 @@ class Experiment(object):
         # Record model restart datetimes in telemetry
         telemetry.update_run_job_file(
             file_path=self.job_file,
-            model_restart_datetimes=self.get_model_restart_datetimes()
+            model_restart_datetimes=self.get_model_restart_datetimes(),
+            file_volume=work_dir_volume,
         )
 
         # Remove any outdated restart files
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -397,7 +397,7 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
 def get_size(work_path):
     """Return the total size of all files in the given path, in unit of GB."""
     total_size = 0
-    for dirpath, dirnames, filenames in os.walk(work_path):
+    for dirpath, _, filenames in os.walk(work_path):
         for f in filenames:
             fp = os.path.join(dirpath, f)
             # skip if it is symbolic link
@@ -405,5 +405,4 @@ def get_size(work_path):
                 total_size += os.path.getsize(fp)
 
     # Convert the total size from bytes to gigabytes
-    total_size_gb = (int(total_size) * ureg.byte).to(ureg.gibibyte).magnitude
-    return f"{total_size_gb}GB"
+    return (int(total_size) * ureg.byte).to(ureg.gibibyte).magnitude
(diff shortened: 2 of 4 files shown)


payu/experiment.py
```diff
@@ -902,6 +903,9 @@ def archive(self, force_prune_restarts=False):
         os.makedirs(self.archive_path, exist_ok=True)
         make_symlink(self.archive_path, self.archive_sym_path)
 
+        # Calculate the file volume of the work directory and record in telemetry
+        work_dir_volume = get_size(self.work_path)
```
aidanheerdegen: Calculating the size at this point will include all the restart files.

I'm in two minds. I think the most important information is the volume of the model diagnostic outputs, so what goes into the `outputXXX` directories. This fluctuates depending on the diagnostics being output, and is arguably the most important metric for users to know when planning an experiment.

The size of the restarts is important, but they don't change size, and they're often/usually pruned to keep only a subset of restarts, so are somewhat manageable.

Qian-HuiChen: Good point. Users can control how many restart directories to be kept. 
I modify the `get_size` function to exclude `RESTART` directory in commit [4db69ac](https://github.com/payu-org/payu/pull/788/commits/4db69acc00e5a5bcf7e70d0624071ccb8c2abd2a).

payu/fsops.py
```diff
@@ -382,4 +385,21 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
...
\ No newline at end of file
+            raise
+
+
+def get_size(work_path):
+    """Return the total size of all files in the given path, in unit of GB."""
+    total_size = 0
+    for dirpath, dirnames, filenames in os.walk(work_path):
+        # Skip restart directory
+        dirnames[:] = [d for d in dirnames if not d.startswith("RESTART")]
```
aidanheerdegen: The specific path used for restarts is model dependent, so I'm not sure this logic will work for all models.

If this calculation can't be done once the files have been moved into the `outputXXX` directory then I'd suggest reverting this change and include restarts and diagnostic outputs.

Sorry for wasting your time!

jo-basevi: Do we want to calculate the size of the output after the archive userscript? (e.g. at the end of the archive method)

aidanheerdegen: > Do we want to calculate the size of the output after the archive userscript? (e.g. at the end of the archive method)

Yes I think so. That is what would be most useful for users.

Qian-HuiChen: Thanks for clarifying this! Commit [78bc7a3](https://github.com/payu-org/payu/pull/788/commits/78bc7a3500560ffefe020d4b16b2f7f10ad56544) calculates the file size of output directory after the archive script is run. Would this be the most helpful for users?

payu/experiment.py
```diff
@@ -961,6 +957,13 @@ def archive(self, force_prune_restarts=False):
         if archive_script:
             self.run_userscript(archive_script, 'archive')
 
+        # Calculate the file volume of the output directory and record in telemetry
+        work_dir_volume = get_size(self.output_path)
+        telemetry.update_run_job_file(
+            file_path=self.job_file,
+            file_volume=work_dir_volume
```
jo-basevi: Could update the restart datetimes here too so the `.update_run_job_file` is called less? Similar to what you did in a previous commit

Qian-HuiChen: Can do ([0a3b7e0](https://github.com/payu-org/payu/pull/788/commits/0a3b7e0c91d3fd225fe91f66bc68372e6fa57762)). I was not sure if I should move it down because `archive_script` might fail and then we lose the information about `model_restart_datetime` in job file.

jo-basevi: True, that's a good point. Hmm.. if it becomes an issue in future maybe we could record restart datetimes and output sizes in a try-except-finally block?

Qian-HuiChen: Sounds good.

payu/experiment.py
```diff
@@ -956,6 +951,16 @@ def archive(self, force_prune_restarts=False):
         if archive_script:
             self.run_userscript(archive_script, 'archive')
 
+        # Calculate the file volume of the output directory
+        work_dir_volume = get_size(self.output_path)
```
aidanheerdegen: It's a small point, but `work_dir_volume` is no longer a good name for this. Maybe `output_volume` would be better?

payu/fsops.py
```diff
@@ -382,4 +385,19 @@ def _run_script(script_cmd: str, control_path: Path) -> None:
...
+    total_size = 0
+    for dirpath, dirnames, filenames in os.walk(work_path):
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            # skip if it is symbolic link
+            if not os.path.islink(fp):
+                total_size += os.path.getsize(fp)
+
+    # Convert the total size from bytes to gigabytes
+    total_size_gb = (int(total_size) * ureg.byte).to(ureg.gibibyte).magnitude
```
aidanheerdegen: Hooray for `pint`!

payu/telemetry.py
```diff
@@ -537,6 +538,8 @@ def update_run_job_file(
         run_info.update(extra_info)
     if timings:
         run_info.update(get_timings_isoformat(timings))
+    if file_volume:
+        run_info["file_volume"] = file_volume
```
aidanheerdegen: Just checking there are no [schema](https://github.com/ACCESS-NRI/schema) implications for this change @jo-basevi.

test/test_fsops.py
```diff
@@ -164,3 +164,32 @@ def test_list_sorted_archive_dirs(setup_test_dir):
...
+    write_dir_with_size(sub_dir, file_sizes)
+
+    # Create a symlink to an external file, should be ignored in size calculation
+    symlink_dir = tmpdir / "test_symlink_dir"
+    write_dir_with_size(symlink_dir, [2000])
+    (test_dir / "symlink_to_dir").symlink_to(symlink_dir)
+
+    # Assert the calculation gets an expected total size
+    expected_size = sum(file_sizes) * 2 / 1024 **3 # Convert bytes to GB
+    assert get_size(test_dir) == f"{expected_size}GB"
```
aidanheerdegen: Is this is a good opportunity to [refactor this test to use fixtures](https://github.com/payu-org/payu/issues/789) and parameterise the file sizes?

payu/telemetry.py
jo-basevi: Currently the schema is quite flexible as it allows additional fields. So when this is merged, it'll be present in the raw data but won't be in the processed tables until we parse the field out. 

The schema is currently stored within tracking-services (https://github.com/ACCESS-NRI/tracking-services/blob/main/server/payu/data_ingestion/schemas/telemetry/1-0-0.json), but I can add a copy to schema repository for visibility?

payu/experiment.py
Qian-HuiChen: Good point. I realize that work_dir_volume is called only once, so I simplify this more in commit [397d5c9](https://github.com/payu-org/payu/pull/788/commits/397d5c97fb771a01f51ea854a17f76db9a7a1654).
```
      telemetry.update_run_job_file(
            file_path=self.job_file,
            model_restart_datetimes=self.get_model_restart_datetimes(),
            output_volume=get_size(self.output_path)
        )
```

test/test_fsops.py
Qian-HuiChen: Yes, it is.
I change the `write_dir_with_size` as a fixture and parameterise the `file_size` in [397d5c9](https://github.com/payu-org/payu/pull/788/commits/397d5c97fb771a01f51ea854a17f76db9a7a1654). It doesn't simplify the code much but it is a good pattern.

payu/telemetry.py
aidanheerdegen: > The schema is currently stored within tracking-services (https://github.com/ACCESS-NRI/tracking-services/blob/main/server/payu/data_ingestion/schemas/telemetry/1-0-0.json), but I can add a copy to schema repository for visibility?

Probably isn't required, and would just be a maintenance burden to keep the in sync, but thanks.

payu/experiment.py
aidanheerdegen: So now I've found out CMIP7 require total amount written ... *sigh*

So I think I've changed my mind. Sorry!

See below.

payu/fsops.py
aidanheerdegen: So I think we should report `restart` and `output` sizes, then we can add them up for a total size if required. e.g. for CMIP reporting.

Qian-HuiChen: In commit [5fc27cf](https://github.com/payu-org/payu/pull/788/commits/5fc27cf039f271c9c10312cf1721b03edde30c22), payu records the `output` and `restart` size into telemetry as two floats, so users can add them up easily if needed.

payu/experiment.py
Qian-HuiChen: Don't worry. It is good that we catch this information before merging :D

aidanheerdegen (approved):
LGTM.

Qian-HuiChen:
https://github.com/ACCESS-NRI/tracking-services/pull/260 adds the `output` and `restart` directory volume to payu telemetry. I will wait for both of them to get approved and merge together.

PR #790, opened 2026-06-29
Add payu_jobs as protected in sync.

### LINKED ISSUE #760: Payu sync can delete running job files
There has been several runs with missing data in the Payu job files when payu is run in series with:
```
sync:
    remove_local_dirs: true
```

The `payu sync` jobs triggered by earlier runs may run concurrently with later `payu run` jobs. So when payu syncs the `payu_jobs/` directory, it deletes the folder.

To fix this issue, we should restrict the syncing the `payu_jobs` directory <= the current run specified, similar to the restarts and outputs in this PR: https://github.com/payu-org/payu/pull/740

Qian-HuiChen:
Previously, `payu sync` can come across to delete running job files if `remove_local_dirs` is set to `true`, and sync job happens after the next run started.

This PR adds `payu_jobs/` as a protected directory when syncing, so job files will not be removed locally even when 
```
sync:
    remove_local_dirs: true
```
Closes #760.

### DIFF
--- a/payu/sync.py
+++ b/payu/sync.py
@@ -232,15 +232,14 @@ class SyncToRemoteArchive():
 
     def build_cmd(self, source_path):
         """Given a source path to sync, return a rsync command"""
-        if source_path.protected:
+        cmd = self.base_rsync_cmd  
+        if not source_path.protected:
             # No local delete option for protected paths
-            cmd = f'{self.base_rsync_cmd} {self.excludes} '
-        elif source_path.is_log_file:
-            cmd = f'{self.base_rsync_cmd} {self.remove_files} '
-        else:
-            cmd = f'{self.base_rsync_cmd} {self.excludes} {self.remove_files} '
+            cmd += f' {self.remove_files}'
+        if not source_path.is_log_file:
+            cmd += f' {self.excludes}'
 
-        cmd += f'{source_path.path} {self.destination_path}'
+        cmd += f' {source_path.path} {self.destination_path}'
         return cmd
 
     def run_cmd(self, source_path):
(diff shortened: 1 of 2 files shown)


payu/sync.py
jo-basevi: So the `.is_log_file` basically means we don't add the excludes flag. The excludes flag typically includes regex patterns that stop files being synced. 

I think it'll be useful to always ensure the job file is synced, so maybe the logic could be something like:

command = base_cmd
if not protected: 
    add self.remove files flag to command
if not logfiles:
    add self.exclude files flag to command 

What do you think?

jo-basevi (dismissed):
The added changes fixes the bug and the tests look good to me!

payu/sync.py
Qian-HuiChen: Do we want payu_jobs to have the `--excludes`?

Qian-HuiChen: I see that your proposed changes will have the same effect as the current code. My question is: do we want to make payu_jobs with `protected=True, is_log_file=True` so it won't have `--excludes`, not `--remove_files`? Is that why you propose the changes?

jo-basevi: Yes, sorry, having both `protected=True, is_log_file=True` is what I meant

jo-basevi: So the command will be similar to just `f'{self.base_rsync_cmd}`

Qian-HuiChen: Got ya! Updated in [2cfd194](https://github.com/payu-org/payu/pull/790/commits/2cfd194c9f39f447cf73973fff2afe891f3acb58)

jo-basevi: Thanks!

jo-basevi (approved):
Nice tests btw! Changes look good to me :)

PR #795, opened 2026-07-01
Raise error when archive not found unless `--new-uuid` is specified.

### LINKED ISSUE #776: shortpath change triggers new uuid and inconsistent archive name
If we have an experiment directory containing `archive` and then `shortpath` is changed (say, from `/scratch/x77` to `/scratch/v45`), `payu sweep`, `payu setup` (and presumably others) apparently can't recognise the existing `archive` and trigger a new uuid:
```
[WARNING] No pre-existing archive found. Generating a new uuid
Updated metadata. Experiment UUID: 1e2b927f-75fa-451f-b450-a6babe0c9ab6
```
and commit a metadata change with the new experiment name and uuid
```
$ git diff HEAD^ HEAD
diff --git a/metadata.yaml b/metadata.yaml
index 89227ee..f04f258 100644
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -30,10 +30,10 @@ license: CC-BY-4.0
 model: ACCESS-OM3
 url: https://github.com/ACCESS-NRI/access-om3-configs
 version: 0.0
-experiment_uuid: 5224aa43-a349-4560-9640-fbaf56a4212e
-name: access-om3-configs-CHUCK-5224aa43
+experiment_uuid: e8f1c851-bcc8-4538-847f-9567048a9cf2
+name: access-om3-configs-CHUCK-e8f1c851
 schema_version: 1-0-3
```
but the archive symlink path is not changed, so `payu setup` gives us inconsistent `work` and `archive` names
```
archive -> /scratch/x77/aek156/access-om3/archive/access-om3-configs-CHUCK-5224aa43
work -> /scratch/v45/aek156/access-om3/work/access-om3-configs-CHUCK-e8f1c851
```

Subsequent `payu sweep; payu setup` does not alter the uuid further, so payu now recognises `archive`, but its path remains inconsistent with the new experiment name.

This seems very odd, for several reasons

1. why doesn't payu recognise an existing archive?
2. why does it recognise the archive after the uuid is changed, even though the path is unaltered and so doesn't match the new experiment name?
3. why isn't the `archive` symlink updated when `shortpath` is changed? Both the shortpath part and experiment name are unaltered.

More intuitive behaviour (at least for me) for a `shortpath` change would be
- don't change uuid when  `shortpath` changes, ie no change in experiment name if `archive` exists
- create and symlink to a new archive dir using the new `shortpath` and copy or symlink the latest restart (if any) into it. Perhaps have a warning if the old archive dir is non-empty, stating its full path so the user can find previous outputs/restarts.

Qian-HuiChen:
Previously, Payu automatically generated a new UUID and print a warning when no existing archive was found.

This PR changes that behavior to raise an error when an archive is not found. 
If the user explicitly provides the `--new-uuid` flag, Payu will instead print a warning and proceed to generate a new UUID.

Closes #776.

### DIFF
--- a/payu/branch.py
+++ b/payu/branch.py
@@ -143,8 +143,7 @@ def checkout_branch(branch_name: str,
                     control_path: Optional[Path] = None,
                     model_type: Optional[str] = None,
                     lab_path: Optional[Path] = None,
-                    parent_experiment: Optional[str] = None,
-                    new_uuid: bool = False) -> None:
+                    parent_experiment: Optional[str] = None) -> None:
     """Checkout branch, setup metadata and add symlinks
 
     Parameters
@@ -216,8 +215,7 @@ def checkout_branch(branch_name: str,
     # Setup Metadata
     is_new_experiment = is_new_experiment or is_new_branch
     metadata.setup(keep_uuid=keep_uuid,
-                   is_new_experiment=is_new_experiment,
-                   new_uuid=new_uuid)
+                   is_new_experiment=is_new_experiment)
 
     # Gets valid prior restart path
     prior_restart_path = None
@@ -347,8 +345,7 @@ def clone(repository: str,
                             model_type=model_type,
                             lab_path=lab_path,
                             parent_experiment=parent_experiment,
-                            start_point=start_point,
-                            new_uuid=True)
+                            start_point=start_point)
         else:
             # Checkout branch
             if branch is None:
--- a/payu/branch.py
+++ b/payu/branch.py
@@ -143,7 +143,8 @@ def checkout_branch(branch_name: str,
                     control_path: Optional[Path] = None,
                     model_type: Optional[str] = None,
                     lab_path: Optional[Path] = None,
-                    parent_experiment: Optional[str] = None) -> None:
+                    parent_experiment: Optional[str] = None,
+                    new_uuid: bool = False) -> None:
     """Checkout branch, setup metadata and add symlinks
 
     Parameters
@@ -215,7 +216,8 @@ def checkout_branch(branch_name: str,
     # Setup Metadata
     is_new_experiment = is_new_experiment or is_new_branch
     metadata.setup(keep_uuid=keep_uuid,
-                   is_new_experiment=is_new_experiment)
+                   is_new_experiment=is_new_experiment,
+                   new_uuid=new_uuid)
 
     # Gets valid prior restart path
     prior_restart_path = None
@@ -345,7 +347,8 @@ def clone(repository: str,
                             model_type=model_type,
                             lab_path=lab_path,
                             parent_experiment=parent_experiment,
-                            start_point=start_point)
+                            start_point=start_point,
+                            new_uuid=True)
         else:
             # Checkout branch
             if branch is None:
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -42,6 +42,10 @@ SCHEMA_COMMIT_HASH = "cff183437134592723b09af6620e5cb190abeb22"
 SCHEMA_URL = f"https://raw.githubusercontent.com/ACCESS-NRI/schema/{SCHEMA_COMMIT_HASH}/au.org.access-nri/model/output/experiment-metadata/{SCHEMA_VERSION}.json"
 placeholder_text = "__REPLACE_ME__"
 
+no_archive_msg = """No pre-existing archive is found for this experiment.
+Payu will generate a new experiment UUID and create a new archive directory.
+To proceed, please rerun the command with the --new-uuid flag."""
+
 class MetadataWarning(Warning):
     pass
 
@@ -111,7 +115,8 @@ class Metadata:
 
     def setup(self,
               is_new_experiment: bool = False,
-              keep_uuid: bool = False) -> None:
+              keep_uuid: bool = False,
+              new_uuid: bool = False) -> None:
         """Set UUID and experiment name.
 
         Parameters:
@@ -120,6 +125,8 @@ class Metadata:
             is_new_experiment: bool, default False
                 If not keep_uuid, generate a new UUID and a branch-uuid aware
                 experiment name. This is set in payu.branch.checkout_branch.
+            new_uuid: bool, default False
+                User confirm to generate a new UUID.
         Return: None
 
         Note: Experiment name is the name used for the work and archive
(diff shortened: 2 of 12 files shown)


payu/branch.py
```diff
@@ -153,7 +153,8 @@ def checkout_branch(branch_name: str,
                     control_path: Optional[Path] = None,
                     model_type: Optional[str] = None,
                     lab_path: Optional[Path] = None,
-                    parent_experiment: Optional[str] = None) -> None:
+                    parent_experiment: Optional[str] = None,
+                    new_uuid: bool = False) -> None:
```
jo-basevi: Instead of `new_uuid`, should re-use the existing `is_new_experiment` function argument:

```
    is_new_experiment: bool, default False
        Create new uuid for this experiment
```

This will require no change then to the clone command, as it'll always generate a new uuid (unless keep_uuid is passed along)

payu/metadata.py
```diff
@@ -111,7 +115,8 @@ def read_file(self) -> CommentedMap:
 
     def setup(self,
               is_new_experiment: bool = False,
-              keep_uuid: bool = False) -> None:
+              keep_uuid: bool = False,
+              new_uuid: bool = False) -> None:
```
jo-basevi: I think is_new_experiment can also be re-used instead of new_uuid here.

payu/metadata.py
```diff
@@ -200,13 +209,17 @@ def set_experiment_name(self,
...
+        elif new_uuid:
             # No archive exists - Detecting new experiment
             warnings.warn(
                 "No pre-existing archive found. Generating a new uuid",
                 MetadataWarning
             )
             self.set_new_uuid(is_new_experiment=True)
+        else:
+            # No archive exists and user did not specify a new or old UUID
+            raise RuntimeError(f"{no_archive_msg}")
```
jo-basevi: Maybe for debugging could provide the laboratory path? So it'll be easier to spot if the project has changed etc.

payu/metadata.py
```diff
@@ -42,6 +42,10 @@
 SCHEMA_URL = f"https://raw.githubusercontent.com/ACCESS-NRI/schema/{SCHEMA_COMMIT_HASH}/au.org.access-nri/model/output/experiment-metadata/{SCHEMA_VERSION}.json"
 placeholder_text = "__REPLACE_ME__"
 
+no_archive_msg = """No pre-existing archive is found for this experiment.
+Payu will generate a new experiment UUID and create a new archive directory.
+To proceed, please rerun the command with the --new-uuid flag."""
```
jo-basevi: This flag has only been added to payu setup, payu run and payu checkout commands - but this error could still be raised in other commands. Maybe suggest rerunning just the `payu setup` or `payu checkout` command?

jo-basevi (commented):
Thanks @Qian-HuiChen, I think this will be a good change. For users not using `payu clone`/ `payu checkout -b` will notice that will need to explicitly create a new uuid every time they start a new experiment (so will need to definitely advertise this change when merged), but hopefully it'll give users more trust that payu is using the archive they expect.

payu/metadata.py
Qian-HuiChen: Updated in commit [89fa57a](https://github.com/payu-org/payu/pull/795/commits/89fa57adb1599a085437893027c111fd3250d5bb)

payu/metadata.py
Qian-HuiChen: Updated in commit [89fa57a](https://github.com/payu-org/payu/pull/795/commits/89fa57adb1599a085437893027c111fd3250d5bb)

payu/metadata.py
Qian-HuiChen: Updated in commit [89fa57a](https://github.com/payu-org/payu/pull/795/commits/89fa57adb1599a085437893027c111fd3250d5bb)

payu/branch.py
Qian-HuiChen: Updated in commit [89fa57a](https://github.com/payu-org/payu/pull/795/commits/89fa57adb1599a085437893027c111fd3250d5bb)

Qian-HuiChen:
In [these lines](https://github.com/payu-org/payu/pull/795/changes#diff-d8a418481defcd6e55ad533ad5e3a410441729fbc719565383585748876a7913R145-R147), I make payu run `set_new_uuid` if the current uuid is `None`. Because I am thinking the case of the brand-new start, it saves user to add the flag. And payu has to generate a new uuid anyway because there is none existent. 

I don't know how to fix the current failed test in test_matadata.py though @jo-basevi

payu/metadata.py
```diff
@@ -137,7 +142,10 @@ def setup(self,
             if self.uuid is None and not is_new_experiment:
                 warnings.warn("No experiment uuid found in metadata. "
                               "Generating a new uuid", MetadataWarning)
-            self.set_new_uuid(is_new_experiment=is_new_experiment)
+                self.set_new_uuid(is_new_experiment=True)
```
jo-basevi: If someone runs with an old version of payu, the archive directory will have the same name as the control directory and they might not have a UUID set in the `metadata.yaml`. 

If they switch to a later version of payu, and run `payu setup` or `payu run`, we want payu to generate a new UUID but still use the old "legacy" archive name. It's technically not a new experiment, we are just adding in a new UUID if one doesn't already exist.

aidanheerdegen: Would it help to add a code comments to explain what cases might fall through to this line.

payu/metadata.py
```diff
@@ -42,6 +43,10 @@
 SCHEMA_URL = f"https://raw.githubusercontent.com/ACCESS-NRI/schema/{SCHEMA_COMMIT_HASH}/au.org.access-nri/model/output/experiment-metadata/{SCHEMA_VERSION}.json"
 placeholder_text = "__REPLACE_ME__"
 
+no_archive_msg = """
+Payu will generate a new experiment UUID and create a new archive directory.
```
aidanheerdegen: I think the wording is confusing when I think about it added to error message:

> Current archive path is not found: Payu will generate a new experiment UUID and create a new archive directory.
> To proceed, please rerun `payu setup` or `payu run or `payu checkout`` with the --new-uuid flag.

Payu will not generate a new experiment ID *unless* the suggests commands are run.

The first sentence is describing the previous behaviour, which this PR is changing.

Maybe "Payu needs to generate an experiment UUID and create a new archive directory"?

Qian-HuiChen: Sounds good. Updated in [d381d47](https://github.com/payu-org/payu/pull/795/commits/d381d47719f38911f3453e9cd5ba49fe30052062)

payu/metadata.py
Qian-HuiChen: Got ya! I removed the warning block since it is not needed anymore. I also added some comments for future reference in [d381d47](https://github.com/payu-org/payu/pull/795/commits/d381d47719f38911f3453e9cd5ba49fe30052062)

jo-basevi (approved):
Had a read through the code and looks good to me! Note I haven't run any manual tests.

aidanheerdegen (approved):
Looks good, thanks.

PR #801, opened 2026-07-08
Add `sync:base_path` to PBS storage flag if `sync:path` is not set.

### LINKED ISSUE #798: `submit` currently doesn't add the sync `base_path` to the PBS storage
The [`submit`](https://github.com/payu-org/payu/blob/d226b11c46b08ffca74163f01f41822fa5cd5019/payu/schedulers/pbs.py#L343)  function searches through a paths in the `config.yaml` file in order to set the required PBS storage flags. It currently checks the `sync:path` option, however doesn't check the newer `sync:base_path` option

https://github.com/payu-org/payu/blob/d226b11c46b08ffca74163f01f41822fa5cd5019/payu/schedulers/pbs.py#L436-L438

As a result, using the `base_path` option can cause jobs to fail, unless you explicitly add the destination area to the storage flags. I think would it make sense to add `sync:base_path` to the list of paths that are searched?

Qian-HuiChen:
Previously payu did not add `sync:base_path` to PBS storage, leading to fail run.

This PR adds the mount of `sync:base_path` to PBS storage. This only happens if `sync:path` is not set, otherwise, `sync:path` is used.

Closes #798.

Manual tests
1. `sync:base_path` points to an unavailable project, and `sync:path` not set. Error out before the PBS submission.
2. `sync:base_path` points to a correct project, but `sync:path` points to an unavailable project. Error out before the PBS submission -> payu priorities `sync:path` and found no access to the assigned project.
3. `sync:base_path` points to a correct project -> new storage is addd to PBS submission.

### DIFF
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -434,7 +434,8 @@ class PBS(Scheduler):
         module_use_paths = pbs_config.get('modules', {}).get('use', [])
         extra_search_paths.extend(module_use_paths)
 
-        remote_sync_directory = pbs_config.get('sync', {}).get('path', None)
+        sync_config = pbs_config.get('sync', {})
+        remote_sync_directory = sync_config.get('path', None) or sync_config.get('base_path', None)
         if remote_sync_directory is not None:
             extra_search_paths.append(remote_sync_directory)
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
Looks good!

Issue #767, opened 2026-06-15
Replace submission layer with HPCpy


bschroeter:
As part of a larger body of work surrounding portability, we're looking to replace the queue submission layer with HPCpy (https://github.com/ACCESS-NRI/hpcpy).

Tasks involved:
- [ ] Identify all locations where HPCpy should be implemented
- [ ] Ensure existing functionality is retained on Gadi (PBS)
- [ ] Generalise workflow to work on Pawsey (SLURM)

If there are complimentary changes required on the HPCpy end, we will link the issues back to here.

I know that there may be existing issues/PRs from previous investigations - in order to maintain clarity, they will be closed/parked in favour of this issue.

aidanheerdegen:
This existing issue has a lot of useful information and links to other issues:

https://github.com/payu-org/payu/issues/568

aidanheerdegen:
Required use-cases

- [Profiling](https://github.com/payu-org/payu/issues/544) (customised PBS script)
- [Cross-platform (customised scripts, cross platform configs or separate queue config from config.yaml)](https://github.com/payu-org/payu/issues/586)
- [Energy monitoring (injecting extra routines into every compute node)](https://github.com/payu-org/payu/issues/641)
- [Debug option](https://github.com/payu-org/payu/issues/340)
- Run without PBS/MPI: [run CABLE locally (personal laptops?)](https://github.com/CABLE-LSM/CABLE/issues/698)

Required functionality
- [Proper module loading](https://github.com/payu-org/payu/issues/522)
- Handling job dependency strategy, e.g https://github.com/payu-org/payu/issues/463
- Support queue/cpus-per-node/memory-per-node check logic
- [Graceful recovery from transient PBS errors](https://github.com/payu-org/payu/issues/43)
- Write PBS script to work directory, would need this to be archived

bschroeter:
@bschroeter to set up a project board, discrete issues for each feature etc.

bschroeter:
@bschroeter to deploy hpcpy to PyPi for testing purposes.

aidanheerdegen:
I realise I misspoke and might have confused others about the order and scope of operations. I think I was confused myself.

`payu` currently submits itself to PBS in the `qsub` command, e.g.
```bash
[...]

PR #802, opened 2026-07-13
Replace PBS submission layer with HPCpy.

Qian-HuiChen:
This PR replaces the submission layer with HPCpy as the first step of resolving #767.

In this PR, PBS job submission is migrated from directly calling `subprocess` to using `hpcpy.client.pbs.PBSClient()`. 

A `--dry-run` flag is added to `payu run`, `payu collate`, and `payu sync`. When enabled, the submission command is generated and returned without submitting the job.

Slurm submission is not migrated to HPCpy in this PR for now.

Acknowledgement to @jo-basevi's draft PR #619.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -209,35 +209,56 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     return payu_env_vars
 
 
-def submit_job(script, config, vars=None, expt=None, current_run=None, type=None):
+def submit_job(script, config, vars=None, expt=None, current_run=None, type=None, dry_run=False):
     """Submit a userscript the scheduler and return the job ID"""
 
     # TODO: Temporary stub to replicate the old approach
     sched_name = config.get('scheduler', DEFAULT_SCHEDULER_CONFIG)
     sched_type = scheduler_index[sched_name]
     sched = sched_type()
-    cmd = sched.submit(script, config, vars)
-    print(cmd)
 
     try:
-        result = subprocess.run(shlex.split(cmd), capture_output=True, check=True, text=True)
+        if sched_name == 'pbs':
+            # Use HPCpy to submit the job in PBS
+            job_or_cmd = sched.submit(script, config, vars, dry_run=dry_run)
+
+            if dry_run:
+                # If dry_run is True, print out the submission command and exit
+                print(f"Printing job submission command without executing it:\n"
+                    f"    {job_or_cmd}")
+                return None
+            else:
+                # Otherwise, print the job ID after submission
+                job_id = job_or_cmd.id
+                print(f"Job submitted with ID: {job_id}")
+        
+        elif sched_name == 'slurm':
+            # Slurm: payu generates the sbatch command and call subprocess.run() to submit the job
+            cmd = sched.submit(script, config, vars)
+            result = subprocess.run(shlex.split(cmd), capture_output=True, check=True, text=True)
+
+            job_id = result.stdout.strip().split()[-1]
+            print(f"Job submitted with ID: {job_id}")
+
+        else:
+            warnings.warn(f"Unsupported scheduler type: {sched_name}. Job submission skipped.")
+            return None
 
     except subprocess.CalledProcessError as e:
-        error_msg = ("Error occurred while submitting job.\n")
-
-        if e.returncode:
-            error_msg += f"Exit code: {e.returncode}\n"
-        if e.stdout:
-            error_msg += f"STDOUT: {e.stdout}\n"
-        if e.stderr:
-            error_msg += f"STDERR: {e.stderr}"
-
-        raise errors.PayuRuntimeError(error_msg)
-
-    # Decode stdout and extract the job ID which is last for both PBS and Slurm
-    result = result.stdout.strip()
-    print(result)
-    job_id = result.split()[-1]
+        # Catch errors from subprocess.run() for Slurm
+        raise errors.PayuRuntimeError(
+            "Error occurred while submitting job.\n"
+            f"Exit code {e.returncode}.\n"
+            f"STDOUT: {e.output}\n"
+            f"STDERR: {e.stderr}"
+        )
+    
+    except Exception as e:
+        # Catch other exceptions from PBS submission
+        raise errors.PayuRuntimeError(
+            "Error occurred while submitting a job to scheduler {sched_name}.\n"
+            f"Error: {str(e)}"
+        ) from e
 
     if expt is not None:
 
@@ -255,7 +276,6 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
             metadata=expt.metadata,
             current_run=current_run,
         )
-
 
     return job_id
 
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -257,7 +257,7 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
     except Exception as e:
         # Catch other exceptions from PBS submission
         raise errors.PayuRuntimeError(
-            "Error occurred while submitting a job to scheduler {sched_name}.\n"
+            f"Error occurred while submitting a job to {sched_name} scheduler.\n"
             f"Error: {str(e)}"
         ) from e
 
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -21,7 +21,6 @@ from pathlib import Path
 
 # Local imports
 import payu
-import payu.envmod as envmod
 from payu.fsops import is_conda
 from payu.models import index as supported_models
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
@@ -225,12 +224,12 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
             if dry_run:
                 # If dry_run is True, print out the submission command and exit
                 print(f"---- Dry run (submission skipped) -----\n"
-                    f"Submission command: {job_or_cmd}")
+                    f"Generated command: {job_or_cmd}")
                 return None
             else:
                 # Print the job ID and command after submission 
                 print(f"------ Job submitted ------\n"
-                    f"Submission command: {job_or_cmd.history[0]}")
+                    f"Submitted command: {job_or_cmd.history[0]}")
                 job_id = job_or_cmd.id
                 print(f"Job ID: {job_id}")
         
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -224,16 +224,15 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
 
             if dry_run:
                 # If dry_run is True, print out the submission command and exit
-                print(f"Printing job submission command without executing it:\n"
-                    f"    {job_or_cmd}")
+                print(f"---- Dry run (submission skipped) -----\n"
+                    f"Submission command: {job_or_cmd}")
                 return None
             else:
-                
-                # Uncomment this after HPCpy PR#71 is included in a release
-                # print(f"Submitted job: {job_or_cmd.history[0]}") 
-                # Print the job ID after submission
+                # Print the job ID and command after submission 
+                print(f"------ Job submitted ------\n"
+                    f"Submission command: {job_or_cmd.history[0]}")
                 job_id = job_or_cmd.id
-                print(f"Job submitted with ID: {job_id}")
+                print(f"Job ID: {job_id}")
         
         elif sched_name == 'slurm':
             # Slurm: payu generates the sbatch command and call subprocess.run() to submit the job
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -228,7 +228,10 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
                     f"    {job_or_cmd}")
                 return None
             else:
-                # Otherwise, print the job ID after submission
+                
+                # Uncomment this after HPCpy PR#71 is included in a release
+                # print(f"Submitted job: {job_or_cmd.history[0]}") 
+                # Print the job ID after submission
                 job_id = job_or_cmd.id
                 print(f"Job submitted with ID: {job_id}")
         
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -173,7 +173,8 @@ def check_user_modulepaths(user_modules, user_modulepaths):
 
     for modulefile in user_modules:
         # Check modulefile exists and is unique or has an exact match
-        check_modulefile(modulefile)
+        # Pass modulepaths so the check uses the correct search paths
+        check_modulefile(modulefile, modulepaths=user_modulepaths)
 
     # Get what is added to PATH if user modules are loaded, without actually loading them
     added_paths = set()
@@ -188,10 +189,20 @@ def check_user_modulepaths(user_modules, user_modulepaths):
     return (added_modules, added_paths)
 
 
-def check_modulefile(modulefile: str) -> None:
+def check_modulefile(modulefile: str, modulepaths: list = None) -> None:
     """Given a modulefile, check if modulefile exists, and there is
-    a unique modulefile available - e.g. if it's version is specified"""
-    output = run_module_cmd("avail --terse", modulefile).stderr
+    a unique modulefile available - e.g. if it's version is specified.
+    
+    Parameters
+    ----------
+    modulefile : str
+        The modulefile to check
+    modulepaths : list, optional
+        List of module paths to search. If provided, these paths are used
+        during the check without permanently modifying MODULEPATH.
+    """
+
+    output = run_module_cmd("avail --terse", modulefile, modulepaths=modulepaths).stderr
 
     # Extract out the modulefiles available - strip out lines like:
     # /apps/Modules/modulefiles:
@@ -213,8 +224,38 @@ def check_modulefile(modulefile: str) -> None:
         )
 
 
-def run_module_cmd(subcommand, *args):
-    """Wrapper around subprocess module command that captures output"""
+def run_module_cmd(subcommand, *args, modulepaths=None):
+    """Wrapper around subprocess module command that captures output.
+    
+    Parameters
+    ----------
+    subcommand : str
+        The module subcommand (e.g., 'avail', 'load')
+    *args : str
+        Arguments to pass to the subcommand
+    modulepaths : list, optional
+        List of module paths to prepend to MODULEPATH for this command only.
+        Does not modify the environment permanently.
+    
+    Returns
+    -------
+    subprocess.CompletedProcess
+        The result of running the module command
+    """
     modulecmd = f"{os.environ['MODULESHOME']}/bin/modulecmd bash"
-    command = f"{modulecmd} {subcommand} {' '.join(args)}"
-    return subprocess.run(command, shell=True, text=True, capture_output=True)
+    
+    # Build the full bash command
+    bash_commands = []
+    
+    if modulepaths:
+        # Properly evaluate the module use command output before running the main command
+        bash_commands.append(f"eval $({modulecmd} use {' '.join(modulepaths)})")
+    
+    bash_commands.append(f"{modulecmd} {subcommand} {' '.join(args)}")
+    
+    # Run in bash shell to properly evaluate module commands
+    return subprocess.run(
+        ['/bin/bash', '-c', ' && '.join(bash_commands)],
+        text=True,
+        capture_output=True
+    )
--- a/payu/envmod.py
+++ b/payu/envmod.py
@@ -128,51 +128,59 @@ def lib_update(required_libs, lib_name):
     return ''
 
 
-def setup_user_modules(user_modules, user_modulepaths):
-    """Run module use + load commands for user-defined modules. Return
-    tuple containing a set of loaded modules and paths added to
-    LOADEDMODULES and PATH environment variable, as result of
-    loading user-modules"""
+def added_path_by_user_module(user_modulepaths, user_module):
+    """Get what paths are added to the environment if a user module is loaded"""
+    cmd = f"""module purge && env | grep '^PATH=' | sort""" 
+    args = ["/bin/bash", "-l", "-c", cmd]
+    result = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, shell=False)
+    stdout, _ = result.communicate()
+    previous_path_string = stdout.strip().replace('PATH=', '', 1)
+    previous_paths = set(previous_path_string.split(os.pathsep))
 
+    cmd = f"""
+module purge && module use {' '.join(user_modulepaths)} && module load {user_module} && env | grep '^PATH=' | sort"""
+    args = ["/bin/bash", "-l", "-c", cmd]
+    result = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, shell=False)
+    stdout, _ = result.communicate()
+    new_path_string = stdout.strip().replace('PATH=', '', 1)
+    new_paths = set(new_path_string.split(os.pathsep))
+
+    paths_added = new_paths - previous_paths
+
+    return paths_added
+
+def check_user_modulepaths(user_modules, user_modulepaths):
+    """ Check user-defined modules and filepaths.
+    Return a set of paths added by loading the user modules, without actually loading them. """
     if 'MODULESHOME' not in os.environ:
         print(
             'payu: warning: No Environment Modules found; ' +
             'skipping running module use/load commands for any module ' +
             'directories/modulefiles defined in config.yaml')
         return (None, None)
-
-    # Add user-defined directories to MODULEPATH
+    
+    # Check user-defined module paths exist
     for modulepath in user_modulepaths:
         if not os.path.isdir(modulepath):
             raise ValueError(
                 f"Module directory is not found: {modulepath}" +
                 "\n Check paths listed under `modules: use:` in config.yaml")
 
-        module('use', modulepath)
-
-    # First un-load all user modules, if they are loaded, so can store
-    # LOADEDMODULES and PATH to compare to later
-    for modulefile in user_modules:
-        if run_module_cmd("is-loaded", modulefile).returncode == 0:
-            module('unload', modulefile)
-    previous_loaded_modules = os.environ.get('LOADEDMODULES', '')
-    previous_path = os.environ.get('PATH', '')
-
     for modulefile in user_modules:
         # Check modulefile exists and is unique or has an exact match
         check_modulefile(modulefile)
 
-        # Load module
-        module('load', modulefile)
-
-    # Create a set of paths and modules loaded by user modules
-    loaded_modules = os.environ.get('LOADEDMODULES', '')
-    path = os.environ.get('PATH', '')
-    loaded_modules = set(loaded_modules.split(':')).difference(
-        previous_loaded_modules.split(':'))
-    paths = set(path.split(':')).difference(set(previous_path.split(':')))
-
-    return (loaded_modules, paths)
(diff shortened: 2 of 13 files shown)


Qian-HuiChen:
Commit [cd2e166](https://github.com/payu-org/payu/pull/802/commits/cd2e1664855734da747bc611987cf1ce2f7116df) use HPCpy to load user's modules. The `envmod.check_user_modulepaths` checks what paths are added to PATH in a separate shell without loading the modules.

Even though the [HPCpy module PR](https://github.com/ACCESS-NRI/hpcpy/pull/68) is merged, it is not released yet. This might fail when deploying to `payu/dev`

payu/cli.py
```diff
@@ -209,35 +209,56 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
...
+        else:
+            warnings.warn(f"Unsupported scheduler type: {sched_name}. Job submission skipped.")
+            return None
 
-    # Decode stdout and extract the job ID which is last for both PBS and Slurm
-    result = result.stdout.strip()
-    print(result)
-    job_id = result.split()[-1]
+    except subprocess.CalledProcessError as e:
+        # Catch errors from subprocess.run() for Slurm
```
aidanheerdegen, line 249: So we don't get an exception bubbling up from HPCPy? How is the error reported in this case?

payu/envmod.py
```diff
@@ -128,51 +128,59 @@ def lib_update(required_libs, lib_name):
...
+    previous_paths = set(previous_path_string.split(os.pathsep))
+
+    cmd = f"""
+module purge && module use {' '.join(user_modulepaths)} && module load {user_module} && env | grep '^PATH=' | sort"""
+    args = ["/bin/bash", "-l", "-c", cmd]
+    result = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, shell=False)
+    stdout, _ = result.communicate()
+    new_path_string = stdout.strip().replace('PATH=', '', 1)
+    new_paths = set(new_path_string.split(os.pathsep))
+
```
aidanheerdegen: There is some duplication here that I think could be avoided and make it more testable e.g. 
```python
def setup_modules(modpaths=None, module=None):
    """
    Return module commands required to load module. Return value is a string.
    Note arguments are optional to support case where no modules are loaded 
    """
    cmds = ["module purge"]
    if modpaths:
        cmds.append('module use {paths}'.format(paths=' '.join(modpaths)))
    if module:
        cmds.append('module load {module}')
    return " && ".join(cmds)

def get_paths(cmd):
    """
    Run cmd in a sub-process and return it's PATH environment as a set of paths
    """
    cmd = f"{cmd} && env | grep '^PATH=' | sort"
    with subprocess.Popen(["/bin/bash", "-l", "-c", cmd], stdout=subprocess.PIPE, 
                           stderr=subprocess.PIPE, text=True, shell=False) as proc:
        stdout, _ = proc.communicate()
    return set(stdout.strip().replace('PATH=', '', 1).split(':'))

def paths_added_by_user_module(user_modulepaths, user_module):
    """
    Return paths (set) added to environment when user module loaded
    """
    previous_paths = get_paths(setup_modules(user_modulepaths, None))
    new_paths = get_paths(setup_modules(user_modulepaths, user_module))

    return new_paths - previous_paths
```
Notes:
1. It is more lines of code, but more testable and potentially reusable
2. I think it is more "understandable", but that is just my opinion
3. I have not tested this! Just taken your code and rearranged it
4. Would using `subprocess.check_output` make this simpler?

payu/schedulers/pbs.py
```diff
@@ -341,33 +348,26 @@ def validate_memory_with_queue_limits(cls, pbs_mem: str, queue: str, n_cpus: int
...
         if python_exe is None:
             python_exe = sys.executable
 
         pbs_flags = []
 
-        pbs_queue = pbs_config.get('queue', 'normal')
-        pbs_flags.append('-q {queue}'.format(queue=pbs_queue))
-
         pbs_project = pbs_config.get('project', os.environ['PROJECT'])
+        pbs_flags.append('-P {project}'.format(project=pbs_project))
```
aidanheerdegen: Could move this into the `if` logic below to make it clear that this will only be set (and used) if the project is in the users groups.

payu/schedulers/pbs.py
```diff
@@ -397,12 +399,13 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
...
 
-        # Append environment variables to qsub command
+        # Add environment variables to qsub command
         # TODO: Support full export of environment variables: `qsub -V`
-        pbs_vstring = ','.join('{0}={1}'.format(k, v)
-                               for k, v in pbs_vars.items())
-        pbs_flags.append('-v ' + pbs_vstring)
+        # pbs_vstring = ','.join('{0}={1}'.format(k, v)
+        #                        for k, v in pbs_vars.items())
+        # pbs_flags.append('-v ' + pbs_vstring)
```
aidanheerdegen: I wonder if this is still relevant. Note I have [made an issue](https://github.com/payu-org/payu/issues/807) for all the `TODO`s

payu/schedulers/pbs.py
```diff
@@ -458,24 +464,24 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
...
-        return cmd
+        # Submit with HPCpy PBSClient
+        # TODO: Is there a way to print the command when dry_run is False?
+        job_or_cmd = client.submit(job_script = f"-- {python_exe} {pbs_script}",
+                      directives = pbs_flags,
+                      dry_run = dry_run,
+                      queue = pbs_config.get('queue', 'normal'),
+                      storage = [storage for storage in storages] if storages else None,
+                      variables = pbs_vars,
+                      render=False, # Save the run script from a template
```
aidanheerdegen:
```suggestion
                      render = False, # Save the run script from a template
```
To keep formatting consistent.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "filelock",
     "questionary",
     "pint",
+    "hpcpy",
```
aidanheerdegen: Do we need a minimum version to support module commands?

test/test_pbs.py
```diff
@@ -362,20 +363,17 @@ def test_find_mounts():
...
 
     # Use new mechanism to return a scheduler
     sched_name = config.get('scheduler', 'pbs')
     sched_type = scheduler_index[sched_name]
     sched = sched_type()
-
-    # Monkey patch pbs_env_init as we don't have a
-    # functioning PBS install in travis
-    payu.schedulers.pbs.pbs_env_init = lambda: True
+    # mock_client.submit.return_value = "mock_pbs_job_id"
```
aidanheerdegen: Will these commented lines be removed?

aidanheerdegen (commented):
Some questions, some suggestions ...

payu/cli.py
Qian-HuiChen: Within this PR:
- Slurm: submit command with subprocess, catch error in `except subprocess.CalledProcessError as e:` ([code](https://github.com/payu-org/payu/pull/802/changes#diff-7c5f02edce42662ef02a3cef6c4022181969d49f9e52d89ff3ab517922b5f796R247)).
- PBS: submit with HPCpy, catch error in `except Exception as e:` ([code](https://github.com/payu-org/payu/pull/802/changes#diff-7c5f02edce42662ef02a3cef6c4022181969d49f9e52d89ff3ab517922b5f796R256))

pyproject.toml
Qian-HuiChen: The module load support in HPCpy is not released yet. However, yes, payu will require a minimum version of HPCpy, likely v0.7.4 or later.

payu/schedulers/pbs.py
Qian-HuiChen: Regarding support for full environment variable export (`qsub -V`), I think it would be useful if HPCpy could accept a `full_variables` argument in `submit()`, and add the `-V` flag to the qsub command when it is enabled. 
On the payu side, payu could accept a `--full-variable` flag from users and pass this signal onto HPCpy.

payu/schedulers/pbs.py
bschroeter: Would it also be reasonable to assume that it should fail if the user is not a member of the project/group? Perhaps it should be checked and fail before this?

payu/schedulers/pbs.py
bschroeter: It is already possible to add a `-V` to the list of directives passed to hpcpy, however, if you need an explicit keyword argument that will do the same thing, then please feel free to raise an issue over on the hpcpy repo.

payu/schedulers/pbs.py
```diff
@@ -458,24 +464,24 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
...
-        # Construct job submission command
-        cmd = 'qsub {flags} -- {python} {script}'.format(
-            flags=' '.join(pbs_flags),
-            python=python_exe,
-            script=pbs_script
-        )
 
-        return cmd
+        # Submit with HPCpy PBSClient
+        # TODO: Is there a way to print the command when dry_run is False?
```
bschroeter: Not yet, although it might come out in the logs.

We could consider tracking the last command (or the command history in general) with a variable local to the client.

Actually, that is a good idea. I will raise an issue for that on hpcpy.

bschroeter: https://github.com/ACCESS-NRI/hpcpy/issues/69

payu/envmod.py
bschroeter: OK, this is what I was concerned about. It looks like you are loading modules programmatically, rather than just doing so in the job script. Do we need to make changes over in hpcpy to support this?

pyproject.toml
bschroeter: v0.8.0 released with modules support. Allow for 10-15mins for it to become available on PyPi/Conda.

bschroeter (commented):
I can't speak to the deep Payu functionality, but it looks OK to me. Some comments and questions throughout.

test/test_pbs.py
Qian-HuiChen: Yes, thanks for catching this. Removed in [692beea](https://github.com/payu-org/payu/pull/802/commits/692beea92242d628424c427f498d4510c4b75875)

payu/schedulers/pbs.py
Qian-HuiChen: Resolved in [692beea](https://github.com/payu-org/payu/pull/802/commits/692beea92242d628424c427f498d4510c4b75875)

payu/schedulers/pbs.py
Qian-HuiChen: Sounds good and clear. Resolved in [692beea](https://github.com/payu-org/payu/pull/802/commits/692beea92242d628424c427f498d4510c4b75875)

payu/envmod.py
Qian-HuiChen: I get what you mean and I agree @aidanheerdegen . Addressed in [692beea](https://github.com/payu-org/payu/pull/802/commits/692beea92242d628424c427f498d4510c4b75875).
Suggested by Aidan: [psutil](https://psutil.readthedocs.io/stable/#:~:text=The%20environment%20variables%20of%20the%20process%20as%20a%20dict.%20Note:%20this%20might%20not%20reflect%20changes%20made%20after%20the%20process%20started.)

Qian-HuiChen: @bschroeter I think it will be cleaner to move the `path_added_by_user_module()` part to HPCpy, and HPCpy can return the `added_path` to payu, because payu will need this later to find the executable. 
I opened an [issue](https://github.com/ACCESS-NRI/hpcpy/issues/70) in HPCpy.

payu/schedulers/pbs.py
Qian-HuiChen: Good to know that I can add `-V` by `directives`! Will do it this way.

pyproject.toml
Qian-HuiChen: Updated pyproject.toml file in [2f6b5ef](https://github.com/payu-org/payu/pull/802/commits/2f6b5eff07742d5f31fb3f51f028f867b3154d00)

payu/schedulers/pbs.py
Qian-HuiChen: Updated in [2f6b5ef](https://github.com/payu-org/payu/pull/802/commits/2f6b5eff07742d5f31fb3f51f028f867b3154d00). If user add `export_env_vars: True` in `config.yaml`, payu will tell HPCpy to add `-V` in `qsub`.

pyproject.toml
Qian-HuiChen: @bschroeter the CI [failed](https://github.com/payu-org/payu/actions/runs/29880582064/job/88800351096?pr=802#:~:text=The%20following%20package,a%20missing%20channel) because there is still no 0.8.0 in anaconda.org (see [website](https://anaconda.org/channels/accessnri/packages/hpcpy/overview)). Could you have a look of this?

payu/envmod.py
aidanheerdegen: > OK, this is what I was concerned about. It looks like you are loading modules programmatically, rather than just doing so in the job script. Do we need to make changes over in hpcpy to support this?

This is loading the modules in a subprocess to determine which paths are added by the module. These paths are then used to ensure the correct executable path is used.

This seems payu specific to me.

payu/schedulers/pbs.py
aidanheerdegen: > Would it also be reasonable to assume that it should fail if the user is not a member of the project/group? Perhaps it should be checked and fail before this?

It would involve duplication of code to check earlier. It seems the right place to do it, when the project code is going to be used. This will fail before job submission if member is not a member of the project.

pyproject.toml
aidanheerdegen: Seems 0.8.0 is still not available in the access-nri anaconda channel.

payu/cli.py
aidanheerdegen: > * PBS: submit with HPCpy, catch error in `except Exception as e:`

So HPCpy will throw an exception?

payu/schedulers/pbs.py
aidanheerdegen: I'm not convinced we should do this at all (we don't know why the TODO was there), and if we do, I think it should be in a separate PR.

pyproject.toml
bschroeter: Looks like a false positive on the publish:

image

I will ping upstream on the publication workflow to see what is going on.

bschroeter: Transient issue resolved, please confirm. https://github.com/ACCESS-NRI/hpcpy/issues/72

payu/cli.py
Qian-HuiChen: I looked into HPCpy, and here's the call chain I found:

* `client.submit()` ([code](https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/client/pbs.py#L188)) calls `submit()` in base.py ([code](https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/client/base.py#L97-L162)).
* `submit()` then calls `shell()` in utilities.py ([code](https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/utilities.py#L42-L52)).
* `shell()` uses `subprocess.run()` to submit the job inside a `try`/`except` block that catches `subprocess.CalledProcessError`.

To test how this bubble up to payu, I added a `raise subprocess.CalledProcessError(...)` inside `shell()`. The error message shows up as expected in `payu run`.

pyproject.toml
Qian-HuiChen: Thanks! I can confirm that the CI workflow runs successfully.

payu/schedulers/pbs.py
aidanheerdegen: Current action is on me to create a separate issue for this.

Qian-HuiChen: Following our discussion, I commented out the `qsub -V` option in [b9efbfb](https://github.com/payu-org/payu/pull/802/commits/b9efbfb2880f9a6bf419224c6b649a83144337ca). This will need to be discussed whether this feature is needed anymore.

payu/schedulers/pbs.py
Qian-HuiChen: Thanks for implementing this! I tested in payu and confirmed that `job_or_cmd.history[0]` returns the submitted payu command. In commit [b9efbfb](https://github.com/payu-org/payu/pull/802/commits/b9efbfb2880f9a6bf419224c6b649a83144337ca), the line is commented out because I want to wait for https://github.com/ACCESS-NRI/hpcpy/issues/71 to be included in a release.

bschroeter: Release hpcpy 0.9.0, stand by for publication.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "filelock",
     "questionary",
     "pint",
+    "hpcpy >=0.8.0",
```
bschroeter: update to 0.9.0

bschroeter (commented):
All good from HPCpy end. Check with @aidanheerdegen for payu stuff.

Qian-HuiChen:
Commit [5bee3d1](https://github.com/payu-org/payu/pull/802/commits/5bee3d1b609ca0d6aa9613d2540612d4dc03ab64) uses HPCpy v0.9.0, which can return the submitted command. Payu now displays as:
```
------ Job submitted ------
Submission command: qsub xxx
Job ID: 1700000.gadi-pbs
```
OR
```
---- Dry run (submission skipped) -----
Submission command: qsub xxx
```
I will need to update model-config-tests ([issue](https://github.com/ACCESS-NRI/model-config-tests/issues/237)) before this PR is merged, because model-config-tests assume the job id is printed as a single line.

payu/cli.py
```diff
@@ -224,16 +224,15 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
 
             if dry_run:
                 # If dry_run is True, print out the submission command and exit
-                print(f"Printing job submission command without executing it:\n"
-                    f"    {job_or_cmd}")
+                print(f"---- Dry run (submission skipped) -----\n"
```
bschroeter, line 226: Just mentioning that it would be more robust to use a logging instance with a streamhandler to point at stdout in future. It is not worth holding up this merge for it though, as it speaks more broadly to how we communicate information back to the user.

payu/cli.py
```diff
@@ -224,16 +224,15 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
...
+                    f"Submission command: {job_or_cmd}")
                 return None
             else:
-                
-                # Uncomment this after HPCpy PR#71 is included in a release
-                # print(f"Submitted job: {job_or_cmd.history[0]}") 
-                # Print the job ID after submission
+                # Print the job ID and command after submission 
+                print(f"------ Job submitted ------\n"
+                    f"Submission command: {job_or_cmd.history[0]}")
```
bschroeter: Just noting that if you get the command over the job instance then this will fail as the `str` object will not have a `history` attribute, nor one that can be indexed.

payu/cli.py
```diff
@@ -224,16 +224,15 @@ def submit_job(script, config, vars=None, expt=None, current_run=None, type=None
...
                 job_id = job_or_cmd.id
-                print(f"Job submitted with ID: {job_id}")
+                print(f"Job ID: {job_id}")
```
bschroeter, line 234: Likewise.

It might be worth either checking that you have a `Job` instance prior to doing these prints, or updating the name of the variable if it is always just a `Job` object.

payu/logger.py
```diff
@@ -3,6 +3,7 @@
 
 # Third Party imports
 from colorama import init, Fore, Style
+import hpcpy
 
 class ColoredFormatter(logging.Formatter):
```
bschroeter, line 8: Fancy, I think I might use this over in HPCpy.

bschroeter (changes requested):
Just a couple on potential attribute issues, if they don't apply, please re-send for review to me as it is otherwise approved from my end.

payu/cli.py
aidanheerdegen: I'm confused about what "command over the job instance" means. Is when the job submission fails?

payu/schedulers/pbs.py
```diff
@@ -341,33 +344,27 @@ def validate_memory_with_queue_limits(cls, pbs_mem: str, queue: str, n_cpus: int
                 f"than the limit of {mem_per_node:.2f}GB per node for queue '{queue}'."
             )
 
-    def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
+    def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None, dry_run=False):
         """Prepare a correct PBS command string"""
 
-        pbs_env_init()
```
jo-basevi, line 347: Safe to remove this function definition too as it's no longer used.

payu/cli.py
bschroeter: When submitting a job with `hpcpy` if you add the `dry_run=True` argument you get a `str` that *would* have been the command executed. It was originally intended as a feature for development.

Submitting with `dry_run=False` will return a `Job` object, which has all of the attributes and methods you would expect.

My comment was just one of completeness, if there were ever a situation that you would trigger the `str` return, then this code will fail.

If it is something that is not reliably going to return a `Job` instance, then we either need to wrap this is in a conditional to return the attributes only if it is an object, or change the way the `submit` method actually works to ensure a `Job` object is always returned, even one that hasn't actually been submitted?

Savvy?

payu/schedulers/pbs.py
```diff
@@ -458,24 +459,24 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
...
+        job_or_cmd = client.submit(job_script = f"-- {python_exe} {pbs_script}",
+                      directives = pbs_flags,
+                      dry_run = dry_run,
+                      queue = pbs_config.get('queue', 'normal'),
+                      storage = [storage for storage in storages] if storages else None,
+                      variables = pbs_vars,
+                      render = False, # Save the run script from a template
+                      module_purge = True,
+                      module_use = module_use_paths,
+                      module = ['pbs'] + (module_names if module_names else []),
```
jo-basevi: I am bit confused as what these lines are doing as the job  submission script seems to have no module commands?

jo-basevi: I assuming the goal is to load modules at the start of a job? If so, what was the reason to load modules at the start of the job - was it to handle different module commands for HPCs, and eventually remove the envmod file?

This loads the PBS and user modules, but not any of the modules in `Experiment.load_modules()` and mpi modules loaded with `envmod.lib_update` (which needs the model executables).

jo-basevi (commented):
Thanks @Qian-HuiChen and @bschroeter for adding this in! 

Just added a couple of comments I have so far.

When running a ACCESS-OM2 configuration (dev-1deg_jra55_iaf), with payu setup on the login node, I get

```
[ERROR] Executable for ocean model not found on path: /scratch/tm70/<user>/access-om2/bin/mom5_access_om
```

Which indicates it wasn't found in the module paths, as payu checks the laboratory `/bin/` directory last.

payu/cli.py
aidanheerdegen: Thanks. I was confused because there is a logic gate that would keep that from happening *for the code as written*. Clearly if the code is modified such that this could occur, then yes it would be bad.

How about this: remove the `except` (it isn't needed) and replace with `try/except` block, and provide a user friendly message if this is any problem accessing the necessary data?

aidanheerdegen: Sorry, didn't notice it was already in a `try/except` outer block. The `return` makes the `else` redundant, so just remove the `else` and remove indent for the rest of the code.

aidanheerdegen: > If it is something that is not reliably going to return a `Job` instance, then we either need to wrap this is in a conditional to return the attributes only if it is an object, or change the way the `submit` method actually works to ensure a `Job` object is always returned, even one that hasn't actually been submitted?

It isn't obvious to me why this might happen. Unless we have good reason to program defensively for this I think it is fine to not add more checks at this point. Especially as it is all wrapped in a `try/except`.

payu/cli.py
Qian-HuiChen: I update the variable name to `cmd` for dry-run, and `job` for normal runs in the latest [commit](https://github.com/payu-org/payu/pull/802/changes/4d87546386b88adce516f5877973104c9507644b#diff-7c5f02edce42662ef02a3cef6c4022181969d49f9e52d89ff3ab517922b5f796R229-R234)

payu/cli.py
Qian-HuiChen: I get rid of the `else` part to make the code more readable in the latest [commit](https://github.com/payu-org/payu/pull/802/changes/4d87546386b88adce516f5877973104c9507644b#diff-7c5f02edce42662ef02a3cef6c4022181969d49f9e52d89ff3ab517922b5f796R229-R234).

payu/schedulers/pbs.py
Qian-HuiChen: Good catch! I removed this function in [4d87546](https://github.com/payu-org/payu/pull/802/commits/4d87546386b88adce516f5877973104c9507644b)

Qian-HuiChen:
Following the discussion with @aidanheerdegen and @jo-basevi, this PR will be limited to replacing PBS job submission with HPCpy ([d72b7fc](https://github.com/payu-org/payu/pull/802/commits/d72b7fc157654c6f4c6f8a019f1106e5b2ccd0e3) revert the module handling changes). 
Module handling will be addressed in a future pull request. The Jinja template for the HPCpy submission script deserves a separate discussion, particularly around where it should be stored and what it should include.

payu/schedulers/pbs.py
```diff
@@ -458,24 +459,20 @@ def submit(self, pbs_script, pbs_config, pbs_vars=None, python_exe=None):
...
-        )
 
-        return cmd
+        # Submit with HPCpy PBSClient
+        # TODO: Is there a way to print the command when dry_run is False?
+        job_or_cmd = client.submit(job_script = f"-- {python_exe} {pbs_script}",
+                      directives = pbs_flags,
+                      dry_run = dry_run,
+                      queue = pbs_config.get('queue', 'normal'),
+                      storage = [storage for storage in storages] if storages else None,
```
aidanheerdegen:
```suggestion
                      storage = list(storages),
```
If the list comprehension is just to convert a `set` to a `list` then above will do it more concisely.

payu/cli.py
Qian-HuiChen: model-config-tests reads the job id and submitted command from [stdout]([url](https://github.com/ACCESS-NRI/model-config-tests/blob/fd001355feed472849ddd6e6df3a19c7f5c90f0d/src/model_config_tests/exp_test_helper.py#L259)). If I change this to `logger.info`, the job ID will instead be written to `stderr`.
I think we should leave this as it is for now. I'll update model-config-tests to use `payu status --json` instead, so it no longer relies on parsing `stdout` or `stderr`. Once that's done, I can switch this over to logging.

payu/schedulers/pbs.py
Qian-HuiChen: Thanks for the cleaner code. Updated in [691472c](https://github.com/payu-org/payu/pull/802/commits/691472ca7150fc69c54af6730f815cefe1d3a852)

payu/cli.py
bschroeter: Agreed

aidanheerdegen (dismissed):

jo-basevi (dismissed):
Thanks, looks good to me!

jo-basevi (approved):
LGTM

PR #803, opened 2026-07-14
Update relative restart path to absolute path.

### LINKED ISSUE #762: Relative paths in `restart:` field
I found this in ESM1.6, but suspect it's a general issue.

I used a relative path in the `restart:` field of config.yaml

`restart: histo_r_from_949`

payu creates manifests fine, without error:

i.e.:

```
...
work/ocean/INPUT/ocean_velocity.res.nc:
  fullpath: histo_r_from_949/ocean/ocean_velocity.res.nc
  hashes:
    binhash: fb16b086adb212eb6ad8192969df9a58
    md5: f64ab76be6a3ed1b3a879b61944c64c4
...
```
that is the correct _md5_ 

But then the files don't get symlinked correctly. The symlink points to a non existant path (because it's a relative path)

```
$ ll work/ocean/INPUT/ocean_velocity.res.nc 
lrwxrwxrwx 1 as2285 tm70 44 Jun 10 10:42 work/ocean/INPUT/ocean_velocity.res.nc -> histo_r_from_949/ocean/ocean_velocity.res.nc
```

Image

(The model then doesn't run)

I don't know what the correct behaviour should be.

Maybe `payu` shouldn't allow this, or maybe is should resolve an absolute path ?

Qian-HuiChen:
Previously, `payu setup` and `payu run` would fail when a relative `restart:` path was configured in `config.yaml`.

This PR resolves relative restart paths to absolute paths, assuming that the restart paths is relative to the control directory. 

Closes #762

### DIFF
--- a/payu/models/model.py
+++ b/payu/models/model.py
@@ -155,7 +155,7 @@ class Model(object):
         if self.prior_restart_path and not os.path.isabs(self.prior_restart_path):
             # Update to absolute path, assuming it is relative to control path
             rel_path = os.path.join(self.expt.control_path, self.prior_restart_path)
-            self.prior_restart_path = os.path.realpath(rel_path)
+            self.prior_restart_path = os.path.realpath(rel_path, strict=True)
             print(f"Prior restart path is set as a relative path, resolved to {self.prior_restart_path}")
 
         if len(self.expt.models) > 1:
--- a/payu/models/model.py
+++ b/payu/models/model.py
@@ -151,6 +151,13 @@ class Model(object):
         self.prior_output_path = self.expt.prior_output_path
         self.prior_restart_path = self.expt.prior_restart_path
 
+        # If prior restart path exists and is a relative path
+        if self.prior_restart_path and not os.path.isabs(self.prior_restart_path):
+            # Update to absolute path, assuming it is relative to control path
+            rel_path = os.path.join(self.expt.control_path, self.prior_restart_path)
+            self.prior_restart_path = os.path.realpath(rel_path)
+            print(f"Prior restart path is set as a relative path, resolved to {self.prior_restart_path}")
+
         if len(self.expt.models) > 1:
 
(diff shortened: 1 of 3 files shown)


payu/models/model.py
```diff
@@ -151,6 +151,13 @@ def set_model_output_paths(self):
         self.prior_output_path = self.expt.prior_output_path
         self.prior_restart_path = self.expt.prior_restart_path
 
+        # If prior restart path exists and is a relative path
+        if self.prior_restart_path and not os.path.isabs(self.prior_restart_path):
+            # Update to absolute path, assuming it is relative to control path
+            rel_path = os.path.join(self.expt.control_path, self.prior_restart_path)
+            self.prior_restart_path = os.path.realpath(rel_path)
```
aidanheerdegen:
```suggestion
            self.prior_restart_path = os.path.realpath(rel_path, strict=True)
```
My reading of [the docs](https://docs.python.org/3/library/os.path.html#os.path.realpath) suggests this is required to make sure it errors out if the relative path doesn't exist.

aidanheerdegen (commented):
Looks good, one small suggestion/comment.

payu/models/model.py
Qian-HuiChen: `experiment.py` checks whether user-specified restart path exists before the model sets the restart path (see [code](https://github.com/payu-org/payu/blob/96f35f710c44fbe5be102fca736e930c2dcfb4cd/payu/experiment.py#L409-L413)). In this case, payu will raise an error before reaching `self.prior_restart_path = os.path.realpath(rel_path, strict=True)`. I added the additional parameter in place 
([commit](https://github.com/payu-org/payu/pull/803/commits/831385cee6883f7cd07cd495ef8e2cd84373c190)), since it also catches other `OSError` exceptions.

aidanheerdegen (approved):

PR #806, opened 2026-07-20
Update softfilelock corresponding to filelock 3.30.0

Qian-HuiChen:
A recent release of the `filelock` package changed the behavior of file lock expiration. From `filelock`[ 3.30.0](https://github.com/tox-dev/filelock/releases/tag/3.30.0), `FileLock` ignores the `lifetime` parameter; lock expiration now only applies to `SoftFileLock`.

This PR updates payu job file to use `SoftFileLock` instead of `FileLock`, making sure that the lock expires after `LOCK_LIFETIME = 8`.

### DIFF
--- a/payu/telemetry.py
+++ b/payu/telemetry.py
@@ -13,7 +13,7 @@ import requests
 import threading
 from typing import Any, Optional
 import warnings
-from filelock import FileLock, Timeout
+from filelock import SoftFileLock, Timeout
 
 import cftime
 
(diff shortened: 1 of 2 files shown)


jo-basevi (approved):
LGTM. The CI tests seem to be stuck in queue and Github Status shows there's some existing issues with Github Actions (https://www.githubstatus.com/) so I wonder if that is causing the issue

Qian-HuiChen:
Thanks! I will wait for an hour to see if Github will solve the issue.

Issue #521, opened 2024-09-25
Building intake-esm datastores


anton-seaice:
Far from an original thought, but no one has made an issue yet :)

For current model shared output, we are making some of these accessible to users through the ACCESS-NRI Intake Catalog. This is a catalog of Intake-ESM datastores, where mostly one model run is one datastore.

For new model runs, providing a uniform interface for interacting with the output data through Intake-ESM datastores, improves the sharability and portability of analysis. And means that analysis workflows are ready for sharing in working groups.

To generate these datastores, a builder is required. These builders:
```
 'AccessCm2Builder',
 'AccessEsm15Builder',
 'AccessOm2Builder',
 'AccessOm3Builder'
```
are defined (currently) in https://github.com/accESS-NRI/access-nri-intake-catalog. There should be a builder for MOM6/SIS2 (i.e. GFDL-OM4 / COSIMA Panantarctic) soon too. These builders are parallelised but don't use dask. There's no capability to modify existing datastores. 

For ACCESS-OM3, we've implemented this as [an archive userscript](https://github.com/COSIMA/om3-scripts/blob/main/payu_config/archive_scripts/build_intake_ds.py). I believe this uses all the cores available within the run job. At the current resolutions and short run times this works ok. It would probably be better to qsub this into a different job, and make an intentional decision about the number of resources needed.  

Currently, most users load the intake catalog through hh5 - which is not available to the CI worker for model configs. To implement in payu, we would need to add "access-nri-intake-catalog" as a dependency in `pyproject.toml`. And then I suggest we try and add as a general step during payu archive for the models with builders.

jo-basevi:
As it is used in a user-script (rather than payu project code), it might make more sense to add `access-nri-intake-catalog` directly to the payu conda environment? Similarly to what was done for post-processing dependencies `mule` and `um2nc` for ESM1.5 configs https://github.com/ACCESS-NRI/payu-condaenv/blob/main/env.yml

anton-seaice:
I'm suggesting it gets used in payu project code ... instead of a user script its added to the payu archive or payu sync step in payu code.

aidanheerdegen:
> For new model runs, providing a uniform interface for interacting with the output data through Intake-ESM datastores, improves the sharability and portability of analysis. And means that analysis workflows are ready for sharing in working groups.

Yeah this is a great idea, and definitely something we want to support. 

[...]

PR #808, opened 2026-07-22
Make intake datastores for archived/synced output

anton-seaice:
A very tentative start point to get payu to make an intake datastore.

Contributes to #521 

needs to be run in a PBS job after syncing has completed. probably one node on the normalsr queue is a good start point.

### DIFF
--- a/payu/datastore.py
+++ b/payu/datastore.py
@@ -1,0 +1,67 @@
+"""Experiment post-processing - syncing archive to a remote directory
+
+:copyright: Copyright 2011 Marshall Ward, see AUTHORS for details.
+:license: Apache License, Version 2.0, see LICENSE for details.
+"""
+
+# Standard
+import warnings
+
+# Local
+import payu.errors as errors 
+from payu.status import collect_expt_paths
+
+
+class MakeIntakeDatastore():
+    """Class used for generating intake datastores"""
+
+    def __init__(self, expt):
+        self.expt = expt
+        # self.config = self.expt.config.get('datasore', {})
+
+        self.source_paths = []
+
+
+    def run(self):
+        """  
+        Generate an intake datastore in the sync location
+        """
+
+        try:
+            import access_nri_intake.source.builders as builders
+            from access_nri_intake.experiment import use_datastore
+        except ImportError:
+            warnings.warn(
+                "ACCESS-NR intake package not found, "
+                "skipping datastore generation"
+            )
+            return
+
+        builder_map = {
+            'access': builders.AccessEsm15Builder,
+            'access-esm1.6': builders.AccessEsm16Builder,
+            'access-om2': builders.AccessOm2Builder,
+            'access-om3': builders.AccessOm3Builder,
+            'mom': builders.Mom6Builder
+        }
+
+        if not self.expt.model_name in builder_map.keys():
+            warnings.warn(
+                f"No intake datastore builder found for {self.expt.model_name}, "
+                " skipping datastore generation"
+            )
+
+            return
+
+        expt_paths = collect_expt_paths(self.expt)
+
+        description = (
+            f"intake-esm datastore for experiment {expt_paths['experiment_name']} "
+            f"({expt_paths['experiment_uuid']})"
+        )
+
+        use_datastore(
+            experiment_dir=expt_paths["sync_path"],
+            description=description, 
+            builder=builders[self.expt.model_name]
+        )


aidanheerdegen:
Thanks for kicking this off @anton-seaice.

I think it makes sense to create a new subcommand, `payu index` (? Happy for other ideas) so it can be run when required, or in other contexts, e.g. to create an index in the `laboratory` on `/scratch`.

In this way the same pattern could used as is done for `sync`

https://github.com/payu-org/payu/blob/master/payu/experiment.py#L1006-L1021

though it would/should be contained within the `if syncing` logic.

Would need to exclude the intake datastore files from syncing if there are any in the laboratory, as they would overwrite any intake catalogue in the `sync` location.

Qian-HuiChen:
To exclude the intake datastore files from syncing, one way is to modify the exclude flags in [sync.py](https://github.com/payu-org/payu/blob/ea566c37f8dbe96ad5fc74ea4b1a848f740750de/payu/sync.py#L218-L234).

anton-seaice:
> I think it makes sense to create a new subcommand, `payu index` (? Happy for other ideas) so it can be run when required, or in other contexts, e.g. to create an index in the `laboratory` on `/scratch`.

Would the path be an argument? how does payu know whether to create it in the laboratory location or the sync location ?

aidanheerdegen:
> Would the path be an argument? 

I can imagine it working like `payu collate`, where the path is optional, and if specified that is used, otherwise use the default. 

Default would be to generate in the sync location if `sync: True`. Actually it could just call `payu sync` in that case to make sure the most recent files are available.

If `sync: False` the default would be to create it in the laboratory.

> how does payu know whether to create it in the laboratory location or the sync location ?

See above.

anton-seaice:
Ok I set Claude on this, i think it's doing all the requested features:

1. If payu sync is enabled, then build it in the sync location, and delete it from the archive location if it exists. See example in `/g/data/tm70/as2285/payu-dev/pr808-tests/3-catalog-with-sync`

2. If sync is not enabled, then build in the archive location: `/g/data/tm70/as2285/payu-dev/pr808-tests/2-catalog-no-sync`

3. If not enabled, do nothing: `/g/data/tm70/as2285/payu-dev/pr808-tests/1-baseline`

anton-seaice:
This is ready for review @aidanheerdegen @charles-turner-1

anton-seaice:
Adding 

```
catalog:
  enable: true
```

to config.yaml is enough, but PBS resources can be set too

payu/datastore.py
```diff
@@ -0,0 +1,96 @@
...
+            )
+            return
+
+        builder_map = {
+            'access': builders.AccessEsm15Builder,
+            'access-esm1.6': builders.AccessEsm16Builder,
+            'access-om2': builders.AccessOm2Builder,
+            'access-om3': builders.AccessOm3Builder,
+            'mom': builders.Mom6Builder
+        }
```
charles-turner-1: There are other builders in this module - is there any reason not to include them all?

payu/datastore.py
```diff
@@ -0,0 +1,96 @@
...
+            f"intake-esm datastore for experiment "
+            f"{expt_paths['experiment_name']} "
+            f"({expt_paths['experiment_uuid']})"
+        )
+
+        builder_kwargs = {}
+        if issubclass(builder, builders.AccessEsm15Builder):
+            # AccessEsm15Builder/AccessEsm16Builder require an explicit
+            # `ensemble` argument
+            builder_kwargs['ensemble'] = False
```
charles-turner-1: This might not be true as of v1.6.6, I'll double check that

charles-turner-1 (commented):
Couple of minor comments from me - I can't really comment super accurately on the payu specifics though. Otherwise looks good!

payu/datastore.py
```diff
@@ -0,0 +1,96 @@
...
+        builder_map = {
+            'access': builders.AccessEsm15Builder,
+            'access-esm1.6': builders.AccessEsm16Builder,
+            'access-om2': builders.AccessOm2Builder,
+            'access-om3': builders.AccessOm3Builder,
+            'mom': builders.Mom6Builder
+        }
+
+        builder = builder_map.get(self.expt.model_name)
+        if builder is None:
```
anton-seaice: I am wondering if it should fail earlier if this model isn't supported

payu/datastore.py
anton-seaice: The other builders are not for models run through payu, or for observational data.

(With the possible exception of ROMS ?)

anton-seaice: i.e. should ROMS be added ?

charles-turner-1: Ahh, of course. not really qualified to comment on ROMS, don't know who would be?

aidanheerdegen: > i.e. should ROMS be added ?

To `payu` or intake?

payu/datastore.py
aidanheerdegen: Could make `catalog` a method of the model and leave it non-implemented for non-supported builderss

aidanheerdegen (commented):
Also note that the job dependency logic is being refactored in #832. Once that is merged this PR will need to be rebased and modified in a similar way.

payu/datastore.py
charles-turner-1: Payu, ROMS is already in intake

anton-seaice: ROMS is in both payu and intake, should it be n the list is my question ? (e.g. is the format for ROMS output standardised enough that the intake builder works for most ROMS output if run through payu)

payu/datastore.py
anton-seaice: This suggestion feels a bit confusing to me, as mostly the builder is only for a top-level model, where lots of models are implemented as sub-models to other models ?

aidanheerdegen: True. We do have some other places where we test logic like "is this a submodel". Not saying it's the best idea, just that it is a possibility.

payu/datastore.py
aidanheerdegen: No idea.

Qian-HuiChen:
Thanks @anton-seaice for the draft! It looks very nice. If you don't mind, I will take over this PR from tomorrow and keep working on this

Qian-HuiChen:
I get this warning since I added `access-nri-intake` to pyproject.toml:
```
<frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
```
Seems to be the conflict between netCDF4 and numpy.

I am not sure if I should put `builders.Mom6Builder` in mom6.py driver. I run `payu catalog` in mom6 configuration and it failed.

aidanheerdegen:
> I get this warning since I added `access-nri-intake` to pyproject.toml:
> 
> ```
> <frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
> ```
> 
> Seems to be the conflict between netCDF4 and numpy.

This is why we said we should have agreed organisational standards for the versions of python packages we support. To avoid problems like this. 

@charles-turner-1 may have some useful input into the exact reason for the conflict and how it can be resolved.
 
> I am not sure if I should put `builders.Mom6Builder` in mom6.py driver. I run `payu catalog` in mom6 configuration and it failed.

Without more detail on the nature of the error I can't speculate why, or if this is a specific problem with making the intake method model based.

charles-turner-1:
> > I get this warning since I added `access-nri-intake` to pyproject.toml:
> > ```
> > <frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Seems to be the conflict between netCDF4 and numpy.
> 
> This is why we said we should have agreed organisational standards for the versions of python packages we support. To avoid problems like this.
> 
> @charles-turner-1 may have some useful input into the exact reason for the conflict and how it can be resolved.
> 

Unfortunately this is one I'm yet to figure out. Where are you inferring that it's related to netCDF4 from @Qian-HuiChen?

Qian-HuiChen:
The MOM6 error is raised from this [validate_parser function](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/8350a0251c37644299b76de62257a27e12c70174/src/access_nri_intake/source/builders.py#L227-L246):
```pytb
[ERROR] Parser returns no valid assets.
            Try parsing a single file with Builder.parser(file)
            Last failed asset: /scratch/tm70/qc4677/mom6/archive/mom6-perturb-5729b928/output063/Depth_list.nc
            Asset parser return: {'INVALID_ASSET': '/scratch/tm70/qc4677/mom6/archive/mom6-perturb-5729b928/output063/Depth_list.nc', 'TRACEBACK': 'Traceback (most recent call last):\n  File "/scratch/tm70/qc4677/payu-training-venv/lib64/python3.11/site-packages/access_nri_intake/source/builders.py", line 223, in _parser_catch_invalid\n    return cls.parser(file, **kwargs)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/scratch/tm70/qc4677/payu-training-venv/lib64/python3.11/site-packages/access_nri_intake/source/builders.py", line 631, in parser\n    raise ParserError(f"Cannot determine realm for file {file}")\naccess_nri_intake.source.builders.ParserError: Cannot determine realm for file /scratch/tm70/qc4677/mom6/archive/mom6-perturb-5729b928/output063/Depth_list.nc\n'}
```

Qian-HuiChen:
>Where are you inferring that it's related to netCDF4

I googled this warning message and found a similar issue in this [repo](https://github.com/Unidata/netcdf4-python/issues/1354). There was a PR closed this issue. However,even though I made sure my netCDF4 is 1.7.4, this warning still comes up.

I also send this to AI and find out I can trigger this warning message by:
```
$ python -c "import numpy;import warnings;warnings.filterwarnings('error');import netCDF4"
  File "<string>", line 1, in <module>
  File "/scratch/tm70/qc4677/payu-training-venv/lib64/python3.11/site-packages/netCDF4/__init__.py", line 11, in <module>
    from ._netCDF4 import *
  File "src/netCDF4/_netCDF4.pyx", line 1, in init netCDF4._netCDF4
RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject

```

charles-turner-1:
On the error:
`access_nri_intake.source.builders.ParserError: Cannot determine realm for file /scratch/tm70/qc4677/mom6/archive/mom6-perturb-5729b928/output063/Depth_list.nc`

The Mom6Builder currently assumes it can obtain the realm from the filename:
```
    @classmethod
    def parser(cls, file) -> dict:
        output_nc_info = cls.parse_ncfile(file)
        ncinfo_dict = output_nc_info.to_dict()

        if "ocean" in ncinfo_dict["filename"]:
            realm = "ocean"
        elif "ice" in ncinfo_dict["filename"] or "roms" in ncinfo_dict["filename"]:
            realm = "seaIce"
        else:
            raise ParserError(f"Cannot determine realm for file {file}")
        ncinfo_dict["realm"] = realm

        ncinfo_dict["file_id"] = ".".join(
            [
                str(ncinfo_dict["realm"]),
                str(ncinfo_dict["frequency"]),
                str(ncinfo_dict["file_id"]),
            ]
        )

        return ncinfo_dict
```
See https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/8350a0251c37644299b76de62257a27e12c70174/src/access_nri_intake/source/builders.py#L622 for context.


Is this something we're able to / should relax? It should be straightforward if we have the realm in the metadata or something now. 


Let's not mess with netCDF4 for now. I did some digging into this at the start of 2025. Basically, that file (https://github.com/Unidata/netcdf4-python/blob/master/src/netCDF4/_netCDF4.pyx) is a gigantic Cython file with a bunch of nogil statements, cdefs, etc.

We were getting piles of segmentation faults surrounding that area at the time. They've mostly stopped now, and I don't know why. I'd rather not poke the bear.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "questionary",
     "pint",
     "hpcpy >=0.9.0",
+    "access-nri-intake",
```
anton-seaice: for what it's worth, there was talk of making the dependency on access-nri-intake optional by including it in the payu environment but not here

Qian-HuiChen: Thanks! I used the "try import" logic in your code and move "access-nri-intake" to optional dependency. Phew the noisy warning message is gone (see [e1b884f](https://github.com/payu-org/payu/pull/808/commits/e1b884f21276e495b577d4ff517ad441b44dc9f8))

Qian-HuiChen:
The noisy warning message disappeared after I change the "access-nri-intake" as an optional dependency. Now we only have the MOM6 realm not found problem.

aidanheerdegen (commented):
I believe the repetition of code could be avoided by making a class that contains this code and use it as a mixin for the models that support intake catalogue generation.

See https://github.com/payu-org/payu/blob/master/payu/models/mom_mixin.py and it being used https://github.com/payu-org/payu/blob/master/payu/models/mom6.py#L82

The build name could be saved as a top level class variable and passed to the generic mixin function as a string, and the [relevant builder function returned from a `getattr` call](https://github.com/ACCESS-NRI/access-nri-intake-catalog/blob/8350a0251c37644299b76de62257a27e12c70174/tests/test_builders.py#L444).

payu/subcommands/status_cmd.py
```diff
@@ -33,15 +33,16 @@ def runcmd(lab_path, config_path, json_output,
...
+            # Determine archive path
+            lab = Laboratory(config_path=config_path, lab_path=lab_path)
+            warnings.filterwarnings("error", category=MetadataWarning)
+            
+            expt = Experiment(lab, config_path=config_path)
 
-        archive_path = Path(expt.archive_path)
-        control_path = Path(expt.control_path)
+            archive_path = Path(expt.archive_path)
+            control_path = Path(expt.control_path)
```
charles-turner-1:
```suggestion
    with open(os.devnull, 'w') as devnull, redirect_stdout(devnull):
        # Determine archive path
        lab = Laboratory(config_path=config_path, lab_path=lab_path)
        warnings.filterwarnings("error", category=MetadataWarning)
            
        expt = Experiment(lab, config_path=config_path)

        archive_path = Path(expt.archive_path)
        control_path = Path(expt.control_path)
```

Looks a tad cleaner to me?

test/test_datastore.py
```diff
@@ -0,0 +1,170 @@
...
+    except Exception as e:
+        print(e)
+
+    yield
+
+    # Remove tmp directory
+    try:
+        shutil.rmtree(tmpdir)
+    except Exception as e:
+        print(e)
```
charles-turner-1, line 38: Excuse my ignorance here, but why doesn't this remove the lab and ctrl dirs?

charles-turner-1 (commented):
One minor comment & a question. 

Generally agree with Aidans comment about DRY, but I think those `make_intake_datastore` functions could just be pure functions from what I can see? Doesn't look like any of the `make_intake_datastore` code relies on `self` at all?

Would be simpler than a mixin but there are probably reasons/conventions I'm unaware of

test/test_datastore.py
Qian-HuiChen: In test/common.py:
```
ctrldir = tmpdir / ctrldir_basename
labdir = tmpdir / 'lab'
```
When I wrote this, I was thinking removing `tmpdir` would also remove the `ctrldir` and `labdir`.

payu/datastore.py
Qian-HuiChen: Any update about this 👀 
When I read the builder.py, it seems like no error is raised even if `ensemble` is `True`

payu/subcommands/status_cmd.py
Qian-HuiChen: Thanks! It does look cleaner. Updated in [dd4cd68](https://github.com/payu-org/payu/pull/808/commits/dd4cd686ccfa85444a8806e56ecc4f3a6e202d29)

Qian-HuiChen:
Thanks @aidanheerdegen  for the suggestions!
In [da16c00](https://github.com/payu-org/payu/pull/808/commits/da16c00dd2072ba207b79017a5f2ddd99b32b894), I created a IntakeMixin class that includes the `make_intake_datastore` function. Then I added this class to all supported models, e.g., 
```
class Mom6(IntakeMixin, MomMixin, Fms):
class AccessEsm1p6(IntakeMixin, Model):
```
For unsupported models, they still falls back to `Model` class and report a `NotImplementedError`.

payu/models/model.py
```diff
@@ -486,4 +486,46 @@ def get_cur_expt_time_using_submodel(self, model_types):
...
+
+        # Dynamically get the builder using getattr()
+        builder = getattr(builders, self.intake_builder)
+
+        description = f"Intake-ESM datastores for experiment {expt_name} ({expt_uuid})"
+
+        use_datastore(
+                    experiment_dir=datastore_path,
+                    description=description,
+                    builder=builder,
```
aidanheerdegen: It isn't necessary to pass in the `builder` type explicitly, this mixin class can access it via `self.intake_builder`.

I guess the `experiment_dir` and experiment UUID could also be access through the model class, at least as default values that could be overridden if that was desirable.

Qian-HuiChen: >It isn't necessary to pass in the builder type explicitly, this mixin class can access it via self.intake_builder.

I’m not sure what the first suggestion mean.
For the second suggestion, I can update the function as follows. 
```
def make_intake_datastore(self):
        """Generate an intake-esm datastore for the experiment output.
        """
        try:
            from access_nri_intake.source import builders as builders
            from access_nri_intake.experiment import use_datastore
        except ImportError:
            warnings.warn("access_nri_intake not found, skip datastore generation.")
            return

        # Dynamically get the builder using getattr()
        builder = getattr(builders, self.intake_builder)

        description = f"Intake-ESM datastores for experiment {self.expt.name} ({self.expt.metadata.uuid})"

        use_datastore(
                    experiment_dir=self.expt.datastore_path,
                    description=description,
                    builder=builder,
                    builder_kwargs=self.builder_kwargs if self.builder_kwargs is not None else {},
                )
```

aidanheerdegen: I'm not quite sure what I meant by the first point, as you're already using `self.intake_builder`. Perhaps I didn't read carefully enough. So sorry for the confusion.

Yes that is the sort of thing I was thinking. If we have the state variables already in `self` may as well use them (despite me saying it was bad just constantly dipping into `.expt` for information).

Qian-HuiChen: Thanks for explaining. It does look cleaner than pass it in the arguments. Please see [e8e0dbd](https://github.com/payu-org/payu/pull/808/commits/e8e0dbdc8c786c93ac6d9b747e60afe3e4428e53) for the changes

payu/subcommands/catalog_cmd.py
```diff
@@ -0,0 +1,97 @@
...
+# Local
+from payu import cli
+from payu.experiment import Experiment
+from payu.laboratory import Laboratory
+import payu.subcommands.args as args
+from payu import fsops
+import payu.errors as errors
+
+title = 'catalog'
+parameters = {'description': 'Generate an intake-esm datastore for the '
```
aidanheerdegen, line 16:
```suggestion
parameters = {'description': 'Generate an intake-esm datastore catalog for the '
```
As the command is called `catalog` it makes sense to use the name in the description to make that association more clear.

payu/subcommands/catalog_cmd.py
```diff
@@ -0,0 +1,97 @@
...
+    pbs_vars = cli.set_env_vars(init_run=init_run,
+                                lab_path=lab_path,
+                                dir_path=dir_path)
+
+    catalog_config = pbs_config.get('catalog', {})
+
+    default_ncpus = 104
+    default_queue = 'normalsr'
+    default_mem = '500GB'
+    default_walltime = '01:00:00'
```
aidanheerdegen, line 51: WOAH! This is a lot of resources to make an intake catalog.  Is this correct?

payu/experiment.py
```diff
@@ -1018,6 +1018,28 @@ def profile(self):
...
+        enabled, otherwise it is built in the archive directory."""
+        self.datastore_path = self.archive_path
+
+        # If syncing is enabled, use the sync destination path to build the datastore
+        if self.config.get('sync', {}).get('enable', False):
+            sync_path = self.get_sync_destination()
+
+            if sync_path is not None:
+                self.datastore_path = sync_path
+                remove_datastore(self.archive_path)
```
aidanheerdegen, line 1033: Can you include a comment as to why it is necessary to remove the catalog if it is being written to a remote path?

payu/experiment.py
```diff
@@ -1281,3 +1314,14 @@ def enable_core_dump():
...
+def remove_datastore(directory):
+        """Remove any existing datastore files (json, csv and hash) from
+        `directory`, if present."""
+        directory = Path(directory)
+        # Also matches the "<name>_invalid_assets_<timestamp>.csv" file
+        # use_datastore() writes if some assets fail to parse.
+        patterns = [f"{DATASTORE_NAME}*", f".{DATASTORE_NAME}*"]
+        for pattern in patterns:
+            for path in directory.glob(pattern):
+                path.unlink()
```
aidanheerdegen, line 1327: This will only work if the sync path is a directory on the same filesystem. See above.

payu/experiment.py
```diff
@@ -1018,6 +1018,28 @@ def profile(self):
...
 
+    def make_datastore(self):
+        """Build an intake-esm datastore for the experiment output.
+        The datastore is built in the sync destination if syncing is
+        enabled, otherwise it is built in the archive directory."""
+        self.datastore_path = self.archive_path
+
+        # If syncing is enabled, use the sync destination path to build the datastore
+        if self.config.get('sync', {}).get('enable', False):
+            sync_path = self.get_sync_destination()
```
aidanheerdegen, line 1029: This will only work if the sync path is a directory on the same filesystem. `rsync` urls are supported, but they wouldn't allow deletion of files, or indeed generation of a catalog at all, so should be explicitly checked for.

test/test_sync.py
```diff
@@ -295,6 +296,15 @@ def test_check_uuid_value_error(monkeypatch):
     with pytest.raises(errors.PayuRuntimeError, match="Mismatched experiment UUIDs in sync destination."):
         sync.set_destination_path()
 
+# A datastore in the archive directory is always excluded from sync - it's
+# stale/irrelevant once syncing is enabled, since the datastore is instead
+# (re)built at the sync destination.
```
aidanheerdegen, line 301: The catalogue would also be incorrect if sync'ed as it doesn't refer to the data in the sync location if it is generated in the archive.

aidanheerdegen (commented):
Some questions, some suggestions.

payu/subcommands/catalog_cmd.py
Qian-HuiChen: Added in the latest commit

payu/subcommands/catalog_cmd.py
Qian-HuiChen: I test it with OM2 and give payu-catalog 2 cpus. It took 27 GB and 3min walltime. I updated the default resource to 2cpu, 50GB, 15min in the latest commit

payu/experiment.py
Qian-HuiChen: Sure, I added:
```
# Remove any existing datastore in the archive to avoid duplication/confusion
    remove_datastore(self.archive_path)
```

payu/experiment.py
Qian-HuiChen: I am not quite sure about this comment. This remove_datastore is only applied to archive_path. Can I ask for more details about this comment?

payu/experiment.py
Qian-HuiChen: Thanks for pointing this out! I totally missed the case of rsync. 

In the latest `get_sync_destination()`, it returns None for `rsync` and unconfigured sync path, and datastore will be created in archive.

test/test_sync.py
Qian-HuiChen: I extend the comment as
```
# Always exclude the datastore from sync. The archive copy is stale or
# irrelevant once syncing is enabled because the datastore is (re)built
# at the sync destination.
# Syncing the catalogue from the archive would also be incorrect because
# it references data in the archive location, not the sync destination.
```
Is this what this comment mean?

Qian-HuiChen:
@charles-turner-1 this PR is ready for review. It will be great if you can help me confirm 
- argument for ESM1.5 and 1.6, if `{ensemble: False} is still needed
- should realm file mandatory for MOM6Builder
- `payu catalog` work as expected from a model developer's perspective, especially for the default resource request (2 cpus, 15 min walltime, 50GB)

charles-turner-1:
> @charles-turner-1 this PR is ready for review. It will be great if you can help me confirm 
> 
> - argument for ESM1.5 and 1.6, if `{ensemble: False} is still needed

It is currently, but we can change that - I think it's just that a default was never set.


> - should realm file mandatory for MOM6Builder

I don't think it should be mandatory, no. It's inferred from filenames, which is maybe not the smartest way of doing things though.

> - `payu catalog` work as expected from a model developer's perspective, especially for the default resource request (2 cpus, 15 min walltime, 50GB)

Yeah, that should be fine. Typically takes a minute or two tops.

test/test_sync.py
aidanheerdegen: Yes, thanks!

payu/experiment.py
aidanheerdegen: Oh right. My mistake then. Sorry for the misdirection.

payu/subcommands/catalog_cmd.py
aidanheerdegen: Does the intake process parallelise? I ask because 50GB is equivalent to about 10 Sapphire Rapids CPUs.

Though the cost is the same, the normal queue (Cascade Lake architecture) has many more nodes, so likely a shorter queue time. Though 4GB RAM per cpu, so the effective charge for the same memory will be a slightly higher.

test/test_sync.py
aidanheerdegen, line 1: There doesn't seem to be a test for the changed behaviour in `get_sync_destination`?

aidanheerdegen (commented):
Just a question about a test, otherwise looks great!

payu/subcommands/catalog_cmd.py
charles-turner-1: Yeah, the builders will parallelise to use pretty much all the resources available to them. I think we can get away with a lot less, typically a large ARE instance sized session is plenty of resources and will get it done in a minute or so.

test/test_sync.py
Qian-HuiChen: I try to mimic the rsync behaviour [here](https://github.com/payu-org/payu/pull/808/changes#diff-4ba73b8e2032c185b77337f2c0e38b4b0c4fbe1cfe359aedbd43d80d243f8594R97-R99)

Please let me know if this is similar to what a remote sync config would look like 👀

Qian-HuiChen:
> > argument for ESM1.5 and 1.6, if `{ensemble: False} is still needed

> It is currently, but we can change that - I think it's just that a default was never set.

@charles-turner-1 I can open a PR to update this today if no one else is working on this atm :)

Qian-HuiChen:
With this [PR](https://github.com/ACCESS-NRI/access-nri-intake-catalog/pull/690) in intake-catalog repo closed, `"ensemble"` is no longer a required argument for ESMBuilders. 
I will need to update the access-nri-intake version in payu pyproject.toml, after the intake-catalog do a new release to include the new changes :)

test/test_datastore.py
```diff
@@ -0,0 +1,181 @@
...
+        'sync': {
+            'enable': True,
+            'base_path': None,
+            'path': None,
+        }
+    }, False),
+    # Sync is enabled with a remote URL, should create datastore in archive
+    ({
+        'sync': {
+            "url": "test.domain",
```
aidanheerdegen, line 97: Where is `url` consumed/used? I can't find it anywhere.

I think it is sufficient to use a path that would be something like `server.domain:/path/on/remote`

Issue #387, opened 2023-11-28
Record last commit hash of previous experiment


aidanheerdegen:
Automated `metadata.yaml` creation is added in this PR https://github.com/payu-org/payu/pull/384

If the commit hash of the the last commit in a previous experiment (when the experiment ID last changed) were saved in `metadata.yaml` it would make it trivially easy to specify the range of commits that constitute an experiment. The `refA..refB` syntax used for selecting a range of git commits automatically excludes the initial commit, `refA` in this example.

Otherwise you would need to walk the commit history of `metadata.yaml` in a repo looking for changes.

Would not be saved if not known (so field would be absent).

aidanheerdegen:
Closed in #384

jo-basevi:
Hi @aidanheerdegen, just following on from the conversation last week of adding this in. I think in that initial metadata PR, https://github.com/payu-org/payu/pull/384#issuecomment-1853316336, it was decided against adding a `parent_experiment_commit` as parent experiment was defined as the model run that has the restarts.

So maybe it needs a name that doesn't conflict with `parent_experiment` or `previous_experiment` and highlights that it is a git commit of the start of the configuration? `base_configuration_commit`?

PR #811, opened 2026-07-24
Add parent branch off time in metadata

### LINKED ISSUE #809: Record parent branch time in `metadata.yaml` when starting from a restart
Hi,

Processing the outputs from ESM1.6 for CMORisation I have found a gap which makes it difficult to describe the parent / child relationship when an experiment branches out from a previous one.

Payu already records `parent_experiment` (UUID) in `metadata.yaml` when branching from a restart, which is great. The one thing missing is the **model time at which the restart was taken in the parent run**.

This matters for two reasons:

- **CMIP compliance** : `branch_time_in_parent` (a float in the parent's time units, e.g. `182135.0 days since 0001-01-01`) is a mandatory global attribute in CMORised output. Without it you can't pass compliance checking. Right now users have to dig this out manually from restart files or run logs every time.
- **Provenance** : the UUID tells you *which* parent was used, but not *where* in that run you branched from. That's a gap in the audit trail.

Payu already reads the restart files at setup time to determine the run start point, so the information is there. It would be very useful to also write it into `metadata.yaml` as `branch_time_in_parent` and `parent_time_units`.

Qian-HuiChen:
In this PR, when a restart path is provided to `payu checkout -b`, payu gets the model finish time from the restart files and records it as `parent_experiment_branch_time` in `metadata.yaml`.

For example: `parent_experiment_branch_time: '1900-10-01T00:00:00'`. This field gets updated when creating another new branch.
If restart path is not provided, payu will remove `parent_experiment_branch_time ` field from metadata.

Feature requested by CMIP developers and closes #809 .


Qian-HuiChen:
I will need to update schema https://github.com/ACCESS-NRI/schema/blob/main/au.org.access-nri/model/output/experiment-metadata/1-0-3.json

--- Edit: ---
PR opened: https://github.com/ACCESS-NRI/schema/pull/75

payu/branch.py
```diff
@@ -219,14 +220,17 @@ def checkout_branch(branch_name: str,
...
         prior_restart_path = check_restart(restart_path=restart_path,
                                            archive_path=metadata.archive_path)
+        branch_off_time = get_branch_off_time(prior_restart_path, lab)
 
     # Create/update and commit metadata file
     metadata.write_metadata(set_template_values=True,
                             restart_path=prior_restart_path,
-                            parent_experiment=parent_experiment)
+                            parent_experiment=parent_experiment,
+                            branch_off_time = branch_off_time)
```
aidanheerdegen:
```suggestion
                            branch_off_time=branch_off_time)
```
Remove spaces for consistency.

payu/branch.py
```diff
@@ -237,6 +241,23 @@ def checkout_branch(branch_name: str,
...
+def get_branch_off_time(restart_path: Path, lab: Laboratory) -> Optional[str]:
+    """Get the model time of the provided restart path, return a string or None"""
+    try:
+        expt = Experiment(lab, metadata_off = True)
+        expt.restart_path = restart_path
+        datetimes = expt.get_model_restart_datetimes()
+
+        model_finish_time = datetimes.get('model_finish_time', None)
+        if model_finish_time:
+            return model_finish_time.strftime('%Y-%m-%dT%H:%M:%S')
```
aidanheerdegen:
```suggestion
        return datetimes['model_finish_time'].strftime('%Y-%m-%dT%H:%M:%S')
```
I think it is possible to not have any gate logic here as you're in a `try/except` block so it just won't work if there isn't a `model_finish_time` key.

payu/branch.py
```diff
@@ -237,6 +241,23 @@ def checkout_branch(branch_name: str,
     switch_symlink(Path(lab.archive_path), control_path, experiment, 'archive')
     switch_symlink(Path(lab.work_path), control_path, experiment, 'work')
 
+def get_branch_off_time(restart_path: Path, lab: Laboratory) -> Optional[str]:
+    """Get the model time of the provided restart path, return a string or None"""
+    try:
+        expt = Experiment(lab, metadata_off = True)
+        expt.restart_path = restart_path
+        datetimes = expt.get_model_restart_datetimes()
```
aidanheerdegen: This is very cool! So simple! Nice creative reuse of existing functionality.

payu/metadata.py
```diff
@@ -37,6 +37,7 @@
 CREATED_FIELD = "created"
 MODEL_FIELD = "model"
 METADATA_FILENAME = "metadata.yaml"
+BRANCH_OFF_TIME_FIELD = "branch_off_parent_time"
```
aidanheerdegen: I'm not sure about this name. The suggestion in the issue was `branch_time_in_parent`.

Would you be happy with that?

aidanheerdegen (changes requested):

jo-basevi (commented):
Looks good @Qian-HuiChen!

Do we want to also get the restart datetime when payu generates new metadata outside of `payu clone/checkout` - e.g. when `payu setup --new-uuid` runs? In `Experiment.init()` before it sets up metadata, it could try getting `self.model.get_restart_datetime(self.prior_restart_path)`.

I wondered if we could initialise the experiment to setup the metadata (to re-use code) in `checkout_branch()`. We would have to setup `restart:` in `config.yaml` before. But I think checkout_branch passes additional arguments of `keep_uuid` and `parent_experiment` so maybe it doesn't make it simpler?

I noticed if restart directories aren't found in checkout branch, it is set to warnings. If we are saving restart information in the metadata.yaml now, maybe it should be upgraded to errors?

aidanheerdegen:
And just a note that this doesn't work well for experiments where the restart files have had their model time modified, e.g. CMIP7 ensembles where ensemble members are generated by choosing different start dates and normalising restarts back to a common start date. In that case it would work if the `clone` or `branch` was done *before* the time modification script was run.

Do we want to be able to suppress this in cases where it would be incorrect? Is that *yet-another* command line option flag?

payu/metadata.py
Qian-HuiChen: The issue mentioned that `branch_time_in_parent` is a float expressed in the parent experiment's time units (e.g., 182135.0 days since 0001-01-01). In payu's metadata.yaml, this value would instead be displayed as the model time in a date-time format (e.g., 1950-10-01T00:00:00).
I was unsure whether it would be appropriate to reuse the same name (`branch_time_in_parent`) given the difference in representation, so I introduced a new field name. If this naming difference is not a concern, I am happy to reuse `branch_time_in_parent` for consistency with the issue.

aidanheerdegen: It's a good point, but I think any potential confusion is outweighed by using a well known name which is discoverable and its purpose is clear. It's also obviously a timestamp, which also helps to make it obvious (and we have a schema!).

aidanheerdegen:
Another question: does this method rely on the `restartXXX` and `ouputXXX` directories still being co-located in a laboratory directory?

payu/metadata.py
joshuatorrance: It looks like similar fields are going to be added to the output file metadata (i.e the global attrs in the netCDF files). These are sorted alphabetically in the files and documentation so it would be nice if these could be sorted next to related fields (e.g. `parent_experiment_uuid`).

How about `parent_experiment_branch_time` for the name?

Qian-HuiChen:
> Do we want to be able to suppress this in cases where it would be incorrect? Is that yet-another command line option flag?

Thanks for the additional information about time modification script. Yes, it is easy and practical to by default write the parent_branch_time, and suppress it if `record_parent_branch_time` is `False` in config. Implemented in [b171c7a](https://github.com/payu-org/payu/pull/811/commits/b171c7a554c3e0b8b076b9b579e5bf826934f3c4).

> Another question: does this method rely on the restartXXX and ouputXXX directories still being co-located in a laboratory directory?

I tested it by moving the `restartXXX` to a `/scratch/$PROJ/$USER/tmp`. Payu still gets the parent branch time and adds it to metadata.yaml.

Qian-HuiChen:
>Do we want to also get the restart datetime when payu generates new metadata

Sounds a great idea. In [b171c7a](https://github.com/payu-org/payu/pull/811/commits/b171c7a554c3e0b8b076b9b579e5bf826934f3c4), I refactored the experiment.init to involve `_get_parent_branch_time`, which is reused in checkout_branch. The restart path is added to `config.yaml` in `checkout_branch` before calling `expt.init()`. Manual test confirms that `payu setup --new-uuid` adds the `parent_experiment_branch_time` into metadata.

>if restart directories aren't found in checkout branch, it is set to warnings

This is updated to errors.PayuFileNotFound in [b171c7a](https://github.com/payu-org/payu/pull/811/commits/b171c7a554c3e0b8b076b9b579e5bf826934f3c4).

payu/metadata.py
Qian-HuiChen: Sounds good! Thanks for the additional information (sorted alphabetically) as well.
I adopt the name to `parent_experiment_branch_time` in metadata.yaml in commit
[b171c7a](https://github.com/payu-org/payu/pull/811/commits/b171c7a554c3e0b8b076b9b579e5bf826934f3c4)

payu/branch.py
Qian-HuiChen: Yeah this makes sense and it makes the code cleaner. Updated in https://github.com/payu-org/payu/pull/811/changes#diff-55ea2a26a18367691eafb8c83d27151ee5f0213f2a2b394f54ba10956fc9a70dR185-R192

payu/branch.py
Qian-HuiChen: I see. Updated as https://github.com/payu-org/payu/pull/811/changes#diff-55ea2a26a18367691eafb8c83d27151ee5f0213f2a2b394f54ba10956fc9a70dR131-R134

aidanheerdegen:
As this PR is modifying the `metadata.yaml` schema and making a related change, it would be good to close https://github.com/payu-org/payu/issues/387 by adding the parent experiment commit hash. Following on from above maybe call it `parent_experiment_branch_commit`? @joshuatorrance may have an opinion ...

joshuatorrance:
> As this PR is modifying the `metadata.yaml` schema and making a related change, it would be good to close #387 by adding the parent experiment commit hash. Following on from above maybe call it `parent_experiment_branch_commit`? @joshuatorrance may have an opinion ...

Looks good to me. Leading with "parent_experiment" should group it conceptually and alphabetically it with the other parent fields.

Do you think this one should be added to output file schema (i.e. output data spec)?

Qian-HuiChen:
I am not quite sure about the approach for `parent_experiment_commit_hash` here.

### --- Approach 1 ---

1. Record the latest commit hash that creates/updates `metadata.yaml` as `metadata_commit_hash` in the metadata (as discussed in #387).
2. When `payu checkout -r restart_path`, find the metadata file associated with the given restart path -> read the previous `metadata_commit_hash`, and record it as `parent_experiment_commit_hash` in the new metadata.

With this approach, users can compare `parent_experiment_commit_hash` with the current commit hash to see the metadata and control directory changes.
However, this is not necessarily the commit hash when the given restart path was created. It may not match the `parent_branch_time` either if users do additional runs after the metadata is updated. Would `parent_experiment_commit_hash` become misleading in this case?

### --- Approach 2 ---

1. Record the git log in all archives (#813).
2. Find the git commit hash associated with `Run00N`, where `N` matches the given `restart00N`.

With this, `parent_experiment_commit_hash` would be the commit where `parent_experiment_branch_time` is exactly the model finish time. If this is the intended behaviour, I would prefer to address it in a separate PR, after issue #813 is resolved.

@aidanheerdegen @joshuatorrance Could you please let me know what information `parent_experiment_commit_hash` is carrying?

aidanheerdegen:
I think we just need to use the git REF, checkout/clone the repo and record the current commit before making a commit that updates the `metadata.yaml`.

docs/source/usage.rst
```diff
@@ -473,6 +473,49 @@ containing the restart.
...
+   |                  | ``parent_experiment_branch_time``      | The model time corresponding to the      |
+   |                  |                                        | restart path used to create this branch. |
+   +                  +----------------------------------------+------------------------------------------+
+   |                  | ``parent_experiment_branch_commit``    | The Git commit hash of the source branch |
+   |                  |                                        | when the new branch was created.         |
+   +                  +----------------------------------------+------------------------------------------+
+   |                  | ``schema_version``                     | The version of the ACCESS-NRI metadata   |
+   |                  |                                        | schema used for this experiment.         |
+   +------------------+----------------------------------------+------------------------------------------+
+
```
aidanheerdegen, line 518: Nice!

payu/branch.py
```diff
@@ -196,6 +195,9 @@ def checkout_branch(branch_name: str,
 
         parent_experiment = uuid
 
+    # Build a dictionary to store information about parent experiment
+    parent_info = {'parent_experiment': parent_experiment, 'parent_hash': parent_hash}
+
```
aidanheerdegen, line 200: Is it worth checking `parent_hash` is not None before adding it to `parent_info`?

payu/experiment.py
```diff
@@ -73,7 +73,8 @@ def wrapper(self, *args, **kwargs):
 
 
 class Experiment(object):
-    def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config_path=None, is_new_experiment=False):
+    def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config_path=None, 
+                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=dict()):
         self.init_timings()
```
aidanheerdegen: It is often [not a good idea to use mutable objects as function argument defaults](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). Ask me how I know.

payu/experiment.py
```diff
@@ -126,8 +127,16 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
 
         self.set_output_paths()
 
-        # Create metadata file and move to archive
-        self.metadata.write_metadata(restart_path=self.prior_restart_path)
+        # Add parent_branch_time and parent_hash to parent_info
+        if self.config.get('record_parent_branch_commit', True) is False:
```
```suggestion
        if not self.config.get('record_parent_branch_commit', True):
```
Just pointing out can just check the negation of it if you want. It is perfectly fine to specifically check `is False`, but on the internet there are a lot of opinions that it is poor practice to check `is False` or `is True` because there are other python things that are "truthy".

aidanheerdegen (commented):
Looks good, just a couple of suggestions/questions.

payu/experiment.py
Qian-HuiChen: How do you know?

aidanheerdegen: I made that mistake not once, but twice! And got strange behaviour that was difficult to debug. I felt really silly the second time.

payu/branch.py
Qian-HuiChen: I am not quite sure about the concern here.
For new branch: If `parent_hash` is `None`, the `parent_experiment_branch_commit` field will be pop out in metadata.
For existing branch: `parent_hash` is `None`, but the `metadata` is not written/changed.
Could you please give more information about this?

payu/experiment.py
Qian-HuiChen: I see. I update it in [ddd26c2](https://github.com/payu-org/payu/pull/811/commits/ddd26c20414b65cc49278e2c818594c7122d7d56).

payu/experiment.py
Qian-HuiChen: To avoid future me from painfully debugging, I replace the `parent_info=dict()` to `parent_info=None` in [ddd26c2](https://github.com/payu-org/payu/pull/811/commits/ddd26c20414b65cc49278e2c818594c7122d7d56)

payu/branch.py
aidanheerdegen: My comment didn't take into account the broader context. Thanks for explaining.

aidanheerdegen (approved):
Looks good, thanks!

Issue #805, opened 2026-07-17
payu clone interactive mode improvement


Qian-HuiChen:
`payu clone` in interactive mode cannot construct a command of:
```
payu clone -B <source_branch> <url> <local_dir>
```
When the user chooses to clone from the same branch, the interactive workflow automatically adds the `--keep-uuid` flag and request for a restart path. As a result, there is no option to keep the same branch while using the default restart configuration.

If this no-restart behaviour is intentional and allowed, payu should also display a warning that this is not recommended and explain the implications of doing so.

anton-seaice:
In my view it's impractical to specify all of `-B` `-b` and `--restart` flags when typing in a payu clone command, so could it be enforced later? (at setup maybe)

Qian-HuiChen:
When typing in the payu clone command (non-interactive mode), `-B`, `-b` and `--restart` flags are optional, only `<url>`  and `<local_dir>` are required field.

anton-seaice:
Oh I read this backwards, I thought the suggestions was to change the manual clone command to match interactive clone

Qian-HuiChen:
I am thinking the other way: changing the interactive mode to match the manual command 😄

Qian-HuiChen:
Closed by #814

PR #814, opened 2026-07-29
Ask for shortpath and restart path in payu clone.

### LINKED ISSUE #810: payu clone should allow specifying archive path
It is common for users to store experiments across multiple projects, depending on where storage is available. Having to change the default project each time is inconvenient.
It would be useful for `payu clone` interactive mode to prompt:
```
Do you want to override the archive path? (default: </path/to/archive>)
```
If the user chooses to do so, the designated archive path will be written to `config.yaml` as the `shortpath` setting.

Qian-HuiChen:
This PR improves the interactive `payu clone` workflow by:
* adding a prompt asking, `Do you want to override the shortpath? (Default: '/scratch/{project}')`, where `project` is read from the environment variable.
* adding a prompt asking whether the user wants to specify a restart path for both the new experiment and existing experiment.

Closes #810 and #805.


Qian-HuiChen:
If the user answers `yes` to `Do you want to override the shortpath?`, `payu` adds `--shortpath /new/path/` to the `payu clone` command. I am thinking this way because users can copy the complete command and reuse it in scripts.

During the cloning process, `payu` also prints a confirmation message:
```
Added 'shortpath: /scratch/$new-PROJECT' to configuration file: config.yaml
```

Qian-HuiChen:
@anton-seaice will the updated payu clone workflow meet your needs? See [flowchart](https://github.com/ACCESS-NRI/payu/blob/5e5885b5a2a1f91efd08d1951eee1648706458cc/docs/diagrams/payu_clone_flowchart.svg).

anton-seaice:
It's ok by me

The old method forced users to make a new experiment branch when cloning from a supported configuration, which is possibly a good thing (rather than just letting them running experiments from the config branch)

aidanheerdegen:
> The old method forced users to make a new experiment branch when cloning from a supported configuration, which is possibly a good thing (rather than just letting them running experiments from the config branch)

Is that because it is possible to clone a config and say it is not a new experiment?

I'm wondering if the question "is this a new experiment" shouldn't be asked if there is no experiment UUID? If there is no UUID it has to be a new experiment, which would be the case for the configs.

Qian-HuiChen:
Instead of asking "is this a new experiment, yes or no", should payu ask 
```
Is this a new experiment?
- Yes, create a new UUID as a new experiment
- No, keep using the existing UUID in the configuration
```
What do you think?

aidanheerdegen:
> What do you think?

Yes, I think that is much clearer. Thanks. But I would use experiment rather than configuration.

Qian-HuiChen:
This PR also fixes the issue #829 about missing flowchart,  since this PR is modifying the same file.

payu/payu_clone_flowchart.ascii
atteggiani, line 1: This file should also be added to the `pyproject.toml`'s [package-data](https://github.com/payu-org/payu/blob/d6f80d728e95aae2286290cd761b445864ee9681/pyproject.toml#L72).
```
[tool.setuptools.package-data]
"payu" = [
  "telemetry/*.json", 
  "payu_clone_flowchart.ascii",
]
```

Qian-HuiChen: Oops Sorry that I missed this part :) Thanks!
Implemented in [dfe9770](https://github.com/payu-org/payu/pull/814/commits/dfe9770cece7bf225a873dafc50ca4afc93e6d1d)

docs/diagrams/payu_clone_flowchart.svg
aidanheerdegen, line 1: What do the red arrows/path signify?

aidanheerdegen (dismissed):
Looks great, thanks! Just one small question ...

docs/diagrams/payu_clone_flowchart.svg
Qian-HuiChen: The red arrows distinguish the workflow for `cloning from a branch` versus `cloning from a tag/commit`. If the user chooses `cloning from a tag/commit`, they skip the "Is this a new experiment?" prompt and go straight to "Name your new branch".

PR #821, opened 2026-07-30
Clean up TODO messages in comments

### LINKED ISSUE #807: TODO comments should reference an issue
There are 71 comments in 23 files with `TODO`.
e.g.
```
payu/experiment.py:# TODO: To be removed
payu/experiment.py:        # TODO: replace with dict, check versions via key-value pairs
payu/experiment.py:        # TODO: __init__ should not be a config dumping ground!
payu/experiment.py:        self.expand_shell_vars = True   # TODO: configurable
payu/experiment.py:        # TODO: Move to run/collate/sweep?
payu/experiment.py:        # TODO: Move this stuff somewhere else
payu/experiment.py:        # TODO: Rename this to self.submodels
payu/experiment.py:        # TODO: Consolidate this profiling stuff
payu/experiment.py:        # TODO: per-model output streams?
payu/experiment.py:        # TODO: check case counter == 0
payu/experiment.py:        # TODO: Move this function to a setup file if setup is moved to
payu/experiment.py:        # TODO: This just copies the existing config.yaml file, but we should
payu/experiment.py:        # TODO: Make this more configurable
payu/experiment.py:        # TODO: More uniform support needed here
payu/experiment.py:            # TODO: Check for MPI library mismatch across multiple binaries
payu/experiment.py:            # TODO: New Open MPI format?
payu/experiment.py:        # TODO: Consider making this default
payu/experiment.py:            # TODO: Replace with mpirun -x flag inputs
payu/experiment.py:        # TODO: Need a model-specific cleanup method call here
payu/experiment.py:        # TODO: Fix the IO race conditions!
payu/experiment.py:        # TODO: model outstreams and pbs logs need to be handled separately
payu/experiment.py:            # TODO: New path may still exist!
payu/experiment.py:        # TODO: Previous logic was to prune all restarts if self.repeat_run
```

There should be an issue number associated with each comment (the same issue can apply to more than one comment).

Without this it is very hard to action the TODO, or know if it is still relevant.

Ideally we can work backwards and find relevant issues, or create new ones, and add the number to the comment.

aidanheerdegen:
Clean up TODO messages in comments. 

Comment is assessed. If it is still relevant and should be fixed an issue number is added to the comment. If there is not an existing issue one is created. 

If no longer relevant there are 2 possible courses of action:

1. Delete the comment
2. Remove TODO and convert to a code comment

Fixes #807


Qian-HuiChen (approved):
Looks good to me. Thanks for creating the issues and linking them!

PR #832, opened 2026-08-12
Add job dependency: run > collate > postscript > sync.

### LINKED ISSUE #774: Handle job dependency
HPCpy handles job dependency using the underlying schedulers native syntax. We should decide if this is enough for payu's requirements.

See:
https://github.com/ACCESS-NRI/hpcpy/blob/d6924bd05272eb5452d78b3d74ef22bf8c2e59b6/hpcpy/client/pbs.py#L90

Then we can address issues like this: https://github.com/payu-org/payu/issues/463

### LINKED ISSUE #787: A separate section of userscript status in payu status
Since we implemented collation in `payu status`, we can also do it for userscript.
Expected:
```
===============================
Run: 39
Run ID:
:::
  ----------- Collate Info ------------
::
  ---------- Userscript Info ----------
::
===============================
```

Qian-HuiChen:
This PR adds job dependency to payu in the order of run > collate > postscript > sync. Sync now includes the last output and begins after postscript is finished.
Each step generates a job file.
This PR also updates `payu status` to
```
Run: 45
  Job ID:            176017091.gadi-pbs
  Run ID:            bcdff5eeda6d72d0c0c45b8c6e0c39c83f244da8
  Stage:             archive
::
  Error Log:         xxx
  Job File:          xxx
 ------------- Collate Info -------------
  Job ID:            176017180.gadi-pbs
  Starts After:      ['176017091.gadi-pbs']
  Stage:             exited
  Total Queue Time:  0h 1m 39s
  Output Log:        xxx
  Error Log:        xxx
  Job File:          xxx
  ------------- Postscript Info -------------
  Job ID:            176017185.gadi-pbs
  Starts After:      ['176017091.gadi-pbs', '176017180.gadi-pbs']
  Stage:             exited
  Output Log:        xxx
  Error Log:         xxxx
  Job File:          xxx
  ------------- Sync Info -------------
  Job ID:            176017188.gadi-pbs
  Starts After:      ['176017091.gadi-pbs', '176017180.gadi-pbs', '176017185.gadi-pbs']
  Stage:             exited
  Total Queue Time:  0h 2m 39s
  Output Log:        xxx
  Error Log:        xxx
  Job File:          xxx
```
Closes #774 
Closes #787

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -416,6 +416,24 @@ Other experiment runs will not be harmed by this command.
 Postprocessing
 ==============
 
+By default, payu submits all enabled postprocessing jobs automatically after the model run has completed.
+Postprocessing includes 
+- collate, 
+- postscript, 
+- sync. 
+These jobs can also be run manually if required.
+When payu submits postprocessing jobs, it adds job dependencies to ensure that they run in the correct order:: 
+
+   collate -> postscript -> sync
+
+If a postprocessing job failed, any subsequent job(s) will be terminated and removed from the queue.
+You can use ``payu status`` to check the status of all postprocessing jobs.
+To clean up postprocessing jobs that have been killed, please run ``payu status --update`` once.
+
+
+Collate
+--------
+
 Model output in parallel jobs is sometimes divided across several files, which
 can be inconvenient for analysis. Payu offers a ``collate`` subcommand to
 collate these separated files into a single file. This is only necessary, and 
@@ -446,6 +464,21 @@ using the ``-c`` option::
 
 This allows payu to use the collation settings from the specified ``config.yaml``
 while reading the uncollated files from the directory provided with ``-d``.
+
+When ``payu collate`` is called, payu will submit the subsequent postscript and sync jobs 
+after collation has completed, if they are enabled in the configuration file. 
+
+Postscript
+----------
+
+You may configure the postscript job in the configuration file (see :ref:`config`),
+which will run after the model run and collation (if enabled) has completed.
+To manually run the postscript for a specific run ``K``, please use::
+
+   payu postscript -i K
+
+Sync
+----
 
 To manually sync experiment output files to a remote archive, firstly ensure
 that ``path`` in the ``sync`` namespace in ``config.yaml``, 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -59,14 +59,6 @@ core_modules = ['python', 'payu']
 # Default payu parameters
 default_restart_freq = 5
 
-# Default job dependency
-DEFAULT_DEPENDENCIES = {
-    "collate": ["run"],
-    "postscript": ["run", "collate"],
-    "sync": ["run", "collate", "postscript"],
-}
-
-
 def timeit(time_name):
     """Decorator to time a function and store the elapsed time in seconds
     to the timings dictionary in the class"""
@@ -980,14 +972,16 @@ class Experiment(object):
             sp.check_call(shlex.split(cmd))
 
     def build_workflow(self):
-        """Build the workflow for the current run, including run, collate, postscript and sync steps."""
-        workflow = []
+        """Build the workflow for the current experiment, e.g.,
+        {"run": None, "collate": None, "postscript": None, "sync": None}. 
+        The value will be updated as job id when available"""
+        workflow = dict()
         if self.config.get('collate', {}).get('enable', True):
-            workflow.append('collate')
+            workflow['collate'] = None
         if self.postscript:
-            workflow.append('postscript')
+            workflow['postscript'] = None
         if self.config.get('sync', {}).get('enable', False):
-            workflow.append('sync')
+            workflow['sync'] = None
         return workflow
     
     @timeit("payu_collate_duration_seconds")
@@ -1054,14 +1048,14 @@ class Experiment(object):
     def set_userscript_env_vars(self):
         """Save information of output directories and current run to
         environment variables, so they can be accessed via user-scripts"""
-        os.environ.update(
-            {
+        update_env_vars = {
                 'PAYU_CURRENT_OUTPUT_DIR': self.output_path,
                 'PAYU_CURRENT_RESTART_DIR': self.restart_path,
                 'PAYU_ARCHIVE_DIR': self.archive_path,
                 'PAYU_CURRENT_RUN': str(self.counter)
             }
-        )
+        os.environ.update(update_env_vars)
+        return update_env_vars
 
     def run_userscript(self, script_cmd: str, type: str):
         """Run a user defined script or subcommand at various stages of the
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -1006,7 +1006,7 @@ class Experiment(object):
 
     @timeit("payu_sync_duration_seconds")
     def sync(self):
-        # Update collate stage to running
+        # Update sync stage to running
         telemetry.update_job_file(
             file_path=self.get_job_file(type='sync'),
             data={"stage": "running"}
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -970,19 +970,6 @@ class Experiment(object):
                 expt=self.counter
             )
             sp.check_call(shlex.split(cmd))
-
-    def build_workflow(self):
-        """Build the workflow for the current experiment, e.g.,
-        {"run": None, "collate": None, "postscript": None, "sync": None}. 
-        The value will be updated as job id when available"""
-        workflow = dict()
-        if self.config.get('collate', {}).get('enable', True):
-            workflow['collate'] = None
-        if self.postscript:
-            workflow['postscript'] = None
-        if self.config.get('sync', {}).get('enable', False):
-            workflow['sync'] = None
-        return workflow
     
     @timeit("payu_collate_duration_seconds")
     def collate(self):
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -38,7 +38,6 @@ from payu.fsops import make_symlink, read_config, movetree
 from payu.fsops import list_sorted_archive_dirs
 from payu.fsops import run_script_command
 from payu.fsops import get_size, str_to_bool
-from payu.fsops import get_job_id_given_job_type
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 from payu.models import index as model_index
 from payu.runlog import Runlog
@@ -49,6 +48,7 @@ from payu.metadata import Metadata
 import payu.telemetry as telemetry
 from payu.git_utils import get_git_repository
 import payu.errors as errors
+
 
 # Setup logger
 logger = logging.getLogger(__name__)
@@ -88,6 +88,7 @@ class Experiment(object):
                  runlog_off=False, repeat=False):
         self.init_timings()
         self.lab = lab
+        self.config_path = config_path
         # Check laboratory directories are writable
         self.lab.initialize()
 
@@ -970,16 +971,6 @@ class Experiment(object):
             restart_volume_gb=get_size(self.restart_path),
         )
 
-        collate_config = self.config.get('collate', {})
-        collating = collate_config.get('enable', True)
-        if collating:
-            cmd = '{python} {payu} collate -i {expt}'.format(
-                python=sys.executable,
-                payu=self.payu_path,
-                expt=self.counter
-            )
-            sp.check_call(shlex.split(cmd))
-
         if self.config.get('hpctoolkit', False):
             cmd = '{python} {payu} profile -i {expt}'.format(
                 python=sys.executable,
@@ -988,9 +979,17 @@ class Experiment(object):
             )
             sp.check_call(shlex.split(cmd))
 
-        # Call postprocessing step after collation
-        self.postprocess()
-
+    def build_workflow(self):
+        """Build the workflow for the current run, including run, collate, postscript and sync steps."""
+        workflow = []
+        if self.config.get('collate', {}).get('enable', True):
+            workflow.append('collate')
+        if self.postscript:
+            workflow.append('postscript')
+        if self.config.get('sync', {}).get('enable', False):
+            workflow.append('sync')
+        return workflow
+    
     @timeit("payu_collate_duration_seconds")
     def collate(self):
         """ Run model collation and record the time taken in seconds to run collation"""
@@ -1023,67 +1022,6 @@ class Experiment(object):
     def profile(self):
         for model in self.models:
             model.profile()
-
-    def get_dependency_job_ids(self, current_job_type: str) -> list:
-        if current_job_type not in self.dependencies:
-            raise errors.PayuRuntimeError(
-                f"Current job type '{current_job_type}' not found in dependencies list: "
-                f"{self.dependencies}"
-            )
-
-        depends_on = []
-        for job_type in self.dependencies[current_job_type]:
-            job_id = get_job_id_given_job_type(self, job_type)
-
-            if job_id:
-                depends_on.append(job_id)
-        return depends_on
-            
-            
-    def postprocess(self):
-        """Submit any postprocessing scripts or remote syncing if enabled"""
-
-        # First submit postprocessing script
-        if self.postscript:
-            self.set_userscript_env_vars()
-            envmod.setup()
-            envmod.module('load', 'pbs')
-
-            # Postscript depends on run and collate jobs, if any of them exist
-            depends_on = self.get_dependency_job_ids('postscript')
-            
-            # Submit through HPCpy
-            # Job name is set to "payu_postscript"
-            postscript_job = client.submit(job_script = f"{self.postscript}",
-                      directives = ['-N payu_postscript'],
-                      depends_on = depends_on
-                      )
-            print(f"--- Postscript submitted ---\n"
-                f"Submitted command: {postscript_job.history[0]}\n"
-                f"Job ID: {postscript_job.id}.\n")
-
-            # Write postscript job file to archive/payu_jobs/{run_number}/postscript/{job_id}-gadi-pbs.json
-            telemetry.write_queued_job_file(
-                        archive_path=Path(self.archive_path),
-                        job_id=postscript_job.id,
-                        type='postscript',
-                        scheduler=self.scheduler,
-                        metadata=self.metadata,
-                        current_run=self.counter,
-                        depends_on=depends_on,
-                    )
-
-        # Submit a sync script if remote syncing is enabled
-        sync_config = self.config.get('sync', {})
-        syncing = sync_config.get('enable', False)
-        if syncing:
-            cmd = '{python} {payu} sync -i {expt}'.format(
-                python=sys.executable,
-                payu=self.payu_path,
-                expt=self.counter
-            )
-
-            sp.check_call(shlex.split(cmd))
 
     @timeit("payu_sync_duration_seconds")
     def sync(self):
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -27,6 +27,9 @@ import logging
 # Extensions
 from ruamel.yaml import YAML
 from packaging import version
+import hpcpy
+
+client = hpcpy.client.pbs.PBSClient()
 
 # Local
 import payu
@@ -34,8 +37,8 @@ from payu import envmod
 from payu.fsops import make_symlink, read_config, movetree
 from payu.fsops import list_sorted_archive_dirs
 from payu.fsops import run_script_command
-from payu.fsops import needs_subprocess_shell
 from payu.fsops import get_size, str_to_bool
+from payu.fsops import get_job_id_given_job_type
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 from payu.models import index as model_index
 from payu.runlog import Runlog
@@ -55,6 +58,13 @@ core_modules = ['python', 'payu']
 
 # Default payu parameters
 default_restart_freq = 5
+
+# Default job dependency
+DEFAULT_DEPENDENCIES = {
+    "collate": ["run"],
+    "postscript": ["run", "collate"],
+    "sync": ["run", "collate", "postscript"],
+}
 
 
 def timeit(time_name):
@@ -978,9 +988,8 @@ class Experiment(object):
             )
             sp.check_call(shlex.split(cmd))
 
-        # Ensure postprocessing runs if model not collating
-        if not collating:
-            self.postprocess()
+        # Call postprocessing step after collation
+        self.postprocess()
 
     @timeit("payu_collate_duration_seconds")
     def collate(self):
@@ -1015,6 +1024,22 @@ class Experiment(object):
         for model in self.models:
             model.profile()
 
+    def get_dependency_job_ids(self, current_job_type: str) -> list:
+        if current_job_type not in self.dependencies:
+            raise errors.PayuRuntimeError(
+                f"Current job type '{current_job_type}' not found in dependencies list: "
+                f"{self.dependencies}"
+            )
+
+        depends_on = []
+        for job_type in self.dependencies[current_job_type]:
+            job_id = get_job_id_given_job_type(self, job_type)
+
+            if job_id:
+                depends_on.append(job_id)
+        return depends_on
+            
+            
     def postprocess(self):
         """Submit any postprocessing scripts or remote syncing if enabled"""
 
@@ -1024,13 +1049,29 @@ class Experiment(object):
             envmod.setup()
             envmod.module('load', 'pbs')
 
-            # Name of this PBS job is set to "payu_postscript"
-            cmd = 'qsub -N payu_postscript {script}'.format(script=self.postscript)
+            # Postscript depends on run and collate jobs, if any of them exist
+            depends_on = self.get_dependency_job_ids('postscript')
+            
+            # Submit through HPCpy
+            # Job name is set to "payu_postscript"
+            postscript_job = client.submit(job_script = f"{self.postscript}",
+                      directives = ['-N payu_postscript'],
+                      depends_on = depends_on
+                      )
+            print(f"--- Postscript submitted ---\n"
+                f"Submitted command: {postscript_job.history[0]}\n"
+                f"Job ID: {postscript_job.id}.\n")
 
-            if needs_subprocess_shell(cmd):
-                sp.check_call(cmd, shell=True)
-            else:
-                sp.check_call(shlex.split(cmd))
+            # Write postscript job file to archive/payu_jobs/{run_number}/postscript/{job_id}-gadi-pbs.json
+            telemetry.write_queued_job_file(
+                        archive_path=Path(self.archive_path),
+                        job_id=postscript_job.id,
+                        type='postscript',
+                        scheduler=self.scheduler,
+                        metadata=self.metadata,
+                        current_run=self.counter,
+                        depends_on=depends_on,
+                    )
 
         # Submit a sync script if remote syncing is enabled
         sync_config = self.config.get('sync', {})
@@ -1042,18 +1083,16 @@ class Experiment(object):
                 expt=self.counter
             )
 
-            if self.postscript:
-                print('payu: warning: postscript is configured, so by default '
-                      'the lastest outputs will not be synced. To sync the '
-                      'latest output, after the postscript job has completed '
-                      'run:\n'
-                      '    payu sync')
-                cmd += f' --sync-ignore-last'
-
             sp.check_call(shlex.split(cmd))
 
     @timeit("payu_sync_duration_seconds")
     def sync(self):
+        # Update collate stage to running
+        telemetry.update_job_file(
+            file_path=self.get_job_file(type='sync'),
+            data={"stage": "running"}
+        )
+
         # RUN any user scripts before syncing archive
         envmod.setup()
         pre_sync_script = self.userscripts.get('sync')
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -423,19 +423,3 @@ def str_to_bool(value):
         raise ValueError(f"Input is not acceptable. Please use 'True' or 'False' or '1' or '0'.")
     
 
-def get_job_id_given_job_type(expt, job_type):
-    """Get the job ID out of the most recent job file, given a specific type (run, collate, sync) for the current experiment."""
-    job_dir = Path(expt.archive_path) / "payu_jobs" / str(expt.counter) / job_type
-
-    if not job_dir.exists():
-        return None
-
-    # Get the most recent job file in the job directory
-    job_files = sorted(job_dir.glob("*.json"), key=lambda f: f.stat().st_mtime, reverse=True)
-    if not job_files:
-        return None
-
-    # Read the job ID from the most recent job file
-    with job_files[0].open('r') as f:
-        job_data = json.load(f)
-        return job_data.get("scheduler_job_id", None)
--- a/payu/fsops.py
+++ b/payu/fsops.py
@@ -205,7 +205,9 @@ def check_exe_path(payu_path, pbs_script):
     if not os.path.isabs(pbs_script):
         pbs_script = os.path.join(payu_path, pbs_script)
 
-    assert os.path.isfile(pbs_script)
+    if not os.path.isfile(pbs_script):
+        raise errors.PayuRuntimeError(
+            'Payu executable not found at {0}.'.format(pbs_script))
 
     return pbs_script
 
@@ -419,3 +421,21 @@ def str_to_bool(value):
         return False
     else:
         raise ValueError(f"Input is not acceptable. Please use 'True' or 'False' or '1' or '0'.")
+    
+
+def get_job_id_given_job_type(expt, job_type):
+    """Get the job ID out of the most recent job file, given a specific type (run, collate, sync) for the current experiment."""
+    job_dir = Path(expt.archive_path) / "payu_jobs" / str(expt.counter) / job_type
+
+    if not job_dir.exists():
+        return None
+
+    # Get the most recent job file in the job directory
+    job_files = sorted(job_dir.glob("*.json"), key=lambda f: f.stat().st_mtime, reverse=True)
+    if not job_files:
+        return None
+
+    # Read the job ID from the most recent job file
+    with job_files[0].open('r') as f:
+        job_data = json.load(f)
+        return job_data.get("scheduler_job_id", None)
--- a/payu/status.py
+++ b/payu/status.py
@@ -4,6 +4,7 @@ payu runs by inspecting the job files generated for telemetry,
 scheduler stdout/stderr logs, and querying the scheduler
 """
 from pathlib import Path
+import re
 from typing import Any, Optional
 import warnings
 from datetime import datetime
@@ -454,10 +455,38 @@ def display_expt_paths(expt_paths):
     print("=" * line_width)
 
 
+def parse_exit_status_from_logs(stdout):
+    """
+    Attempt to parse the exit status from stdout/stderr logs.
+    
+    Returns:
+        Exit status code (int) if found, None otherwise
+    """
+    if not stdout or not stdout.exists():
+        logger.debug(f"Stdout file {stdout} does not exist, cannot parse exit status.")
+        return None
+    
+    try:
+        content = stdout.read_text()
+        # Look for a line that is "Exit status: N"
+        match = re.search(r"Exit Status:\s*(\d+)", content)
+        if match:
+            return int(match.group(1))
+    except (IOError, ValueError, OSError) as e:
+        logger.debug(f"Could not read exit status from stdout file {stdout}: {e}")
+    
+    return None
+
+
 def update_postscript_job_file(data, scheduler, job_file, stdout, stderr):
-    """Check if the postscript log file exists,  update the stage to exited and update scheduler info.
-     Otherwise, update from queued to queued/running,
-    Return the updated job file data"""
+    """Check if the postscript log file exists, update the stage to exited and update scheduler info.
+    Otherwise, update to running of status is "R".
+    
+    When the job has aged out of the scheduler, attempt to parse the exit status
+    from the log files as a fallback mechanism.
+    
+    Return the updated job file data
+    """
     job_id = data.get("scheduler_job_id")
         
     # If both stdout and stderr exist, update the stage to exited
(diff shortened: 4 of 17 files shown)


Qian-HuiChen:
image

Qian-HuiChen:
Documentation needs to be updated before this PR gets merged.

payu/fsops.py
```diff
@@ -405,4 +407,22 @@ def get_size(work_path):
...
+    """Get the job ID out of the most recent job file, given a specific type (run, collate, sync) for the current experiment."""
+    job_dir = Path(expt.archive_path) / "payu_jobs" / str(expt.counter) / job_type
+
+    if not job_dir.exists():
+        return None
+
+    # Get the most recent job file in the job directory
+    job_files = sorted(job_dir.glob("*.json"), key=lambda f: f.stat().st_mtime, reverse=True)
+    if not job_files:
+        return None
```
jo-basevi: I think there's a risk another earlier job file could be modified last (with users running `payu status --update`). 

Initially wondered `payu run` could submit everything at once so we know what the required UUIDs, but then running `payu-run` directly on the login node wouldn't submit post-processing jobs. So then could submit all post-processing jobs in the `payu-run` job. But then if `payu collate` was run manually, we might expect the follow-on tasks to be started.. So maybe we can't easily submit everything at once...

As everything is currently sequential, would it make it easier to only pass one dependent job ID to the next job? Because `payu collate` starts after `payu run` we don't need to add `payu run` job dependency to `payu sync`. This means you can get the job ID from the currently running job OR in postscripts/sync case, when post-script job is submitted. This would require adding job ID dependency argument to `payu-collate`, `payu-sync` commands.

jo-basevi: Though I do think this approach of inspecting job files could be useful when someone submits `payu run` then immediately runs another `payu` command that starts another job e.g. `payu run`, `payu collate` as payu can check if is there anything jobs currently running (and probably raise an error in this case).

payu/experiment.py
```diff
@@ -1015,13 +1055,29 @@ def postprocess(self):
...
+            # Write postscript job file to archive/payu_jobs/{run_number}/postscript/{job_id}-gadi-pbs.json
+            telemetry.write_queued_job_file(
+                        archive_path=Path(self.archive_path),
+                        job_id=postscript_job.id,
+                        type='postscript',
+                        scheduler=self.scheduler,
+                        metadata=self.metadata,
+                        current_run=self.counter,
+                        depends_on=depends_on,
+                    )
```
jo-basevi: I wonder if it's worth moving this logic that submits the job and updates the queued job file outside of experiment.py? E.g. there's some similar logic in `cli.py` so could move this to a function there so similar logic lives in the same file. Also fine to leave as-is though as it uses a lot of experiment class values.

payu/experiment.py
jo-basevi, line 1056: Yay it's nice having this removed!

payu/experiment.py
```diff
@@ -1033,18 +1089,16 @@ def postprocess(self):
...
             sp.check_call(shlex.split(cmd))
 
     @timeit("payu_sync_duration_seconds")
     def sync(self):
+        # Update collate stage to running
+        telemetry.update_job_file(
+            file_path=self.get_job_file(type='sync'),
+            data={"stage": "running"}
+        )
+
```
jo-basevi, line 1025: Would love for the sync-ed job file to include the path that files are being synced to- but this can be added in later PR

jo-basevi (commented):
Thanks @Qian-HuiChen - it is going to be great having sync and postscript information displayed in the `payu status`!

payu/subcommands/run_cmd.py
aidanheerdegen, line 1: What about moving the `expt.collate` and `expt.postprocess` calls (and logic) to `run_cmd.runscript`?

That way you could do as @jo-basevi suggested and grab the job IDs directly from the submission, and choose which single dependency is required at that point.

Something like:
```python
        try:
            expt.setup()
            expt.run()
            expt.archive(force_prune_restarts=run_args.force_prune_restarts)
            job_id = expt.scheduler.get_job_id(short=False) 
            # Move logic to query collation status to expt.collate
            collate_job_id = expt.collate(depends_on=job_id)
            if collate_job_id is not None:
                # Postprocess job will depend on collate job
                job_id = collate_job_id
            postprocess_job_id = expt.postprocess(depends_on=job_id)
            run_status = 0
        except:
            run_status = 1
            raise
        finally:
            # Record job information for experiment run
            record_run(
                timings=expt.timings,
                scheduler=expt.scheduler,
                status=run_status,
                config=expt.config,
                file_path=expt.job_file,
                archive_path=Path(expt.archive_path),
            )
```
Note: I have not tested this, and I may have omitted/not-understood some
important details, so please do push back if you this this won't work or
is a bad idea.

aidanheerdegen (commented):
I haven't looked through all the changes, as what I'm suggested below would likely require some other changes.

aidanheerdegen:
After discussing IRL I propose a re-factor of how the control flow logic. 

The high level description of which is: 
- move the orchestration of the experiment run stages to `run_cmd.runscript`
- call the `runcmd` functions directly, rather than use subprocess, and return the job_ids so they can be used in dependency management
- make a postscript subcommand that can be called independently

Still not sure where to put the logic which checks in `config.yaml` if the `sync`/`collate`/`postscript` should run. Could move it into the `runcmd` function, which would mean it gets check if the function is called directly from the command line, but we'd need to duplicate that for the `runscript` in case it was called directly, e.g. `payu-collate`. So could have a separate function in the subcommand file that is called in both places.

Or could create separate `run_collate`,`run_postscript` and `run_sync` functions in `experiment.py`, but that seems a little confusing and messy.

payu/experiment.py
Qian-HuiChen: I made a small change in cli.submit_job in [558abe1](https://github.com/payu-org/payu/pull/832/commits/558abe1954e8bac422ba52b16416cc69ebe57731) so postscript can use cli.submit_job as well

payu/fsops.py
Qian-HuiChen: Good point. I discard this method. Please see the new design in [558abe1](https://github.com/payu-org/payu/pull/832/commits/558abe1954e8bac422ba52b16416cc69ebe57731) that passes down the job id in run_cmd.py

payu/subcommands/run_cmd.py
Qian-HuiChen: Thanks for the idea! I started from this idea and made the changes in [558abe1](https://github.com/payu-org/payu/pull/832/commits/558abe1954e8bac422ba52b16416cc69ebe57731)

payu/experiment.py
Qian-HuiChen: I created an issue #834 to make sure I won't forget to do this :)

Qian-HuiChen:
This is an animation that explains the design in commit [558abe1](https://github.com/payu-org/payu/pull/832/commits/558abe1954e8bac422ba52b16416cc69ebe57731)


https://github.com/user-attachments/assets/73cbf336-eb08-4f41-ae1e-d5189aa38996

payu/subcommands/collate_cmd.py
```diff
@@ -11,22 +11,42 @@
...
+            lab_path=expt.lab.basepath,
+            dir_path=expt.output_path,
+            depends_on=depends_on,
+            exist_workflow=True
+        )
+    return job_id
+
 
-def runcmd(model_type, config_path, init_run, lab_path, dir_path, dry_run=False):
+def runcmd(model_type, config_path, init_run, lab_path, dir_path, dry_run=False, depends_on=None, exist_workflow=False):
```
jo-basevi, line 40: When submitted from the command line, it's just `payu collate -i {expt_counter}`. A number of these have defaults in the `args.py`, so wondering if that should be represented here? e.g. 



```suggestion
def runcmd(model_type=None, config_path=None, init_run=None, lab_path=None, dir_path=None, dry_run=False, depends_on=None, exist_workflow=False):
```

Then it can just be called with 
````
runcmd(init_run=expt.counter, depends_on=depends_on, exist_workflow=True)
````

File: payu/subcommands/collate_cmd.py [hunk 5]
```diff
@@ -11,22 +11,42 @@
...
 
-    pbs_config = fsops.read_config(config_path)
+    pbs_config = read_config(config_path)
     pbs_vars = cli.set_env_vars(init_run=init_run,
                                 lab_path=lab_path,
                                 dir_path=dir_path)
 
+    # Pass exist_workflow as an environment variable so runscript() can access it
+    pbs_vars['PAYU_EXIST_WORKFLOW'] = str(exist_workflow).lower()
```
jo-basevi, line 48: Maybe worth adding exist_workflow to `pbs_vars` so all Payu environment variables are set in the same palce and it's only set in the environment when set to True?

payu/subcommands/collate_cmd.py
```diff
@@ -109,19 +130,43 @@ def runscript(**run_args):
...
+                    next_depends_on = sync_job_id
+
+        # If collation succeeds, then collate_status is set to 0
         collate_status = 0
     except:
         # If collation fails, then collate_status is set to 1
         collate_status = 1
+        # If collate fails, remove job files of all dependent jobs
+        # Only when collate builds its own workflow
+        if not exist_workflow and workflow:
```
jo-basevi, line 166: Could get into issues when the `workflow` variable doesn't exist yet, e.g. job fails in `expt.collate()`. Could set `workflow = None` before the `try:`.

payu/experiment.py
```diff
@@ -56,6 +59,13 @@
 # Default payu parameters
 default_restart_freq = 5
 
+# Default job dependency
+DEFAULT_DEPENDENCIES = {
+    "collate": ["run"],
+    "postscript": ["run", "collate"],
+    "sync": ["run", "collate", "postscript"],
+}
+
```
jo-basevi, line 68: Is this safe to remove?

payu/status.py
```diff
@@ -446,4 +451,29 @@ def display_expt_paths(expt_paths):
...
+    job_id = data.get("scheduler_job_id")
+        
+    # If both stdout and stderr exist, update the stage to exited
+    if stdout and stderr and stdout.exists() and stderr.exists():
+        data["stage"] = "exited"
+
+        # Update the scheduler info by calling qstat -xf -F json job_id
+        if job_id and scheduler:
+            data["scheduler_job_info"] = scheduler.get_job_info(job_id)
+            data["payu_postscript_status"] = data["scheduler_job_info"].get("Jobs", {}).get(job_id, {}).get("Exit_status")
```
jo-basevi, line 470: Job may no longer be present in the scheduler logs by the time `payu status` runs. Unfortunately we then don't know job exit code unless we pass the standard output logs.

This will also be an issue with the `payu status --update` might assume the postscript job has failed. 

We could add a scheduler method that attempts to parse the Exit Code from the stdout logs, but will need to gracefully fail if it isn't able to.

jo-basevi (commented):
Hi, I just had an initial read through of the new changes which I thought looked good.

I haven't tested any of the changes yet, but are you able to give a short description of what you have manually tested?

payu/subcommands/postscript_cmd.py
```diff
@@ -0,0 +1,70 @@
...
+    postscript = pbs_config.get('postscript', {})
+    if not postscript:
+        print("No postscript commands found in config.yaml. Skipping postscript submission.")
+        return None
+
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+
+    expt.set_userscript_env_vars()
```
jo-basevi, line 51: I wonder if payu can always pass the environment variables set here to the PBS job? 

Also, does the following postscript in ESM1.6 configurations work with these changes:

```
postscript: -v PAYU_CURRENT_OUTPUT_DIR,PROJECT  -lstorage=${PBS_NCI_STORAGE} ./scripts/NetCDF-conversion/UM_conversion_job.sh
```

https://github.com/ACCESS-NRI/access-esm1.6-configs/blob/2b521812e7403bda4912e6173efe092f460c57b5/config.yaml#L168

Qian-HuiChen: Thanks for raising this issue. I update it to https://github.com/payu-org/payu/pull/832/changes#diff-dccfc013a62dc574ec55dd300eae0a402537fbc168a9c4068abab0bc2eeced67R54-R56 so postscript can have the environment variables from set_userscript_env_vars. Previous commit cannot parse `${PBS_NCI_STORAGE}`. I added `os.path.expandvars(postscript)` and now it works.

payu/status.py
Qian-HuiChen: Good point.
I added a [function](https://github.com/payu-org/payu/pull/832/changes#diff-fca37d36316f2b9077aa237ee539c593de2ed038ada93dad5c27153114b727c8R458) called `parse_exit_status_from_logs`, in case this job does not exist in `qstat -xf`. `update_postscript_job_file` will look for the latest scheduler from `get_job_info` first. If not found, it looks for the exit status from stdout. If both are None, `data["payu_postscript_status"]` is left unchanged. (I don't know what would be the best status code here)

payu/experiment.py
Qian-HuiChen: Removed ~

payu/subcommands/collate_cmd.py
Qian-HuiChen: Good point! I initialized `workflow` before try/except in the latest commit

payu/subcommands/collate_cmd.py
Qian-HuiChen: Yes yes, I moved this to cli.set_env_vars in the latest commit

payu/subcommands/collate_cmd.py
Qian-HuiChen: I see! I updated the submit_collate and submit_sync accordingly in the latest commit

Qian-HuiChen:
I did the following manual tests with different workflows and marked down the `payu status` behaviors:
image

aidanheerdegen:
> I did the following manual tests with different workflows and marked down the `payu status` behaviors

Thanks not only for the doing the testing, but also reporting it so succinctly in the PR. 

Out of interest, did the manual testing surface any bugs or unexpected behaviour?

Qian-HuiChen:
>Out of interest, did the manual testing surface any bugs or unexpected behaviour?

It does! A few bugs were revealed. One I recalled was the [expt.get_job_file(type=step)](https://github.com/payu-org/payu/pull/832/changes/558abe1954e8bac422ba52b16416cc69ebe57731#diff-324a5f44cb1bcbc568834715b78a4d79acd3f492cf4f3335097f521ab67f95a0R168) didn't work for `payu-collate` or `payu-run`, which leads me to build the workflow as {"step": "job_id"} and use [telemetry.get_job_file_path_with_id](https://github.com/payu-org/payu/pull/832/changes/cbefae7b011965b972f39d2e536b72a31dd3f880..300f0cd0b2d3447d49a6401d5967c47b022db7eb#diff-324a5f44cb1bcbc568834715b78a4d79acd3f492cf4f3335097f521ab67f95a0R168) instead. I fix some others along the way of testing.

payu/subcommands/collate_cmd.py
```diff
@@ -10,22 +10,38 @@
...
 parameters = {'description': 'Collate tiled output into single output files'}
 
 arguments = [args.model, args.config, args.initial, args.laboratory,
              args.dir_path, args.dry_run]
 
+def submit_collate(expt, depends_on=None):
+    """ Submit the collate job by calling runcmd.
+    Return the job id of the collate job"""
+    job_id = runcmd(
+            init_run=expt.counter,
```
aidanheerdegen:
```suggestion
def submit_collate(run_counter, depends_on=None):
    """ Submit the collate job by calling runcmd.
    Return the job id of the collate job"""
    job_id = runcmd(
            init_run=run_counter,
```
If only the `expt.counter` variable is used might be best to just pass that value in directly.

payu/subcommands/collate_cmd.py
```diff
@@ -93,9 +109,10 @@ def runcmd(model_type, config_path, init_run, lab_path, dir_path, dry_run=False)
     expt = Experiment(lab)
```
aidanheerdegen, line 114: I'm not sure this is necessary as it doesn't seem like the `expt` object is being queried here or in `runscript` where it is passed. 

I know this hasn't been changed, but just noticed it seems superfluous.

payu/subcommands/collate_cmd.py
```diff
@@ -109,20 +126,54 @@ def runscript(**run_args):
     for var in pbs_vars:
         os.environ[var] = str(pbs_vars[var])
 
+    # Get exist_workflow from environment variable (passed from runcmd via PBS)
+    exist_workflow = os.environ.get('PAYU_EXIST_WORKFLOW', 'false').lower() == 'true'
```
aidanheerdegen: Should the workflow be passed as an argument like all the other input arguments?

payu/subcommands/collate_cmd.py
```diff
@@ -109,20 +126,54 @@ def runscript(**run_args):
...
+            next_depends_on = expt.scheduler.get_job_id(short=False)
+            for step in workflow.keys():
+                if step == 'postscript':
+                    postscript_job_id = submit_postscript(expt, depends_on=next_depends_on)
+                    next_depends_on = postscript_job_id
+                elif step == 'sync':
+                    sync_job_id = submit_sync(expt, depends_on=next_depends_on)
+                    next_depends_on = sync_job_id
+                # Update the workflow dictionary with the job ID for each step
+                workflow[step] = next_depends_on
```
aidanheerdegen: This block isn't run if a workflow already exists. Is that right?

payu/subcommands/collate_cmd.py
```diff
@@ -109,20 +126,54 @@ def runscript(**run_args):
...
+        # Only when collate builds its own workflow
+        if not exist_workflow and workflow:
+            for step, job_id in workflow.items():
+                job_file_path = get_job_file_path_with_id(
+                                    archive_path=Path(expt.archive_path),
+                                    run_number=expt.counter,
+                                    job_id=job_id,
+                                    type=step
+                                )
+                job_file_path.unlink(missing_ok=True)
```
aidanheerdegen: It is necessary for the `collate` job to clean up the job files from the other dependent jobs because they will be killed by the PBS server and so never run?

payu/subcommands/postscript_cmd.py
```diff
@@ -0,0 +1,67 @@
...
+arguments = [args.model, args.config, args.laboratory, args.initial, args.dry_run]
+
+            
+def submit_postscript(expt, depends_on=None):
+    """ Submit postprocessing script if configured in config.yaml. Return the job id of the postscript job"""
+    job_id = runcmd(
+        model_type=expt.lab.model_type,
+        config_path=expt.config_path,
+        init_run=expt.counter,
+        lab_path=expt.lab.basepath,
```
aidanheerdegen:
```suggestion
def submit_postscript(model_type, config_path, run_counter, basepath, depends_on=None):
    """ Submit postprocessing script if configured in config.yaml. Return the job id of the postscript job"""
    job_id = runcmd(
        model_type=model_type,
        config_path=config_path,
        init_run=counter,
        lab_path=basepath,
```
I think it's best to pass the arguments in explicitly. It's a bit more code, but I think it's good as the caller code has visibility into what is being used in the function.

payu/subcommands/postscript_cmd.py
```diff
@@ -0,0 +1,67 @@
...
+    if not postscript:
+        print("No postscript commands found in config.yaml. Skipping postscript submission.")
+        return None
+
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+
+    envmod.setup()
+    envmod.module('load', 'pbs')
```
aidanheerdegen: Is it necessary to load `pbs` if HPCPy is submitting the job?

payu/subcommands/run_cmd.py
```diff
@@ -189,13 +195,41 @@ def runscript(**run_args):
...
+                    collate_job_id = submit_collate(expt, depends_on=next_depends_on)
+                    next_depends_on = collate_job_id
+                elif step == 'postscript':
+                    postscript_job_id = submit_postscript(expt, depends_on=next_depends_on)
+                    next_depends_on = postscript_job_id
+                elif step == 'sync':
+                    sync_job_id = submit_sync(expt, depends_on=next_depends_on)
+                    next_depends_on = sync_job_id
+                # Update the workflow dictionary with the job ID for each step
+                workflow[step] = next_depends_on
```
aidanheerdegen:
```suggestion
            for step in workflow.keys():
                # Update the workflow dictionary with the job ID for each step
                workflow[step] = fn[step](expt, depends_on=next_depends_on)
```
As the pattern is the same for all step types can store the required function in the `fn` dictionary

payu/subcommands/run_cmd.py
```diff
@@ -13,10 +13,15 @@
...
-from payu.telemetry import record_run
+from payu.telemetry import get_job_file_path_with_id, record_run
 from payu.schedulers.pbs import PBS
 import payu.errors as errors
 
+# Local imports for subcommand job submission
+from payu.subcommands.collate_cmd import submit_collate
+from payu.subcommands.postscript_cmd import submit_postscript
+from payu.subcommands.sync_cmd import submit_sync
+
```
aidanheerdegen, line 19:
```suggestion
fn = {
    'collate':    submit_collate,
    'postscript': submit_postscript,
    'sync':       submit_sync,
}
```
Populate a function lookup dict for use in `runscript`

payu/subcommands/sync_cmd.py
```diff
@@ -3,25 +3,40 @@
...
+def submit_sync(expt, depends_on=None):
+    """ Submit the sync job by calling runcmd.
+    Return the job id of the sync job"""
+    sync_config = expt.config.get('sync', {})
+    job_id = runcmd(
+        init_run=expt.counter,
+        sync_restarts = sync_config.get('restarts', False),
+        sync_ignore_last = sync_config.get('ignore_last', False),
+        depends_on=depends_on,
+    )
```
aidanheerdegen:
```suggestion
def submit_sync(run_counter, sync_restarts, ignore_last, depends_on=None):
    """ Submit the sync job by calling runcmd.
    Return the job id of the sync job"""
    sync_config = expt.config.get('sync', {})
    job_id = runcmd(
        init_run=run_counter,
        sync_restarts=sync_restarts,
        sync_ignore_last=ignore_last,
        depends_on=depends_on,
    )

```
Again I'm suggesting passing in the values explicitly.

payu/subcommands/sync_cmd.py
```diff
@@ -80,4 +102,24 @@ def runscript(**run_args):
...
+        record_run(
+            timings=expt.timings,
+            scheduler=expt.scheduler,
+            status=sync_status,
+            config=expt.config,
+            file_path=job_file_path,
+            archive_path=Path(expt.archive_path),
+            type="sync",
+            stage="exited"
+        )
```
aidanheerdegen, line 131: Very clean code!

payu/status.py
```diff
@@ -446,4 +452,69 @@ def display_expt_paths(expt_paths):
...
+    
+    Return the updated job file data
+    """
+    job_id = data.get("scheduler_job_id")
+        
+    # If both stdout and stderr exist, update the stage to exited
+    if stdout and stderr and stdout.exists() and stderr.exists():
+        data["stage"] = "exited"
+
+        # Update the scheduler info by calling qstat -xf -F json job_id
```
aidanheerdegen: This comment is PBS specific

payu/subcommands/collate_cmd.py
```diff
@@ -10,22 +10,38 @@
 from payu.experiment import Experiment
 from payu.laboratory import Laboratory
 import payu.subcommands.args as args
-from payu.telemetry import record_run
-from payu import fsops
+from payu.telemetry import record_run, get_job_file_path_with_id
+from payu.fsops import read_config
+
+from payu.subcommands.postscript_cmd import submit_postscript
+from payu.subcommands.sync_cmd import submit_sync
```
aidanheerdegen:
```suggestion
from payu.subcommands.sync_cmd import submit_sync
from payu.subcommands.run_cmd import fn  
```
Import `fn`: function mapping dict

payu/subcommands/collate_cmd.py
```diff
@@ -109,20 +126,54 @@ def runscript(**run_args):
...
+            for step in workflow.keys():
+                if step == 'postscript':
+                    postscript_job_id = submit_postscript(expt, depends_on=next_depends_on)
+                    next_depends_on = postscript_job_id
+                elif step == 'sync':
+                    sync_job_id = submit_sync(expt, depends_on=next_depends_on)
+                    next_depends_on = sync_job_id
+                # Update the workflow dictionary with the job ID for each step
+                workflow[step] = next_depends_on
+
```
aidanheerdegen:
```suggestion
            for step in workflow.keys():  
                # Update the workflow dictionary with the job ID for each step  
                workflow[step] = fn[step](expt, depends_on=next_depends_on)  

```
Need to check if importing the `fn` dict works as-is, or if a version is required that points to the functions in the correct scope.

aidanheerdegen (changes requested):
It's really coming together! I've requested changes that reduce the line count, but if you think they reduce the readability I don't mind if you don't want to adopt them.

payu/subcommands/collate_cmd.py
Qian-HuiChen: This is passed into cli.submit in the next line, and later used to define run number + write job file.

payu/subcommands/collate_cmd.py
Qian-HuiChen: Correct, if exist_workflow is True, this whole block is skipped

payu/subcommands/collate_cmd.py
aidanheerdegen: Oh dear! So it is. Sorry for the misdirect/noise.

Qian-HuiChen:
I agree with @aidanheerdegen that a workflow class would be good to manage building a workflow dictionary and submit jobs of different stages. [0ec9e52](https://github.com/payu-org/payu/pull/832/commits/0ec9e52cc6080d9499b143ae4d85101a2df83f25) is the first attempt.

However, I couldn't find a good way to clean up the follow-up jobs if the first job fails. I tried a few approaches:
- Calling self.workflow: when Workflow is initialised, the workflow is reset to None, 
- Saving the workflow in os.environ: I couldn't pass the follow-up job IDs (postscript/sync) through the environment because these IDs are not available until after the collate job has been submitted.
- Writing the workflow dictionary to the job file: when the collate job is being submitted, payu can't get the scheduler job ID of the original run job.

Currently, user can run `payu status --update` to remove the job files of killed jobs. I am not sure if a Workflow.clean_up is needed anymore.

aidanheerdegen:
> However, I couldn't find a good way to clean up the follow-up jobs if the first job fails. I tried a few approaches

The workflow could spawn a job that has a `depend:afternotok:job-id` dependency. It will only run if the job fails.

See https://ncar-hpc-docs.readthedocs.io/en/latest/pbs/job-dependencies/

That job could do whatever clean-up was necessary.

Unfortunately HPCPy hard codes `afterok` as the dependency:

https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/constants/pbs.py#L21

If `afternotok` was required it would need a code change in HPCPy to support defining the dependency type. Arguably this is a good idea anyway.

payu/workflow.py
```diff
@@ -0,0 +1,54 @@
...
+                from payu.subcommands.collate_cmd import submit_collate
+                depends_on = submit_collate(self.run_number, depends_on)
+
+            elif step == "postscript":
+                from payu.subcommands.postscript_cmd import submit_postscript
+                depends_on = submit_postscript(self.run_number, depends_on)
+
+            elif step == "sync":
+                from payu.subcommands.sync_cmd import submit_sync
+                depends_on = submit_sync(self.run_number, depends_on, self.config)
```
aidanheerdegen: Is there a way the call signature of `submit_sync` could be made to be the same as the call signature of the other `submit_*` functions? If so then maybe there doesn't have to be a `submit_` function in every subcommand? Can it just be generic and call the respective command, or make it inline in this class?

Qian-HuiChen:
@bschroeter Would it be possible to add an argument in HPCpy that supports a dependency like `depend:afternotok:job-id`?
Currently, the `depends_on` argument starts a job only if the previous job completes as `ok` ([see code](https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/constants/pbs.py#L21)). It would be great to have another argument that starts a job if the previous job fails. This would be useful for running cleanup stages.

payu/subcommands/collate_cmd.py
Qian-HuiChen: This is a good point. I think collate_cmd.py only needs to clean up "other" job files when `payu-run` called `collate` + "others", and `collate` failed. So this part should be not in the `try/except` but in `finally`

payu/subcommands/postscript_cmd.py
Qian-HuiChen: Turns out it is not necessary. It is removed now.

payu/subcommands/run_cmd.py
Qian-HuiChen: This inspired me in the latest commit. I build a function lookup dictionary in Workflow class

payu/status.py
Qian-HuiChen: Removed in one of the recent commits

payu/subcommands/collate_cmd.py
Qian-HuiChen: In previous in-person conversation, we discussed that this will lead to circular import.

payu/workflow.py
Qian-HuiChen: In [fc52342](https://github.com/payu-org/payu/pull/832/commits/fc52342c1ebae150fae2aa5f1e60c98f9cf956dc), I added `config=None`  to all `submit_*` function so they all look the same. And then I build a function lookup dictionary in Workflow class, so we can do:
```
    fn={
        "collate": submit_collate,
        "postscript": submit_postscript,
        "sync": submit_sync
        }
     for step in self.workflow_steps.keys():
            fn[step](self.run_number, depends_on, self.config)
```

aidanheerdegen: Looks good!

bschroeter:
> @bschroeter Would it be possible to add an argument in HPCpy that supports a dependency like `depend:afternotok:job-id`? Currently, the `depends_on` argument starts a job only if the previous job completes as `ok` ([see code](https://github.com/ACCESS-NRI/hpcpy/blob/40c18a204f085d09262bc10735e38d507279ab87/hpcpy/constants/pbs.py#L21)). It would be great to have another argument that starts a job if the previous job fails. This would be useful for running cleanup stages.

In short, yes. This is possible. You're right in that the current implementation works on a success-based dependency model.

I will have to take a look at the SLURM equivalents of the PBS `notok` etc. syntax. I see no issue with implementing this, though it will change the API to submit jobs.

Please feel free to raise the issue over on the HPCpy repo. I'll hash out some ideas.

Qian-HuiChen:
@bschroeter Sounds great. I opened an [issue](https://github.com/ACCESS-NRI/hpcpy/issues/73) in HPCpy

payu/workflow.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def submit_workflow(self, depends_on):
+        """Submit all later jobs in the workflow in order, passing job IDs as dependencies,
+        Update the value in workflow dictionary."""
+        fn = self.import_subcommands()
+
+        for step in self.workflow_steps.keys():
+            self.workflow_steps[step] = fn[step](self.run_number, depends_on, self.config)
+
+            depends_on = self.workflow_steps[step]
```
aidanheerdegen, line 35:
```suggestion
            # Next step depends on this step
            depends_on = self.workflow_steps[step]
```

payu/subcommands/collate_cmd.py
```diff
@@ -4,28 +4,48 @@
...
+import json
 
 # Local
 from payu import cli
 from payu.experiment import Experiment
 from payu.laboratory import Laboratory
 import payu.subcommands.args as args
 from payu.telemetry import record_run
-from payu import fsops
+from payu.fsops import read_config
```
aidanheerdegen:
```suggestion
from payu.fsops import read_config, str_to_bool
```
See below

payu/subcommands/collate_cmd.py
```diff
@@ -138,3 +167,11 @@ def runscript(**run_args):
             type="collate",
             stage="exited"
         )
+
+        # Submit follow-up jobs in the workflow, if payu-collate succeed and not called by payu-run
+        exist_workflow = os.environ.get('PAYU_EXIST_WORKFLOW', 'false').lower() == 'true'
```
aidanheerdegen:
```suggestion
        exist_workflow = str_to_bool(os.environ.get('PAYU_EXIST_WORKFLOW', 'false'))
```

payu/subcommands/collate_cmd.py
```diff
@@ -138,3 +167,11 @@ def runscript(**run_args):
...
+
+        if collate_status == 0 and not exist_workflow:
+            workflow.build_workflow()
+            workflow.workflow_steps.pop('collate', None)
```
aidanheerdegen: Best practice is not to reach inside classes and mutate them, but either provide methods to do this, or avoid the necessity to do so.

I propose that you refactor `Workflow` slightly, and when instantiating it pass in the steps that should be skipped.

See comments on workflow for some suggestions.

payu/workflow.py
aidanheerdegen, line 1: As mentioned elsewhere I'd suggest adding the ability to specify steps to skip.

The most straightforward way to do this would be to have an optional argument to `build_workflow`, e.g. `skip=None` and don't add those steps to the workflow.

If you were keen you could refactor this class, not store `config` as a member variable, and don't pass it directly to `__init__`, but instead define a `@classmethod` function that takes a `config` and returns a Workflow object.

aidanheerdegen (changes requested):
Looks good, just some suggestions ...

payu/workflow.py
Qian-HuiChen: Thanks for pointing out it is not good to pop the workflow_step outside the class. I refactor the Workflow class with a classmethod in [e23be12](https://github.com/payu-org/payu/pull/832/commits/e23be12a7f6c203895dd0bb0cfcc1d187ccd5f72). It also accepts an argument as `skip_step`.

payu/subcommands/collate_cmd.py
Qian-HuiChen: Updated in [e23be12](https://github.com/payu-org/payu/pull/832/commits/e23be12a7f6c203895dd0bb0cfcc1d187ccd5f72)

payu/workflow.py
Qian-HuiChen: Good idea to add a note. Added in [e23be12](https://github.com/payu-org/payu/pull/832/commits/e23be12a7f6c203895dd0bb0cfcc1d187ccd5f72)

payu/workflow.py
```diff
@@ -3,22 +3,25 @@
 class Workflow:
     """Class to manage the workflow of run, collate, postscript, sync jobs."""
     
-    def __init__(self, config, run_number=None):
-        self.run_number = run_number
+    def __init__(self, workflow_steps, config=None, run_number=None):
```
aidanheerdegen: Is it still necessary to pass in `config` and save it as a member variable now you have the `read_config` class method?

payu/workflow.py
```diff
@@ -3,22 +3,25 @@
...
+        """Read the config dictionary and return a dictionary of workflow steps to be executed.
+        If skip_step is provided, that step will be skipped in the workflow."""
+        workflow_steps = dict()
 
+        if config.get('collate', {}).get('enable', True) and skip_step != 'collate':
+            workflow_steps['collate'] = None
+        if config.get('postscript', None) and skip_step != 'postscript':
+            workflow_steps['postscript'] = None
+        if config.get('sync', {}).get('enable', False) and skip_step != 'sync':
+            workflow_steps['sync'] = None
```
aidanheerdegen, line 21: My brain is screaming out "PLEASE LOOP OVER THE STEPS!" but I know the logic for each step is different, but it doesn't stop my brain yelling at me.

Nevertheless, I think the `skip_step` check should be first in each `if`, e.g. 
```python
if skip_step != 'sync' and config.get('sync', {}).get('enable', False):
```
As there is no need to check any other logic if it is true.

Or you could not do those checks and just `pop` it off at the end similar to the previous logic.

aidanheerdegen (commented):
Just a couple of small points, otherwise looking great

payu/workflow.py
Qian-HuiChen: I am also wondering about this. In submit_sync:
```
def submit_sync(counter, depends_on=None, config=None):
    """ Submit the sync job by calling runcmd.
    Return the job id of the sync job"""
    sync_config = config.get('sync', {})
    try:
        job_id = runcmd(
            init_run=counter,
            sync_restarts = sync_config.get('restarts', False),
            sync_ignore_last = sync_config.get('ignore_last', False),
            depends_on=depends_on,
            )
```
it needs to read out some settings from the config.
If we don't want to pass config as a member variable, I can do `submit_workflow(self, depends_on, config)`.
What do you think?

aidanheerdegen: Ah good point. Yes I think it is better to pass in `config` to the `submit_workflow` call (assuming it is readily available).

Qian-HuiChen: Changes implemented in [34b2afc](https://github.com/payu-org/payu/pull/832/commits/34b2afc1bf6dc5d5cafedee0d87d403119cad0f1)

payu/workflow.py
Qian-HuiChen: Yeah unfortunately each step has a slightly different way to read from config.
In [34b2afc](https://github.com/payu-org/payu/pull/832/commits/34b2afc1bf6dc5d5cafedee0d87d403119cad0f1), I put the skip_step check before reading from config, so python can short circuit and go to next line early.

aidanheerdegen (dismissed):
Nice work! There was quite a bit of refactoring of old crufty code required, but I think the result is a big improvement in functionality and readability.

aidanheerdegen:
Always forgetting .. are there documentation changes required?

Qian-HuiChen:
Phew, I almost missed the documentation update! Thanks for catching this. See [38ab8eb](https://github.com/payu-org/payu/pull/832/commits/38ab8ebb0bfa6ce124251f2a5ce9cf709269669b)

While updating the documentation, I realised that `payu sync` is the recommended command, which is different from `payu collate -i K`. I added a line in `sync_cmd.py` to make sure the run number is not incremented when running `payu sync`. I did the same for `payu postscript`, even though I documented it as `payu postscript -i K`.

docs/source/usage.rst
```diff
@@ -416,6 +416,24 @@ Other experiment runs will not be harmed by this command.
...
+- postscript, 
+- sync. 
+These jobs can also be run manually if required.
+When payu submits postprocessing jobs, it adds job dependencies to ensure that they run in the correct order:: 
+
+   collate -> postscript -> sync
+
+If a postprocessing job failed, any subsequent job(s) will be terminated and removed from the queue.
+You can use ``payu status`` to check the status of all postprocessing jobs.
+To clean up postprocessing jobs that have been killed, please run ``payu status --update`` once.
```
aidanheerdegen, line 431:
```suggestion
Run ``payu status --update`` to ensure the status of killed jobs is correctly reported.
```

aidanheerdegen (commented):
Thanks for the doc updates and well done spotting the bug. I just have one last small suggestion.

docs/source/usage.rst
aidanheerdegen: "Clean up" sounds a bit alarming. Am I right that it is just the `payu status` that would otherwise be affected? Or are there are actual files that need to be delete or moved?

aidanheerdegen: Ok there is clean-up. I will remove this objection!

aidanheerdegen (approved):
WAHOO!

docs/source/usage.rst
Qian-HuiChen: The subsequent job files will still live in the file system, even though the server killed the job itself.
Running `payu status --update` will remove killed job files.

PR #833, opened 2026-08-13
After `payu sweep --hard`, `payu setup` works without `--new-uuid`

### LINKED ISSUE #827: Reset experiment UUID with hard sweep
As [reported](https://github.com/payu-org/payu/issues/825), after `payu sweep --hard` it is necessary to use the `--new-uuid` option when running `payu setup` or `payu run`.

This is new behaviour because of the [change to not default to creating a new experiment UUID by default](https://github.com/payu-org/payu/pull/795). The intention of that change was to avoid inadvertent errors when `payu` could no longer find the `archive` corresponding to an experiment UUID. 

In the case of `payu sweep --hard` there is no possibility the `archive` is "misplaced", as the user just deleted it. So to restore the previous default behaviour after `payu sweep --hard` the experiment UUID should be deleted from the `metadata.yaml`.

Qian-HuiChen:
This PR updates UUID to `HARD_SWEPT_UUID` in metadata when `payu sweep --hard` is called.
In `metadata.setup`, if `UUID == HARD_SWEPT_UUID`, payu will generate a new uuid, use branch-uuid as experiment name, instead of raising an error saying archive not found.
`payu setup` will not require a `--new-uuid` flag after `payu sweep --hard`
Closes #827

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -1136,6 +1136,9 @@ class Experiment(object):
                 print('Removing symlink {0}'.format(self.archive_sym_path))
                 os.remove(self.archive_sym_path)
 
+            # Remove UUID from metadata file
+            self.metadata.wipe_uuid()
+
         # Remove stdout/err and yaml dumps
         for f in self.output_fnames:
             if os.path.isfile(f):
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -64,6 +64,8 @@ CAN_EDIT_COMMENT: [NAME_FIELD, CONTACT_FIELD, EMAIL_FIELD, CREATED_FIELD,
                     PARENT_BRANCH_TIME_FIELD, PARENT_HASH_FIELD],
 }
 
+HARD_SWEPT_UUID = "uuid-wiped-by-sweep-hard"
+
 class MetadataWarning(Warning):
     pass
 
@@ -147,18 +149,21 @@ class Metadata:
         Note: Experiment name is the name used for the work and archive
         directories in the Laboratory.
         """
+        # Check if user run payu sweep --hard before
+        hard_swept = self.uuid == HARD_SWEPT_UUID
+
         if not self.enabled:
             # Set experiment name only - either configured or legacy name
             self.set_experiment_name()
 
-        elif self.uuid is not None and (keep_uuid or not is_new_experiment):
+        elif self.uuid is not None and hard_swept is False and (keep_uuid or not is_new_experiment):
             self.set_experiment_name(keep_uuid=keep_uuid,
                                      is_new_experiment=is_new_experiment)
         else:
             # Generate new UUID
             # In older version of payu, there is no UUID in archive name, so self.uuid is None but is_new_experiment could be False.
             # In this case, we still want to pass is_new_experiment = False.
-            self.set_new_uuid(is_new_experiment=is_new_experiment)
+            self.set_new_uuid(is_new_experiment=is_new_experiment, hard_swept=hard_swept)
             print("Generated new experiment uuid: ", self.uuid)
 
         self.archive_path = self.lab_archive_path / self.experiment_name
@@ -184,7 +189,8 @@ class Metadata:
 
     def set_experiment_name(self,
                             is_new_experiment: bool = False,
-                            keep_uuid: bool = False) -> None:
+                            keep_uuid: bool = False,
+                            hard_swept: bool = False) -> None:
         """Set experiment name - this is used for work and archive
         sub-directories in the Laboratory"""
         # Experiment name configuration - this overrides experiment name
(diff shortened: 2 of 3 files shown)


Qian-HuiChen:
@aidanheerdegen I didn't make `payu sweep --hard` remove the UUID. When `self.uuid` is `None`, payu will generate a new UUID, but still there is no archive -> the "no archive found" error remains to appear.

I also noticed that `payu status` and `payu checkout` expect to raise a "no archive found" error when the metadata is not setup. I think it might be useful to distinguish between wiped metadata and broken metadata.

Based on this thoughts, I changed `payu sweep --hard` to set the UUID to `"uuid-wiped-by-sweep-hard"`. Please let me know if I am complicating things or if there is a better idea.

payu/metadata.py
```diff
@@ -147,18 +149,21 @@ def setup(self,
         Note: Experiment name is the name used for the work and archive
         directories in the Laboratory.
         """
+        # Check if user run payu sweep --hard before
+        hard_swept = self.uuid == HARD_SWEPT_UUID
```
aidanheerdegen, line 153: I agree it is good to be explicit, so I like this approach.

payu/metadata.py
```diff
@@ -147,18 +149,21 @@ def setup(self,
...
-        elif self.uuid is not None and (keep_uuid or not is_new_experiment):
+        elif self.uuid is not None and hard_swept is False and (keep_uuid or not is_new_experiment):
             self.set_experiment_name(keep_uuid=keep_uuid,
                                      is_new_experiment=is_new_experiment)
         else:
             # Generate new UUID
             # In older version of payu, there is no UUID in archive name, so self.uuid is None but is_new_experiment could be False.
             # In this case, we still want to pass is_new_experiment = False.
-            self.set_new_uuid(is_new_experiment=is_new_experiment)
+            self.set_new_uuid(is_new_experiment=is_new_experiment, hard_swept=hard_swept)
```
aidanheerdegen:
```suggestion
            self.set_new_uuid(is_new_experiment=(is_new_experiment or hard_swept))
```
Could you do above so `hard_swept` is the equivalent of saying it is a new experiment?

payu/metadata.py
```diff
@@ -184,7 +189,8 @@ def new_experiment_name(self) -> str:
 
     def set_experiment_name(self,
                             is_new_experiment: bool = False,
-                            keep_uuid: bool = False) -> None:
+                            keep_uuid: bool = False,
+                            hard_swept: bool = False) -> None:
```
aidanheerdegen: This option isn't used in the corresponding call.

payu/experiment.py
```diff
@@ -1136,6 +1136,9 @@ def sweep(self, hard_sweep=False):
                 print('Removing symlink {0}'.format(self.archive_sym_path))
                 os.remove(self.archive_sym_path)
 
+            # Remove UUID from metadata file
+            self.metadata.wipe_uuid()
```
aidanheerdegen, line 1140: Will the updated metadata file be committed? Or just have an uncommitted modification?

Qian-HuiChen: Currently the modification of "HARD_SWEPT_UUID" is not committed. I should add it as a commit.

aidanheerdegen: Maybe? Most people will hard sweep and then want to run a new experiment, which will just overwrite the `metadata.yaml` and then write a commit. So the main reason to do it would be to have a commit that signified there had been a hard sweep. I can't decide if that is useful or not.

Qian-HuiChen:
To explain my previous [comment](https://github.com/payu-org/payu/pull/833#issuecomment-5276521962) about not wiping uuid to `None`, I attach the diagrams below.
Left diagram highlights the path taken when the UUID is `None`, before this PR.
Right diagram highlights the path taken when the UUID is marked as `HARD_SWEPT_UUID`, after this PR.
image
image

payu/experiment.py
Qian-HuiChen: I think it will be useful to leave footprints in the git history about changing the UUID to `HARD_SWEPT_UUID`. Implemented in [41fdbef](https://github.com/payu-org/payu/pull/833/commits/41fdbefe60a04198da46093a831f920e0d9c448d)

payu/metadata.py
Qian-HuiChen: Removed in [41fdbef](https://github.com/payu-org/payu/pull/833/commits/41fdbefe60a04198da46093a831f920e0d9c448d)

payu/metadata.py
Qian-HuiChen: Good idea. It makes the code cleaner. Adopted in [41fdbef](https://github.com/payu-org/payu/pull/833/commits/41fdbefe60a04198da46093a831f920e0d9c448d)

aidanheerdegen (approved):
LGTM, thanks.

This is good to merge as long as there are no documentation updates required. I don't think there are. You could add something, but not sure it helps users. I'll leave it to your discretion.

Issue #344, opened 2023-06-21
Need a schema for config.yaml


aidanheerdegen:
`payu` needs a schema for `config.yaml` so we can build tools that can parse the files correctly.

harshula:
**Notes**

https://json-schema-everywhere.github.io/yaml
https://developers.redhat.com/blog/2020/11/25/how-to-configure-yaml-schema-to-make-editing-files-easier

aidanheerdegen:
Another use-case is generating a `config.yaml` with pre-populated default values. This could be used to generate a "full" `config.yaml` artefact in the `work` directory as a better record of an experiment.

Issue #48, opened 2015-10-02
Error with mitgcm executable compiled without netcdf (mnc) support


aidanheerdegen:
Got this error when using an executable compiled without mnc support:

```python
Traceback (most recent call last):
  File "/jobfs/local/pbs/mom_priv/jobs/2567927.r-man2.SC", line 9, in <module>
    run_cmd.runscript()
  File "/apps/payu/0.5.1/lib/payu/subcommands/run_cmd.py", line 112, in runscript
    expt.setup()
  File "/apps/payu/0.5.1/lib/payu/experiment.py", line 345, in setup
    model.setup()
  File "/apps/payu/0.5.1/lib/payu/models/mitgcm.py", line 122, in setup
    f90nml.write(data_mnc_nml, data_mnc_path)
  File "/apps/payu/0.5.1/lib/f90nml/__init__.py", line 65, in write
    nml.write(nml_path, force=force)
AttributeError: 'dict' object has no attribute 'write'
```

Has f90nml changed so it doesn't support writing a vanilla dictionary? Does the dict have to be initialised as an f90nml object first?

marshallward:
Yeah, confirmed.  This is using the older interface (that I thought was still working).
I will fix this in f90nml and close this issue once it's working (and installed on Raijin)

marshallward:
Fixed in f90nml, Raijin install has been updated.

PR #835, opened 2026-08-18
Add runlog-off, repeat and reproduce-off CLI flag in payu run

### LINKED ISSUE #302: Add command line options for runlog and repeat
Options for the `runlog` and `repeat` options would make testing new configurations easier, as runs don't pollute the commit history, and `repeat` means the restart manifest doesn't get updated, which is undesirable in some cases. 

Making these command line options obviates the need to add them to the `config.yaml` and then remove them again before publishing a config.

Qian-HuiChen:
This PR adds `--runlog-off`, `--repeat` and `--reproduce-off` command-line flags to `payu run`.

### `--runlog-off`:
This experiment skips git commits. The runlog section in `config.yaml` is ignored.

### `--repeat`
This run becomes a repeat run, and `config.yaml` repeat setting is ignored.

### `--reproduce=False`
Disables manifest reproducibility checks for executable, input, and restart files. 
When `--reproduce=True` is added, all manifests will be checked for reproducibility (exe, input and restarts). 
The reproduce section in `config.yaml` is ignored.

Closes #302

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -132,9 +132,9 @@ def get_model_type(model_type, config):
 
 
 def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
-                 reproduce=False, force=False, force_prune_restarts=False,
+                 reproduce=None, force=False, force_prune_restarts=False,
                  sync_restarts=False, sync_ignore_last=False, runlog_off=False,
-                 repeat_run=False, reproduce_off=False):
+                 repeat=False):
     """Construct the environment variables used by payu for resubmissions."""
     payu_env_vars = {}
 
@@ -179,11 +179,8 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     if dir_path:
         payu_env_vars['PAYU_DIR_PATH'] = os.path.normpath(dir_path)
 
-    if reproduce:
+    if reproduce is not None:
         payu_env_vars['PAYU_REPRODUCE'] = reproduce
-
-    if reproduce_off:
-        payu_env_vars['PAYU_REPRODUCE_OFF'] = reproduce_off
 
     if force:
         payu_env_vars['PAYU_FORCE'] = force
@@ -200,8 +197,8 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     if runlog_off:
         payu_env_vars['PAYU_RUNLOG_OFF'] = runlog_off
 
-    if repeat_run:
-        payu_env_vars['PAYU_REPEAT_RUN'] = repeat_run
+    if repeat:
+        payu_env_vars['PAYU_REPEAT'] = repeat
 
     # Pass through important module related environment variables
     module_env_vars = ['MODULESHOME', 'MODULES_CMD', 'MODULEPATH', 'MODULEV']
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -133,7 +133,8 @@ def get_model_type(model_type, config):
 
 def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
                  reproduce=False, force=False, force_prune_restarts=False,
-                 sync_restarts=False, sync_ignore_last=False):
+                 sync_restarts=False, sync_ignore_last=False, runlog_off=False,
+                 repeat_run=False, reproduce_off=False):
     """Construct the environment variables used by payu for resubmissions."""
     payu_env_vars = {}
 
@@ -181,6 +182,9 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     if reproduce:
         payu_env_vars['PAYU_REPRODUCE'] = reproduce
 
+    if reproduce_off:
+        payu_env_vars['PAYU_REPRODUCE_OFF'] = reproduce_off
+
     if force:
         payu_env_vars['PAYU_FORCE'] = force
 
@@ -192,6 +196,12 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
 
     if force_prune_restarts:
         payu_env_vars['PAYU_FORCE_PRUNE_RESTARTS'] = force_prune_restarts
+
+    if runlog_off:
+        payu_env_vars['PAYU_RUNLOG_OFF'] = runlog_off
+
+    if repeat_run:
+        payu_env_vars['PAYU_REPEAT_RUN'] = repeat_run
 
     # Pass through important module related environment variables
     module_env_vars = ['MODULESHOME', 'MODULES_CMD', 'MODULEPATH', 'MODULEV']
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -100,9 +100,7 @@ class Experiment(object):
         self.postscript = self.config.get('postscript')
         # repeat prioritise CLI flag > environment variable > config.yaml
         if not repeat:
-            repeat = os.environ.get('PAYU_REPEAT', False)
-        if not repeat:
-            repeat = self.config.get('repeat', False)
+            repeat = os.environ.get('PAYU_REPEAT', False) or self.config.get('repeat', False)
         self.repeat = repeat
 
         # Configuration
@@ -163,7 +161,7 @@ class Experiment(object):
 
         # runlog_off prioritise CLI flag > environment variable > config.yaml (in runlog.py)
         if not runlog_off:
-            runlog_off = os.environ.get('PAYU_RUNLOG_OFF', False)
+            runlog_off = os.environ.get('PAYU_RUNLOG_OFF', 'False').lower() == 'true'
 
         self.runlog = Runlog(self, runlog_off)
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -35,7 +35,7 @@ from payu.fsops import make_symlink, read_config, movetree
 from payu.fsops import list_sorted_archive_dirs
 from payu.fsops import run_script_command
 from payu.fsops import needs_subprocess_shell
-from payu.fsops import get_size
+from payu.fsops import get_size, str_to_bool
 from payu.schedulers import index as scheduler_index, DEFAULT_SCHEDULER_CONFIG
 from payu.models import index as model_index
 from payu.runlog import Runlog
@@ -73,9 +73,9 @@ def timeit(time_name):
 
 
 class Experiment(object):
-    def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config_path=None, 
+    def __init__(self, lab, reproduce=None, force=False, metadata_off=False, config_path=None, 
                  is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None,
-                 runlog_off=False, repeat_run=False, reproduce_off=False):
+                 runlog_off=False, repeat=False):
         self.init_timings()
         self.lab = lab
         # Check laboratory directories are writable
@@ -98,12 +98,12 @@ class Experiment(object):
         # Payu experiment type
         self.debug = self.config.get('debug', False)
         self.postscript = self.config.get('postscript')
-        # repeat_run prioritise CLI flag > environment variable > config.yaml
-        if not repeat_run:
-            repeat_run = os.environ.get('PAYU_REPEAT_RUN', False)
-        if not repeat_run:
-            repeat_run = self.config.get('repeat', False)
-        self.repeat_run = repeat_run
+        # repeat prioritise CLI flag > environment variable > config.yaml
+        if not repeat:
+            repeat = os.environ.get('PAYU_REPEAT', False)
+        if not repeat:
+            repeat = self.config.get('repeat', False)
+        self.repeat = repeat
 
         # Configuration
         self.expand_shell_vars = True
@@ -144,16 +144,14 @@ class Experiment(object):
                                 restart_path=self.prior_restart_path,
                                 parent_info=parent_info)
 
-        if not reproduce:
+        if reproduce is None:
             # check environment for reproduce flag under PBS
-            reproduce = os.environ.get('PAYU_REPRODUCE', False)
-
-        if not reproduce_off:
-            reproduce_off = os.environ.get('PAYU_REPRODUCE_OFF', False)
+            reproduce = os.environ.get('PAYU_REPRODUCE', None)
+            reproduce = str_to_bool(reproduce) if reproduce else None
 
         # Initialize manifest
         self.manifest = Manifest(self.config.get('manifest', {}),
-                                 reproduce=reproduce, reproduce_off=reproduce_off)
+                                 reproduce=reproduce)
 
         # Miscellaneous configurations
         userscript_val = self.config.get('userscripts', {})
@@ -165,10 +163,9 @@ class Experiment(object):
 
         # runlog_off prioritise CLI flag > environment variable > config.yaml (in runlog.py)
         if not runlog_off:
-            self.runlog_off = os.environ.get('PAYU_RUNLOG_OFF', False)
-        else:
-            self.runlog_off = runlog_off
-        self.runlog = Runlog(self)
+            runlog_off = os.environ.get('PAYU_RUNLOG_OFF', False)
+
+        self.runlog = Runlog(self, runlog_off)
 
         #  This is a bit hacky
         default_payu_bin = os.path.dirname(sys.argv[0])
@@ -446,13 +443,13 @@ class Experiment(object):
         no_restarts = self.max_output_index(output_type="restart") is None
         # Check if a user restart directory is avaiable
         user_restart_dir = self.config.get('restart')
-        if (no_restarts or self.repeat_run) and user_restart_dir:
+        if (no_restarts or self.repeat) and user_restart_dir:
             if not os.path.isdir(user_restart_dir):
                 raise errors.PayuConfigError(
                     f"No restart directory found at {user_restart_dir}. "
                     "Check 'restart:' field in config.yaml."
                 )
-            if self.counter > 0 and not self.repeat_run:
+            if self.counter > 0 and not self.repeat:
                 warnings.warn(
                     "Starting run from restart directory (in config.yaml): "
                     f"{user_restart_dir}"
@@ -462,11 +459,11 @@ class Experiment(object):
             prior_restart_dir = 'restart{0:03}'.format(self.counter - 1)
             prior_restart_path = os.path.join(self.archive_path,
                                               prior_restart_dir)
-            if os.path.exists(prior_restart_path) and not self.repeat_run:
+            if os.path.exists(prior_restart_path) and not self.repeat:
                 self.prior_restart_path = prior_restart_path
             else:
                 self.prior_restart_path = None
-                if self.counter > 0 and not self.repeat_run:
+                if self.counter > 0 and not self.repeat:
                     warnings.warn(
                         "No prior restart directory found in archive "
                         "or specified in config.yaml"
@@ -1184,9 +1181,9 @@ class Experiment(object):
         for restart in restarts:
             restart_indices[restart] = int(restart.lstrip('restart'))
 
-        # TODO: Previous logic was to prune all restarts if self.repeat_run
+        # TODO: Previous logic was to prune all restarts if self.repeat
         # Still need to figure out what should happen in this case
-        if self.repeat_run:
+        if self.repeat:
             return [os.path.join(self.archive_path, restart)
                     for restart in restarts]
 
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -74,7 +74,8 @@ def timeit(time_name):
 
 class Experiment(object):
     def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config_path=None, 
-                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None):
+                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None,
+                 runlog_off=False, repeat_run=False, reproduce_off=False):
         self.init_timings()
         self.lab = lab
         # Check laboratory directories are writable
@@ -97,7 +98,12 @@ class Experiment(object):
         # Payu experiment type
         self.debug = self.config.get('debug', False)
         self.postscript = self.config.get('postscript')
-        self.repeat_run = self.config.get('repeat', False)
+        # repeat_run prioritise CLI flag > environment variable > config.yaml
+        if not repeat_run:
+            repeat_run = os.environ.get('PAYU_REPEAT_RUN', False)
+        if not repeat_run:
+            repeat_run = self.config.get('repeat', False)
+        self.repeat_run = repeat_run
 
         # Configuration
         self.expand_shell_vars = True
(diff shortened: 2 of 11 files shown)


Qian-HuiChen:
Updated according to [2db3f5b](https://github.com/payu-org/payu/pull/835/commits/2db3f5bd71fc054dfe23d6aa3d32a493de902b35).
Manual test passed:
1. {runlog: true} in `config.yaml`. `payu run` and `payu-run` add git commits, `payu-run --runlog-off` and `payu run --runlog-off` does not add a commit.
2. no repeat section in `config.yaml`. `payu run --repeat` and `payu-run --repeat` sets `repeat=True` in experiment.py
3. set manifests reproduce to {exe: true} in `config.yaml`. `payu run -r False` and `payu-run -r False` sets reproduce to {input: False, restart: False, exe: False} in manifest.py. `payu run -r` and `payu run -r True` sets reproduce to all True.

payu/subcommands/args.py
```diff
@@ -99,11 +99,21 @@
...
         'dest':     'reproduce',
         'default':  False,
-        'help':     'Only run if manifests are correct',
+        'help': 'A fallback value when manifest reproduce section is not configured in config.yaml',
     }
 }
 
 
+reproduce_off = {
+    'flags': ('--reproduce-off',),
```
aidanheerdegen:
```suggestion
    'flags': ('--no-reproduce', '--no-repro'),
```
There is no hard and fast rules, but the [GNU uses a `no-` prefix to negate options](https://www.gnu.org/prep/standards/html_node/Option-Table.html). Additionally it is quite a bit to type, so even if you decide to keep the original name I think a shorter version would be good. I can't think of a good single letter option that isn't already taken, but perhaps that is ok as this should be something that can't be accidentally turned off.

payu/subcommands/run_cmd.py
```diff
@@ -22,11 +22,22 @@
...
 
 
+def validate_reproduce_flags(reproduce, reproduce_off):
+    """
+    Validate that --reproduce and --reproduce-off are not both set.
+    """
+    if reproduce and reproduce_off:
+        raise errors.PayuConfigError(
+            "--reproduce and --reproduce-off cannot both be specified."
+        )
```
aidanheerdegen: FYI argparse [does support mutually exclusive options](https://docs.python.org/3/library/argparse.html#mutual-exclusion), but I assume this would be annoying to implement with the way CLI arguments are defined in payu. So no action required, just a comment.

payu/manifest.py
```diff
@@ -238,7 +238,7 @@ class Manifest(object):
     methods to operate on them
     """
 
-    def __init__(self, config, reproduce):
+    def __init__(self, config, reproduce, reproduce_off=False):
```
aidanheerdegen:
```suggestion
    def __init__(self, config, reproduce=None, override_config=False):
```
I suggest changing the function (and therefore name) of the argument so that it overrides the `config.yaml` but still uses the value defined in `reproduce`, as this is a more general approach.

payu/manifest.py
```diff
@@ -271,12 +271,16 @@ def __init__(self, config, reproduce):
         # Initialise manifests and reproduce flags
         self.manifests = {}
         self.previous_manifests = {}
-        reproduce_config = self.manifest_config.get('reproduce', {})
         self.reproduce = {}
```
aidanheerdegen:
```suggestion
```
Suggest deleting this if following the suggestion above, as the default value is set there.

payu/manifest.py
```diff
@@ -271,12 +271,16 @@ def __init__(self, config, reproduce):
...
+
+        # reproduce_off flag > config.yaml > reproduce flag
+        reproduce_config = self.manifest_config.get('reproduce', {})
```
aidanheerdegen, line 276:
```suggestion
        if reproduce is not None and override_config:
            # Override config.yaml reproduce settings. Will use value of reproduce 
            reproduce_config = {}
        else:
            reproduce_config = self.manifest_config.get('reproduce', {})
```
As above suggest a boolean argument that switches the order of preference of the `reproduce` argument and the config setting, so the config setting is overidden/ignored.

payu/manifest.py
```diff
@@ -271,12 +271,16 @@ def __init__(self, config, reproduce):
...
         for mf in ['input', 'restart', 'exe']:
             self.init_mf(mf)
-            self.reproduce[mf] = reproduce_config.get(mf, reproduce)
-
+            if reproduce_off:
+                self.reproduce[mf] = False
+            else:
+                self.reproduce[mf] = reproduce_config.get(mf, reproduce)
         # Make sure the manifests directory exists
```
aidanheerdegen: If agree with the previous suggestion this logic could be removed and the old single line reinstated.

Optionally the `override_config` option could be tested in this loop, just set `self.reproduce[mf] = reproduce` rather than `False`.

Arguably it's better to put it in this loop, as it is more explicit, but it does mean the logic is tested 3 times rather than once.

payu/cli.py
```diff
@@ -133,7 +133,8 @@ def get_model_type(model_type, config):
 
 def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
                  reproduce=False, force=False, force_prune_restarts=False,
-                 sync_restarts=False, sync_ignore_last=False):
+                 sync_restarts=False, sync_ignore_last=False, runlog_off=False,
+                 repeat_run=False, reproduce_off=False):
```
aidanheerdegen:
```suggestion
                 repeat_run=False, override_config=False):
``
If you decide to change the argument name, it will need to be done here also.

File: payu/cli.py [hunk 8]
```diff
@@ -181,6 +182,9 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     if reproduce:
         payu_env_vars['PAYU_REPRODUCE'] = reproduce
 
+    if reproduce_off:
+        payu_env_vars['PAYU_REPRODUCE_OFF'] = reproduce_off
+
```
aidanheerdegen:
```suggestion
    if override_config:
        payu_env_vars['PAYU_OVERRIDE_CONFIG'] = override_config

```
And here ...

payu/experiment.py
```diff
@@ -97,7 +98,12 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
         # Payu experiment type
         self.debug = self.config.get('debug', False)
         self.postscript = self.config.get('postscript')
-        self.repeat_run = self.config.get('repeat', False)
+        # repeat_run prioritise CLI flag > environment variable > config.yaml
+        if not repeat_run:
+            repeat_run = os.environ.get('PAYU_REPEAT_RUN', False)
+        if not repeat_run:
+            repeat_run = self.config.get('repeat', False)
+        self.repeat_run = repeat_run
```
```suggestion
        # repeat_run prioritise CLI flag > environment variable > config.yaml
        if not (override_config and repeat_run is not None):
            repeat_run = os.environ.get('PAYU_REPEAT_RUN', None)
            if repeat_run is None:
                repeat_run = self.config.get('repeat', False)
        self.repeat_run = repeat_run
```
Suggest use `override_config` for all CLI options that need to indicate preference over `config.yaml` settings. This also overrides environment settings, so if you want a more general option name could go for `cli_override` or similar.

payu/experiment.py
```diff
@@ -74,7 +74,8 @@ def wrapper(self, *args, **kwargs):
 
 class Experiment(object):
     def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config_path=None, 
-                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None):
+                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None,
+                 runlog_off=False, repeat_run=False, reproduce_off=False):
```
```suggestion
    def __init__(self, lab, reproduce=None, force=False, metadata_off=False, config_path=None, 
                 is_new_experiment=False, keep_uuid=False, set_template_values=False, parent_info=None,
                 runlog=None, repeat=None, override_config=False):
```

Suggest change in argument name, but also setting the default CLI arguments to `None`, so their value can be tested and only used if set. Then `override_config` can be used as the single flag for setting precedence.

Otherwise I can see a proliferation of arguments. I can see there is already a `metadata_off` argument, so maybe this idea starts to seem less appealing!

payu/experiment.py
```diff
@@ -142,9 +148,12 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
             # check environment for reproduce flag under PBS
             reproduce = os.environ.get('PAYU_REPRODUCE', False)
 
+        if not reproduce_off:
+            reproduce_off = os.environ.get('PAYU_REPRODUCE_OFF', False)
+
```
aidanheerdegen, line 149: I think GitHub won't let me make a suggestion that include the `if not reproduce:` at line 147, but this suggestions would require that to change

```suggestion
        if not (override_config and reproduce is not None):
            # check environment for reproduce flag under PBS
            reproduce = os.environ.get('PAYU_REPRODUCE', None)
            if reproduce is None:
                reproduce_off = os.environ.get('PAYU_REPRODUCE', None)

```

payu/experiment.py
```diff
@@ -142,9 +148,12 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
...
         # Initialize manifest
         self.manifest = Manifest(self.config.get('manifest', {}),
-                                 reproduce=reproduce)
+                                 reproduce=reproduce, reproduce_off=reproduce_off)
```
```suggestion
                                 reproduce=reproduce, override_config=override_config)
```

payu/experiment.py
```diff
@@ -154,6 +163,11 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
         if init_script:
             self.run_userscript(init_script, 'init')
 
+        # runlog_off prioritise CLI flag > environment variable > config.yaml (in runlog.py)
+        if not runlog_off:
+            self.runlog_off = os.environ.get('PAYU_RUNLOG_OFF', False)
+        else:
+            self.runlog_off = runlog_off
```
```suggestion
        # runlog_off prioritise CLI flag > environment variable > config.yaml (in runlog.py)
        if not (override_config and runlog is not None): 
            runlog = os.environ.get('PAYU_RUNLOG', runlog)
```
There is some inconsistency in how/when config is set in these three routines, and some of that is because they have different requirements, but I think it makes sense to try and make it as consistent as possible.

payu/experiment.py
```diff
@@ -154,6 +163,11 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
...
         self.runlog = Runlog(self)
```
```suggestion
        self.runlog = Runlog(self, runlog)
```
I suggest passing in this runlog enable boolean into the class initialisation.

payu/runlog.py
```diff
@@ -22,14 +22,20 @@ def __init__(self, expt):
...
-            self.enabled = runlog_config.pop('enable', True)
+            # Fetch and update the runlog config
+            runlog_config = self.expt.config.get('runlog', {})
+            if isinstance(runlog_config, bool):
+                self.enabled = runlog_config
+                runlog_config = {}
+            else:
+                assert isinstance(runlog_config, dict)
+                self.enabled = runlog_config.pop('enable', True)            
+
```
aidanheerdegen: I would suggest changing the function signature to:
```
 def __init__(self, expt, runlog=None, override_config=False):
```

```suggestion
        if (override_config and runlog is not None): 
            runlog_config = { 'enabled': runlog }
        else:
            # Fetch and update the runlog config
            runlog_config = self.expt.config.get('runlog', {})
            if isinstance(runlog_config, bool):
                runlog_config = { 'enabled': runlog }
        self.enabled = runlog_config.pop('enable', True)

```

Note there is no need to check the environment variable state as this is done when called in `experiment.py`.

Also removed the `assert`. Generally want to avoid `assert` in production code ([#367](https://github.com/payu-org/payu/issues/367)). Should use a schema check ([#344](https://github.com/payu-org/payu/issues/344)) for this.

payu/cli.py
```diff
@@ -193,6 +197,12 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     if force_prune_restarts:
         payu_env_vars['PAYU_FORCE_PRUNE_RESTARTS'] = force_prune_restarts
 
+    if runlog_off:
+        payu_env_vars['PAYU_RUNLOG_OFF'] = runlog_off
+
+    if repeat_run:
+        payu_env_vars['PAYU_REPEAT_RUN'] = repeat_run
```
aidanheerdegen:
```suggestion
    if runlog_off:
        payu_env_vars['PAYU_RUNLOG'] = runlog

    if repeat_run:
        payu_env_vars['PAYU_REPEAT'] = repeat
```
For consistency these would also need to be changed.

aidanheerdegen (commented):
This is not the cleanest review, and I'm not vouching for it being completely correct, because I am suggesting some larger changes to try and make the way this is done more consistent as we're using this approach for a number of options now.

So I'm submitting this now to spur some discussion if this is a good approach. And more than happy to discuss.

As I'm suggesting some larger changes I have not reviewed the tests at this point.

payu/manifest.py
aidanheerdegen: Now I'm wondering if we even need `override_config` if the options are initialised to `None`, and so when they're specified as `True` or `False` they override.

Qian-HuiChen:
In the latest commit, payu will do 
- `payu run -r (True)` -> force reproducibility check and ignore the config.yaml
- `payu run -r False` -> skip reproducibility check and ignore the config.yaml
- `payu run --runlog-off` -> turn off runlog and ignore the config.yaml
- `payu run --repeat` -> force a repeat run and ignore the config.yaml

payu/manifest.py
Qian-HuiChen: In the latest commit, payu set `reproduce` to `None` at the beginning. 
If `reproduce==True`, payu forces reproducibility check and ignore config.
If `reproduce==False`, payu skips reproducibility check and ignore config.
If `reproduce==None`, payu reads config and falls back to False.

payu/experiment.py
```diff
@@ -97,7 +98,12 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
         # Payu experiment type
         self.debug = self.config.get('debug', False)
         self.postscript = self.config.get('postscript')
-        self.repeat_run = self.config.get('repeat', False)
+        # repeat prioritise CLI flag > environment variable > config.yaml
+        if not repeat:
+            repeat = os.environ.get('PAYU_REPEAT', False)
+        if not repeat:
+            repeat = self.config.get('repeat', False)
+        self.repeat = repeat
```
aidanheerdegen, line 104:
```suggestion
        self.repeat = os.environ.get('PAYU_REPEAT', False) or self.config.get('repeat', False)
```
I believe you can collapse the logic because python short-circuits boolean logic like this, and will only evaluate the second condition if the first is false.

payu/experiment.py
```diff
@@ -138,9 +144,10 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
                                 restart_path=self.prior_restart_path,
                                 parent_info=parent_info)
 
-        if not reproduce:
+        if reproduce is None:
             # check environment for reproduce flag under PBS
-            reproduce = os.environ.get('PAYU_REPRODUCE', False)
+            reproduce = os.environ.get('PAYU_REPRODUCE', None)
+            reproduce = str_to_bool(reproduce) if reproduce else None
```
aidanheerdegen, line 148:
```suggestion
            if reproduce := os.environ.get('PAYU_REPRODUCE', None):
                reproduce = str_to_bool(reproduce)
```
It's no less code lines, one less `if` and potentially more understandable. Just a suggestion, you don't have to change it if you prefer your formulation.

payu/experiment.py
```diff
@@ -154,7 +161,11 @@ def __init__(self, lab, reproduce=False, force=False, metadata_off=False, config
         if init_script:
             self.run_userscript(init_script, 'init')
 
-        self.runlog = Runlog(self)
+        # runlog_off prioritise CLI flag > environment variable > config.yaml (in runlog.py)
+        if not runlog_off:
+            runlog_off = os.environ.get('PAYU_RUNLOG_OFF', False)
```
aidanheerdegen: Does this need `str_to_bool` similar to above?

payu/fsops.py
```diff
@@ -405,4 +405,14 @@ def get_size(work_path):
                 total_size += os.path.getsize(fp)
 
     # Convert the total size from bytes to gigabytes
-    return (int(total_size) * ureg.byte).to(ureg.gibibyte).magnitude
\ No newline at end of file
+    return (int(total_size) * ureg.byte).to(ureg.gibibyte).magnitude
+
+def str_to_bool(value):
```
aidanheerdegen, line 410:
```suggestion
def str_to_bool(value):
    """Convert a boolean-like value to `bool`. Accepts a boolean directly, or a 
    case-insensitive string in: `{"true", "1", "false", "0"}`.
    """"
```

payu/manifest.py
```diff
@@ -271,11 +271,18 @@ def __init__(self, config, reproduce):
...
             self.init_mf(mf)
-            self.reproduce[mf] = reproduce_config.get(mf, reproduce)
+            if reproduce is not None:
+                # If reproduce is True or False, set all manifests to that value
+                self.reproduce[mf] = reproduce
+
+            else:
+                # If reproduce is None, read config.yaml setting and fall back to False
+                reproduce_config = self.manifest_config.get('reproduce', {})
+                self.reproduce[mf] = reproduce_config.get(mf, False)
```
aidanheerdegen, line 281:
```suggestion
        reproduce_config = self.manifest_config.get('reproduce', {})
        for mf in ['input', 'restart', 'exe']:
            self.init_mf(mf)
            if reproduce is None:
                # If reproduce is None, use config.yaml setting and fall back to False
                self.reproduce[mf] = reproduce_config.get(mf, False)
            else:
                # If reproduce is True or False, set all manifests to that value
                self.reproduce[mf] = reproduce
```
Can move reading the `reproduce_config` outside the loop, as it's only used in the `reproduce is None` case. I'd flip the logic of the `if/else` too, as the relationship between the variable value and the outcome is clearer.

test/test_cli.py
```diff
@@ -193,9 +199,36 @@ def test_parse_run(parser):
     assert args.pop('force_prune_restarts') is True
     assert args.pop('is_new_experiment') is False
     assert args.pop('dry_run') is False
+    assert args.pop('runlog_off') is False
+    assert args.pop('repeat') is False
 
     assert len(args) == 0
 
+def test_parse_run_reproduce_arg(parser):
+    """ Test that --reproduce argument can accpet True or False values. """
```
aidanheerdegen:
```suggestion
    """ Test that --reproduce argument can accept True or False values. """
```
Typo

test/test_setup.py
```diff
@@ -470,7 +470,7 @@ def test_null_userscripts():
 
 
 @patch('payu.git_utils.get_git_repository')
-def test_setup_runlog_enabled_not_git(mock_git_repo):
+def test_setup_runlog_offd_not_git(mock_git_repo):
```
aidanheerdegen: Is it supposed to be `offd`?

test/test_cli.py
Qian-HuiChen: This is my evidence for human-work, instead of AI-coded 😄

test/test_setup.py
Qian-HuiChen: Sorry, this file should not be touched. I guess it was accidentally touched when I tried to replace `runlog_enable` with `runlog_off` in another file.
Reverted in [2a15634](https://github.com/payu-org/payu/pull/835/commits/2a156349a10c245d21364694682011d94c872517)

test/test_cli.py
Qian-HuiChen: Corrected in [2a15634](https://github.com/payu-org/payu/pull/835/commits/2a156349a10c245d21364694682011d94c872517)

payu/manifest.py
Qian-HuiChen: It does make the code easier to read. Adapted in [2a15634](https://github.com/payu-org/payu/pull/835/commits/2a156349a10c245d21364694682011d94c872517)

payu/fsops.py
Qian-HuiChen: Added in [2a15634](https://github.com/payu-org/payu/pull/835/commits/2a156349a10c245d21364694682011d94c872517)

payu/experiment.py
Qian-HuiChen: I changed it to `runlog_off = os.environ.get('PAYU_RUNLOG_OFF', 'False').lower() == 'true'` in the latest commit. Unlike `reproduce`, `runlog_off` only have True/False, it won't be None. I can use `str_to_bool` but I am thinking the `=='true'` way will simpler

payu/experiment.py
Qian-HuiChen: I didn't change to be this way, because if `reproduce` is not None, payu should not read the environment variables. This will happen if user runs `payu-run -r (True/False)`

payu/experiment.py
Qian-HuiChen: Ohhhh I didn't know that. Good idea. Updated in the latest commit. Manual tests of `payu run --repeat` and `payu-run --repeat` passed.

payu/experiment.py
aidanheerdegen: Good point. Sorry for the misleading suggestion.

payu/experiment.py
```diff
@@ -1170,9 +1179,9 @@ def get_restarts_to_prune(self,
         for restart in restarts:
             restart_indices[restart] = int(restart.lstrip('restart'))
 
-        # TODO: Previous logic was to prune all restarts if self.repeat_run
+        # TODO: Previous logic was to prune all restarts if self.repeat
```
aidanheerdegen, line 1182: Is there an issue for this TODO?

aidanheerdegen (dismissed):
Looks good, thanks.

Only question remaining: is there a way to test the logic that has just been inserted? The cli has tests, but the effect of those options aren't tested.

If it's too difficult to do I understand, but if there could be some tests to confirm the various override logic I think it would be worth doing.

payu/experiment.py
Qian-HuiChen: This comment was last changed in 2023, it was initially:
```
# TODO: Previous logic was to prune all restarts if
# self.repeat_run - is that still the case?
```
and updated to 
```
# TODO: Previous logic was to prune all restarts if self.repeat_run
# Still need to figure out what should happen in this case
```
I tested it just then with `payu run --repeat` and `{restart_freq:5}`. Turns out payu prunes (or say removes) every single `restart*` directory. The `{restart_freq:5}` wasn't used at all. Is this an expected behavior?

aidanheerdegen: If repeat is turned on in the middle of a run does it delete all the restarts?

Qian-HuiChen: How can I turn the repeat on in the middle of a run?

aidanheerdegen: Run for a couple of cycles (so there are multiple `restartXXX` directories) and then `payu run --repeat`?

Qian-HuiChen: Before the run, I had restart045, restart046, and restart047. 
Then I called `payu run --repeat` to run #48 in debug mode. I saw that restart045, restart046, restart047, and restart048 were removed one by one.

Qian-HuiChen:
> is there a way to test the logic that has just been inserted? The cli has tests, but the effect of those options aren't tested.

Since the `tes_cli.py` covers that the arguments are parsed correctly, I reckon I should test if `Experiment` is initialized correctly and that the relevant attributes are set as expected. Please see the unit tests added in [c8030c9](https://github.com/payu-org/payu/pull/835/commits/c8030c94fc6477be10a7ba9792d879b98cf6a75f)

payu/experiment.py
Qian-HuiChen: I open an issue #845 for this and update the TODO note in [a89951b](https://github.com/payu-org/payu/pull/835/commits/a89951b0369f58459419260134b91bcecedc87d3)

aidanheerdegen (approved):
Looks great, thanks!

PR #848, opened 2026-08-28
Check for pre-existing restarts if repeat is True.

### LINKED ISSUE #845: Error should be raised if `repeat: True` and restart directory exists before run
Currently, payu removes all restart directories when `repeat: True`. For example, before the run, I had restart045, restart046, and restart047. 
Then I called `payu run --repeat` to run # 48 in debug mode. I saw that restart045, restart046, restart047, and restart048 were removed one by one. 

This behavior was based on the assumption that a repeat run always starts from the initial conditions, and therefore removing restart directories prevents payu from accidentally using a restart.

Payu should check for existing restart directories at the setup stage, and raise an error if `repeat: True` and any restart directory already exists before the run starts. This would avoid silently deleting existing restart directories.

Qian-HuiChen:
Previously, payu removes all existing restarts if `repeat: True`. Even though this is under the assumption that a repeat run always starts from initial condition, this behavior might be dangerous.

This PR adds a check on existing restarts when `repeat: True`, and raise an error if restarts exist.

Closes #845

### DIFF
--- a/payu/experiment.py
+++ b/payu/experiment.py
@@ -128,10 +128,6 @@ class Experiment(object):
 
         self.set_output_paths()
 
-        # If repeat is True, check for pre-existing restarts
-        if self.repeat:
-            self.check_restart_exists()
-
         if parent_info is None:
             parent_info = {}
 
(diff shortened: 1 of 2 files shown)


payu/experiment.py
```diff
@@ -470,6 +474,23 @@ def set_output_paths(self):
         for model in self.models:
             model.set_model_output_paths()
 
+    def check_restart_exists(self):
```
aidanheerdegen: Is it necessary to define a separate function? 

I think the logic can be added in `set_output_paths` below the setting of the `no_restarts` variable:

https://github.com/payu-org/payu/blob/master/payu/experiment.py#L441

And throw a config error `if not no_restarts and self.repeat`.

Qian-HuiChen: Sounds good. I put the check into set_output_paths in [0af261b](https://github.com/payu-org/payu/pull/848/commits/0af261bc0cd3b513962361ed257b3b07c721e078). This makes the code look much cleaner.

aidanheerdegen (approved):
LGTM!

PR #849, opened 2026-08-31
Write minimal provenance information to restarts

### LINKED ISSUE #824: Add minimal provenance information to restart directories
### Background

It is common to branch from other experiments, e.g. branch from control experiment, apply perturbations, and compare the perturbed experiment with the control to see the effect of the perturbation. There is a parent->child relationship between the control experiment and perturbation experiment..

To do this the child experiment typically `payu clone`s the parent experiment from some point in the run (not the end!) and points to the associated restart directory.

It is desirable for `payu` to be able to gather important provenance information from the parent experiment:
- experiment UUID
- git hash at branch point
- experiment time at branch point
- current run number at branch point

This information is only reliably available if the restart directory is a sub-directory of an experiment archive containing `metadata.yaml` and other log files. When restart directories are removed from their original context, and possible renamed, there is no longer reliable machine actionable access to this important provenance data.
 
### Proposal

Add a snippet of metadata to every restart directory containing information required for proper provenance., e.g.

```yaml
experiment_uuid: 0ca32cc9-481a-4a2e-b484-1e1da8b1716c
experiment_time: 1900-10-01T00:00:00
payu_run_id: a19020936dd63c0c1016684dc55bd165e9b3cb96
payu_current_run: 120
```

This is a duplication of information that is available elsewhere in the archive, and is very much a "safety-first" approach to attach relevant metadata where possible.

This issue was first proposed at https://github.com/payu-org/payu/issues/510#issuecomment-5127033262

Qian-HuiChen:
This PR adds minimal provenance information to restart directory as `restart_metadata.yaml`:
```
experiment_uuid: xxx-xxx
name: <cntrl-branch-uuid>
payu_run_id: <git_commit_hash>
payu_current_run: 37
experiment_time: '1903-12-01T00:00:00'
parent_experiment: <parent_experiment_uuid_if_exists_in_metadata>
```
Closes #824

### DIFF
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -14,6 +14,7 @@ import os
 import uuid
 import warnings
 from datetime import datetime
+import cftime
 from pathlib import Path
 from typing import Optional, Union
 
@@ -51,6 +52,7 @@ SCHEMA_VERSION = "1-0-4"
 SCHEMA_COMMIT_HASH = "cff183437134592723b09af6620e5cb190abeb22" 
 SCHEMA_URL = f"https://raw.githubusercontent.com/ACCESS-NRI/schema/{SCHEMA_COMMIT_HASH}/au.org.access-nri/model/output/experiment-metadata/{SCHEMA_VERSION}.json"
 placeholder_text = "__REPLACE_ME__"
+missing_text = "__DISABLED_OR_NOT_IMPLEMENTED__"
 
 no_archive_msg = """
 Payu needs to generate an experiment UUID and create a new archive directory.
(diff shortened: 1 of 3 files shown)


payu/metadata.py
```diff
@@ -449,6 +454,41 @@ def update_parent_info(self,
...
 
+    def write_restart_provenance(self, 
+                                restart_path: Path, 
+                                run_id: str,
+                                run_number: int,
+                                cur_expt_time: datetime,) -> None:
+        """Write provenance information to the restart directory"""
+        if not self.enabled:
+            return
+
```
aidanheerdegen: I'm not sure this should be completely skipped if `self.enabled = False`. There is still useful information that can be set. I also think there is value be explicit in what is being missed in this case too.

So I would recommend still writing the file but setting fields that can't be determined to a value that makes this clear, or write a commented version of the value. Is it only the `uuid` fields that are't available?

payu/metadata.py
```diff
@@ -449,6 +454,41 @@ def update_parent_info(self,
...
+        restart_metadata = CommentedMap()
+
+        # Add experiment UUID, name, run ID, and run number to the restart metadata
+        restart_metadata[UUID_FIELD] = self.uuid
+        restart_metadata[NAME_FIELD] = metadata.get(NAME_FIELD, None)
+        restart_metadata[RUN_ID_FIELD] = run_id
+        restart_metadata[RUN_NUMBER_FIELD] = run_number
+
+        # Add current experiment time to the restart metadata
+        restart_metadata[EXPERIMENT_TIME_FIELD] = cur_expt_time.strftime('%Y-%m-%dT%H:%M:%S')
```
aidanheerdegen: Not all models can return an experiment time. For example MITgcm. Given these models will throw a `NotImplemented`Exception this should be in a `try/except` block and that exception caught and `pass`ed.

Qian-HuiChen: Good catch. For model that are not implemented, the `self.get_model_restart_datetimes()` returns `{}`. So I updated the code to:
```
self.metadata.write_restart_provenance(
            Path(self.restart_path), 
            self.run_id, 
            self.counter,
            model_restart_datetimes.get("model_finish_time", None),
        )
```

payu/metadata.py
Qian-HuiChen: Thanks for pointing this out.
There are cases where
- runlog is disabled, run_id is None
- metadata is disabled, uuid is None.

In the latest commit, I use a function `add_restart_field` to insert the value as `__DISABLED_OR_NOT_IMPLEMENTED__` when uuid/run_id/model_finish_time s is None.

Please note that the `parent_experiment` field will not be added as an entry point if it does not exist in the original metadata.

aidanheerdegen (approved):
Nice clean implementation. Thanks!

PR #851, opened 2026-09-01
Fix handling of init_run=0 in collate workflow

### LINKED ISSUE #850: Collate job mishandles `init_run` data type
I did a test run today and found the collation job failed with the following error,

```
FileNotFoundError: [Errno 2] No such file or directory: '/scratch/tm70/ml0072/access-om3/archive/access-om3-configs-780-dev-MC_8km_jra_ryf-2-995faa67/output001'
```

However the `archive` only contains `output000`. The problem appears to be here,

https://github.com/payu-org/payu/blob/b6962f8d005a75a4f1158e9a2cf53d1003e465d0/payu/cli.py#L166-L169

where `init_run` is now being passed internally as an integer rather than the string received from the clii hence this block is skipped and run 0 is ignored.

I'll provide a fix for this now.

minghangli-uni:
closes #850 

Now this changes the checks to test against `None` preserving 0 as a valid run number.

### DIFF
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -165,7 +165,8 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     # Set the run counters
     if init_run is not None:
         init_run = int(init_run)
-        assert init_run >= 0
+        if init_run < 0:
+            raise errors.PayuRuntimeError("Run number is negative.")
         payu_env_vars['PAYU_CURRENT_RUN'] = init_run
 
     if n_runs:
--- a/payu/cli.py
+++ b/payu/cli.py
@@ -163,7 +163,7 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     payu_env_vars['PAYU_PATH'] = payu_binpath
 
     # Set the run counters
-    if init_run:
+    if init_run is not None:
         init_run = int(init_run)
         assert init_run >= 0
         payu_env_vars['PAYU_CURRENT_RUN'] = init_run
--- a/payu/subcommands/collate_cmd.py
+++ b/payu/subcommands/collate_cmd.py
@@ -114,7 +114,7 @@ def runcmd(model_type=None, config_path=None, init_run=None,
 
     # Submit the collation job and write queue job file
     job_id = cli.submit_job('payu-collate', pbs_config, pbs_vars, expt=expt, 
-                current_run = int(init_run) if init_run else None, type='collate', 
+                current_run = int(init_run) if init_run is not None else None, type='collate',
                 dry_run=dry_run, depends_on=depends_on)
     return job_id
     
--- a/test/test_subcommands.py
+++ b/test/test_subcommands.py
@@ -1,0 +1,31 @@
+from unittest.mock import MagicMock
+import pytest
+from payu.subcommands import collate_cmd, postscript_cmd, sync_cmd
+
+
+@pytest.mark.parametrize(
+    "command_mod, config",
+    [
+        (collate_cmd, {"collate": {}}),
+        (postscript_cmd, {"postscript": "postscript.sh"}),
+        (sync_cmd, {"sync": {}}),
+    ],
+)
+def test_runcmd_preserves_zero_init_run(monkeypatch, command_mod, config):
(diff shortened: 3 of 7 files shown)


aidanheerdegen:
Thanks @minghangli-uni. Can you add a test case that reproduces this case? It would be good to have this in the test suite so we don't introduce a regression.

payu/subcommands/collate_cmd.py
```diff
@@ -114,7 +114,7 @@ def runcmd(model_type=None, config_path=None, init_run=None,
 
     # Submit the collation job and write queue job file
     job_id = cli.submit_job('payu-collate', pbs_config, pbs_vars, expt=expt, 
-                current_run = int(init_run) if init_run else None, type='collate',
```
Qian-HuiChen, line 117: The same logic was used in sync_cmd.py and postscript_cmd.py. Could you help me update those two as well?

payu/cli.py
```diff
@@ -163,7 +163,7 @@ def set_env_vars(init_run=None, n_runs=None, lab_path=None, dir_path=None,
     payu_env_vars['PAYU_PATH'] = payu_binpath
 
     # Set the run counters
-    if init_run:
+    if init_run is not None:
         init_run = int(init_run)
         assert init_run >= 0
         payu_env_vars['PAYU_CURRENT_RUN'] = init_run
```
Qian-HuiChen: I have been trying to fish these random "assert" in payu and change them to error raising. Can you please help me change this one since it is in sight.
```suggestion
        if init_run < 0:
            raise errors.PayuRuntimeError("Run number is negative.")
        payu_env_vars['PAYU_CURRENT_RUN'] = init_run
```

minghangli-uni: I think it would be good to address the others in a separate PR rather than expanding the scope of this one :)

minghangli-uni:
It seems we dont have a test script to house subcommands. Do you mind I create one and put some related tests there?

Qian-HuiChen:
That sounds good. Thanks! A few of the subcommands will submit a PBS job so I don't know what is the best way to write unit tests for them

minghangli-uni:
Hi @Qian-HuiChen I think this is ready now

test/test_subcommands.py
```diff
@@ -0,0 +1,31 @@
...
+    monkeypatch.setattr(
+        command_mod,
+        "Experiment",
+        MagicMock(return_value=experiment),
+    )
+    monkeypatch.setattr(command_mod.cli, "set_env_vars", MagicMock(return_value={}))
+    submit_job = MagicMock(return_value="12")
+    monkeypatch.setattr(command_mod.cli, "submit_job", submit_job)
+
+    assert command_mod.runcmd(init_run=0) == "12"
```
Qian-HuiChen: I was a bit confused when I first saw the "12".
```suggestion
    submit_job = MagicMock(return_value="1000-gadi.pbs")
    monkeypatch.setattr(command_mod.cli, "submit_job", submit_job)

    assert command_mod.runcmd(init_run=0) == "1000-gadi.pbs"
```

Qian-HuiChen (commented):
I couldn’t reproduce the error message before this PR. Could you please provide more details about how you get to the error?

Good news is: I carried out run # 0 in both OM2 and OM3 **with** the changes from this PR, and collation succeeded in both cases. I can also see `PAYU_CURRENT_RUN=0` in the collation PBS job.

I only have one small suggestion regarding the test: I assume the `"12"` is intended to mock a PBS job ID?.

Qian-HuiChen:
Could you please help me change the first line to `if current_counter is not None`:
https://github.com/payu-org/payu/blob/e2bb3b0c1be89acabe4d5692055aa0168b08bb80/payu/experiment.py#L271-L274
Edit:
This might not change anything because `current_counter = os.environ.get('PAYU_CURRENT_RUN')`, so `current_counter` will be a string "0", and python think "0" as `True`.

minghangli-uni:
@Qian-HuiChen back to you!

Qian-HuiChen (approved):
APPROVED!

I found that with the changes, the `restart000` are not collated (👍 ), but without this PR, the `restart000` are collated (this should not happen!).

PR #854, opened 2026-09-02
Add validation of parent uuid and check against the designated restart.

### LINKED ISSUE #782: `payu checkout -p PARENT_UUID` should check if provided UUID is used by an existing experiment
Currently payu supports users to set parent experiment UUID through a command:
```
payu checkout -b new_branch source_branch -p PARENT_UUID
```
As a result, the `metadata.yaml` records a field `parent_experiment: PARENT_UUID`.

However, there lacks a check whether the provided PARENT_UUID is correct and used by an existing experiment. 
I tested it with 
```
payu checkout -b new_branch source_branch -p INVALID_UUID
```
and the `metadata.yaml` in the new branch shows `parent_experiment: INVALID_UUID `.

Ref [comment](https://github.com/payu-org/payu/pull/759#pullrequestreview-4540610733).

Qian-HuiChen:
This PR adds validation of parent experiment uuid before adding it to metadata.
1. Validate it is in UUID4 format. Raise an error if not.
2. Validate that the user-provided parent experiment UUID (`-p` flag) matches the UUID recorded in the restart metadata:
    - read `experiment_uuid` from `restart_metadata.yaml` if it exists.
    - Otherwise, check `metadata.yaml` in the parent directory of the restart (assuming the restart is still in the archive).
3. For UUID_SWEPT_BY_HARD_SWEEP, skip adding to metadata.

Closes #782

### DIFF
--- a/payu/metadata.py
+++ b/payu/metadata.py
@@ -380,11 +380,11 @@ class Metadata:
         model_name = self.metadata_config.get('model', default_model_name)
         return model_name
 
-    def get_parent_experiment(self, prior_restart_path: Path) -> None:
+    def get_parent_experiment(self, prior_restart_path: Path) -> str:
         """Searches UUID in the metadata in the parent directory that
         contains the restart"""
         if prior_restart_path is None:
-            return
+            return None
 
         # Resolve to absolute path
         prior_restart_path = prior_restart_path.resolve()
@@ -396,7 +396,7 @@ class Metadata:
             # Check for metadata file in the archive of the restart directory
             metadata_filepath = Path(prior_restart_path).parent / METADATA_FILENAME
             if not metadata_filepath.exists():
-                return
+                return None
 
         # Read metadata file
         parent_metadata = YAML().load(metadata_filepath)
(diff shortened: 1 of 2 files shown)


Qian-HuiChen:
I did some manual testing, and all the tests act as expected:

1. Run `payu checkout -b` with the `-p` flag and provide a non-UUID4 format -> errors out
2. Run `payu checkout -b` with the `-p` flag and provide a parent UUID mismatch the UUID in the `metadata.yaml` file (in the archive of the restart directory). The restart is provided via `-r`. Also test through `config.yaml`. All errors out. 
3. Similar to (2) but provide a matching parent uuid -> parent_experiment is added to metadata.
4. Move the restart outside the archive and a `restart_metadata.yaml` file exists. Test both a matching and a non-matching UUID provided via the `-p` flag -> behave as expected

payu/metadata.py
```diff
@@ -430,13 +433,18 @@ def update_parent_info(self,
...
+        if parent_experiment_from_restart:
             parent_experiment = self.get_parent_experiment(restart_path)
-        if parent_experiment and parent_experiment != self.uuid:
-            metadata[PARENT_UUID_FIELD] = parent_experiment
-        else:
-            # Remove parent_experiment if entry exists in metadata
+
+        if parent_experiment in (None, HARD_SWEPT_UUID, self.uuid):
+            # Remove parent_experiment entry if it is None/HARD_SWEPT_UUID/current_UUID
             metadata.pop(PARENT_UUID_FIELD, None)
```
aidanheerdegen: Do we need add the special case where `UUID_FIELD == "__DISABLED_OR_NOT_IMPLEMENTED__"` (https://github.com/payu-org/payu/blob/master/payu/metadata.py#L55)

aidanheerdegen (changes requested):
The logic currently is:

1. Get parent_UUID from restart dir or restart parent dir
2. Use parent_UUID if passed in
3. Validate parent_UUID against known "bad" values which are ignored
4. Check valid UUID value
5. If parent_UUID passed in Check UUID matches restart dir UUID

Seems we always want to get the restart dir UUID, either to validate a passed in parent_UUID, or to use as a parent_ID otherwise.

I'd suggest inverting the logic a bit.

Move all validation to `validate_parent_uuid`, return False if it matches the "bad" but not error values, throw an exception as now, otherwise return True.

Then steps are something like:

```python
restart_uuid = self.get_parent_experiment(restart_path)
if parent_experiment:
	if parent_experiment != restart_uuid:
		raise errors.PayuBranchError(
			f"The parent experiment UUID '{parent_experiment}' does not match "
			f"the UUID of the given restart directory '{restart_path}': '{restart_uuid}'. \n"
			f"Suggest fix: leave out the `-p` option to use the UUID of the restarts.")
else:
	parent_experiment = restart_uuid

if self.validate_parent_uuid(parent_experiment):
	metadata[PARENT_UUID_FIELD] = parent_experiment
else:
	# Deal with the consequences, throw an exception?
```

Does this capture the logic flow correctly?

payu/metadata.py
Qian-HuiChen: Good catch! I should definitely filter this out. Added in [9af8b91](https://github.com/payu-org/payu/pull/854/commits/9af8b91360b8787452083b520672fb9c281606da)

test/test_metadata.py
```diff
@@ -751,4 +778,96 @@ def test_add_restart_field(value, expected_value):
...
+        # Test with an invalid UUID format
+        ("mock_parent_uuid", True, None),
+        # Test with an hash-like string
+        ("b6962f8d005a75a4f1158e9a2cf53d1003e465d0", True, None),
+        # Test with emtpy string
+        ("", True, None),
+        # Test with None
+        (None, False, False),
+        # Test with HARD_SWEPT_UUID
+        (HARD_SWEPT_UUID, False, False)
```
aidanheerdegen: Just noting that `missing_text` is not checked

aidanheerdegen (dismissed):
LGTM. One comment about the testing coverage.

I'm a bit concerned that users might not know where a bad UUID is being found, but I also don't want to overwhelm them with status info about parent UUID discovery ...

test/test_metadata.py
Qian-HuiChen: Added in [bd90607](https://github.com/payu-org/payu/pull/854/commits/bd90607138279212507ccd5cae816fb8f4efb5d6)

aidanheerdegen (approved):
Looks good, and loving the testing coverage!

PR #857, opened 2026-09-06
Handle empty-string timestamp, as well as other invalid string and None.

### LINKED ISSUE #791: timestamp is an empty string in pbsnodes cache
Payu run fails in CI tests because timestamp is an empty string in pbsnodes cache file. This could be due to network failure.

Payu should have a check and re-generate the cache file if timestamp is `""`.
Currently payu only rerun pbsnodes if the file is older than 7 days.

Example [failed run](https://github.com/ACCESS-NRI/access-om3-configs/runs/83985560107).

Qian-HuiChen:
This PR addresses the issue of empty-string timestamp. It also handles other invalid strings and `None` as timestamp.

Closes #791 

Unit tests: added `""` and `None` as timestamp.

Manual tests:
- On a newly cloned OM2 configuration, ran `payu run -n 2` successfully.
- Changed the timestamp to `""` in the cached PBS file and ran `payu run`.
     - The timestamp was updated.
     - The cached PBS file was updated to include current job information.

### DIFF
--- a/payu/schedulers/pbs.py
+++ b/payu/schedulers/pbs.py
@@ -59,7 +59,15 @@ def check_pbsnode_file(pbsnodes_json_path):
             # Check if the file is recent (modified within the last 7 days)
             with pbsnodes_json_path.open() as f:
                 pbsnodes_json = json.load(f)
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM

Issue #510, opened 2024-09-13
Embed experiment identifiers in model outputs


aidanheerdegen:
Embedding experiment ID and run commit hashes into model output diagnostics is essential for experiment provenance: it establishes a link between the outputs of an experiment and all the provenance data of the experiment. It means consumers of the data, regardless of where they find it, have the possibility of finding this essential information.

These identifying hashes then have the ability to become [persistent identifiers (PIDs)](https://en.wikipedia.org/wiki/Persistent_identifier) once there is a service to resolve them and expose the related metadata to users. Such a service doesn't exist ... yet. But embedding this information is a necessary precursor.

## Proposal

1. Use experimentID+`git` commit hash as a unique identifier (`exptrunID`?) for each run of a model, where an experiment constitutes a number of such consecutive runs.
2. Embed `exptrunID` as a metadata field in all model output diagnostics, e.g. global netCDF attribute.
3. Where possible add `exptrunID` as an configuration input to the model so the metadata is added when the diagnostic is written. If this isn't possible add metadata after the run has completed.

## Implementation

```
exptrunID = experimentID.gitcommithash
```

Where possible the `exptrunID` should be added as a model configuration input option and written directly into the model outputs. This has two benefits:
1. It is model output format agnostic: as long as the model can write the metadata into a header in the output diagnostic file it doesn't matter what the format is
2. Reduces the number of post-processing steps. From a provenance point of view every transformation is a step that should be captured in the provenance chain, so it adds unnecessary complexity and ambiguity.
This may require code changes in the models themselves. This doesn't have to happen immediately, in the first case post-processing could be utilised until the code supported direct metadata injection. This would be tricky to manage, as it would be model version dependent.

Each model should take care of adding this metadata to the model diagnostic outputs. This means the `model` class should have a stub method `add_output_metadata` that is either not implemented, or has some useful default like adding global attribute to netCDF files.

`add_output_metadata` should be called at `setup` and `archive` stages so that `exptrunID` can be added either before a run, or after it has completed. The method needs to have logic to decide if it runs at `setup` or `archive`. If there isn't a better way, like some call-graph inspection, then the stage should be passed to the method.

## Notes

1. FMS (the GFDL coupler infrastructure used by MOM models) has an [`mpp_write_meta` routine](https://github.com/ACCESS-NRI/FMS/blob/38c7dbef9424b8b263820c7e189a4458f049db78/mpp/mpp_io.F90#L765-L772) for MOM5

[...]

PR #859, opened 2026-09-07
Use addmeta to inject metadata into model diagnostic outputs

### LINKED ISSUE #577: Modifying metadata in model output files
In some cases there are modifications to metadata for model output files that is required.

Best case would be for the model to include this metadata, but it can be difficult/impossible to enable this.

It would be good if there was a standard way to modify metadata, with a standard way to configure it.

One option is to resuscitate the [addmeta](https://github.com/coecms/addmeta) project. Either call this in a `userscript`, or incorporate this functionality directly in `payu`.

aidanheerdegen:
Create an addmeta subcommand and related functions in experiment.py and model.py to call the addmeta tool to add 
metadata to model diagnostic output files.

The design Is focussed on providing a default configuration that should work "out of the box" without requiring user modification, but with enough customisation to allow this to be overridden for a few supported use cases.

There are 3 levels of options:

- **default**: should be sufficient for most users: adding the `metadata.yaml` and `env.yaml` as datasources and injecting the UUID and git run hash as global metadata attributes.

- **global**:  configuration in the `config.yaml` to allow setting of "global'" options like `enable`, `verbose` and `update_history`.

- **model**: each model (or sub-model) queries the top level configuration and then adds model specific configuration before running addmeta. The model specific configuration is typically `files` (a glob pattern that returns all the netcdf files in the model archive folder), `fnregex` to use filename patterns to populate metadata and model specific `metadata.yaml` files that specify values like `realm`.

The model configuration is specified in `metadata.cmd` files: text files with `addmeta` command line options, one per line. This is a way to specify command line options for `addmeta` that avoids directly calling the program.

The default, global and model level configurations are combined, with last taking precedence.

Closes #577 and #510

### DIFF
--- a/payu/addmeta.py
+++ b/payu/addmeta.py
@@ -1,0 +1,140 @@
+"""Tooling to add metata to model output directories using the
+addmeta tool
+
+:copyright: Copyright 2011 Marshall Ward, see AUTHORS for details.
+:license: Apache License, Version 2.0, see LICENSE for details.
+"""
+"""
+do
+    addmeta \
+        -v -s \
+        -d metadata.yaml \
+        -d $PAYU_CURRENT_OUTPUT_DIR/env.yaml \
+        -m scripts/post-processing/addmeta/dataspec.yaml \
+        -m scripts/post-processing/addmeta/${submodel}.yaml \
+        --fnregex='access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc' \
+        $PAYU_CURRENT_OUTPUT_DIR/${submodel}/*.nc
+"""
+
+"""
+addmeta:
+    enable: true                               # default
+    verbose: true                              # default
+    update-history: false                      # default
+    data:
+        id: "{{ metadata.experiment_id }}"     # default
+        run_id: "{{ env.PAYU_RUN_ID }}"        # default
+    metafiles:
+        - metadata.yaml                        # default
+        - PATH_TO_OUTPUT_DIR/env.yaml          # default
+        - addmeta/dataspec.yaml
+    datafiles:
+        - more_metadata.yaml
+    fnregex: 'access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc'
+    submodel:
+        ocean:
+            metafiles:
+                - addmeta/ocean.yaml
+            fnregex: 'access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc'
+        ice:
+            metafiles:
+                - addmeta/ice.yaml
+        atmosphere:
+            metafiles:
+                - addmeta/atmosphere.yaml
+"""
+
+"""
+    # Call signature function to add metadata to model output directories
+    find_and_add_meta(
+        args.files,
+        combine_meta(metafiles),
+        kwdata,
+        args.fnregex,
+        sort_attrs=args.sort,
+        history=history,
+        verbose=verbose,
+    )
+"""
+
+
+from collections import defaultdict
+from types import SimpleNamespace
+
+from addmeta import addmeta
+
+class AddMeta:
+    """Add metadata to model output directories using the addmeta tool"""
+
+    default_options = {
+        'enable': True,
+        'verbose': False,
+        'update-history': False,
+        'data': {},
+        'metafiles': [],
+        'datafiles': [],
+        'fnregex': '',
+    }
+    submodel_options = set('metafiles', 'fnregex')
+
+    def __init__(self, options, submodel_options=None):
+        self.options = SimpleNamespace(**options)
+        self.submodels = {}
+        if submodel_options:
+            for submodel, options in submodel_options.items():
+                self.submodels[submodel] = SimpleNamespace(**options)
+
+    @classmethod
+    def from_config(cls, config):
+        """Create an AddMeta instance from a configuration dictionary"""
+        options = cls.default_options.copy()
+        options.update(config)
+
+        submodel_options = {}
+        for submodel, submodel_config in config.get('submodel', {}).items():
+            if submodel_config.get('enable', True):
+                for key in cls.submodel_options:
+                    if key in submodel_config:
+                        submodel_options[key] = submodel_config[key]
+
+        return cls(options, submodel_options)
+
+    def run(self, submodel):
+        """Run the addmeta tool with the specified configuration"""
+
+        if submodel in self.submodels:
+            for submodel in self.submodels:
+                submodel_options = self.submodels[submodel]
+                options = vars(self.options) | vars(submodel_options)
+
+                addmeta.find_and_add_meta(
+                    files=self.options.files,
+                    metafiles=self.options.metafiles,
+                    data=self.options.data,
+                    fnregex=self.options.fnregex,
+                    sort_attrs=self.options.sort,
+                    history=self.options.history,
+                    verbose=self.options.verbose,
+                )
+            
+        else if self.options.files:
+            addmeta.find_and_add_meta(
+                files=self.options.files,
+                metafiles=self.options.metafiles,
+                data=self.options.data,
+                fnregex=self.options.fnregex,
+                sort_attrs=self.options.sort,
+                history=self.options.history,
+                verbose=self.options.verbose,
+            )
+        
+
+def add_meta_data(expt, config):
+    """Add metadata to model output directories using the addmeta tool"""
+
+    for submodel, submodel_config in config.get('submodel', {}).items():
+        if submodel_config.get('enable', True):
+            addmeta_instance = AddMeta.from_config(submodel_config)
+            addmeta_instance.run()  
+    addmeta_instance = AddMeta.from_config(config)
+    addmeta_instance.run()
--- a/payu/addmeta.py
+++ b/payu/addmeta.py
@@ -6,7 +6,7 @@ addmeta tool
 """
 from types import SimpleNamespace
 
-import addmeta as addmeta_lib
+from addmeta import cli as addmeta_cli
 from addmeta import combine_meta, dict_merge, load_data_files, find_and_add_meta
 
 class AddMeta:
@@ -14,14 +14,17 @@ class AddMeta:
 
     default_options = {
         'enable': True,
+    default_options = {
+        'enable': True,
         'verbose': False,
-        'sort': False,
-        'update-history': False,
-        'datavar': {},
+        'update_history': False,
+        'data': {},
         'metafiles': [],
-        'metalist': [],
         'datafiles': [],
         'fnregex': '',
+        'datavar': [],
+        'metalist': '',
+        'sort': True,
     }
 
     def __init__(self, options):
--- a/payu/addmeta.py
+++ b/payu/addmeta.py
@@ -4,60 +4,6 @@ addmeta tool
 :copyright: Copyright 2011 Marshall Ward, see AUTHORS for details.
 :license: Apache License, Version 2.0, see LICENSE for details.
 """
-"""
-do
-    addmeta \
-        -v -s \
-        -d metadata.yaml \
-        -d $PAYU_CURRENT_OUTPUT_DIR/env.yaml \
-        -m scripts/post-processing/addmeta/dataspec.yaml \
-        -m scripts/post-processing/addmeta/${submodel}.yaml \
-        --fnregex='access-esm1p6\.\w+(?:\.\dd)?\.(?P<var>\w+)\.(?P<freq>\w{2,4})(?:\.\w+)?(?:\.\d{4})?\.nc' \
-        $PAYU_CURRENT_OUTPUT_DIR/${submodel}/*.nc
-"""
-
-"""
-addmeta:
-    enable: true                               # default
-    verbose: true                              # default
-    update-history: false                      # default
-    data:
-        id: "{{ metadata.experiment_id }}"     # default
-        run_id: "{{ env.PAYU_RUN_ID }}"        # default
-    metafiles:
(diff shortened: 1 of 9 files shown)


payu/addmeta.py
```diff
@@ -0,0 +1,49 @@
+"""Tooling to add metata to model output directories using the
+addmeta tool
+
+:copyright: Copyright 2011 Marshall Ward, see AUTHORS for details.
+:license: Apache License, Version 2.0, see LICENSE for details.
+"""
+from types import SimpleNamespace
+
+import addmeta as addmeta_lib
```
Qian-HuiChen:
```suggestion
from addmeta import cli as addmeta_cli
```

payu/addmeta.py
```diff
@@ -0,0 +1,49 @@
...
+    """Add metadata to model output directories using the addmeta tool"""
+
+    default_options = {
+        'enable': True,
+        'verbose': False,
+        'update-history': False,
+        'data': {},
+        'metafiles': [],
+        'datafiles': [],
+        'fnregex': '',
```
Qian-HuiChen, line 23:
```suggestion
    default_options = {
        'enable': True,
        'verbose': False,
        'update_history': False,
        'data': {},
        'metafiles': [],
        'datafiles': [],
        'fnregex': '',
        'datavar': [],
        'metalist': '',
        'sort': True,
```

payu/addmeta.py
```diff
@@ -0,0 +1,49 @@
...
+            if key == 'metafiles':
+                value = self.options.metafiles + value
+            if key == 'data':
+                value = self.options.data | value
+            self.options.__setattr__(key, value)
+
+    def run(self):
+        """Run the addmeta tool with the specified configuration"""
+
+        addmeta_lib.cli.main(self.options)
```
Qian-HuiChen:
```suggestion
        addmeta_cli.main(self.options)
```

payu/models/model.py
```diff
@@ -5,14 +5,18 @@
 """
 
 import os
+from pathlib import Path
 import shutil
 import shlex
 import sys
 import subprocess as sp
 
+import addmeta as addmeta_lib
```
Qian-HuiChen:
```suggestion
from addmeta import cli as addmeta_cli
```

payu/models/model.py
```diff
@@ -486,4 +490,31 @@ def get_cur_expt_time_using_submodel(self, model_types):
...
+        add_meta_config = self.expt.config.get('addmeta', {})
+
+        # Add default files glob. Not supported to define in config.yaml, but can be 
+        # overridden by model-specific addmeta configuration in the model control directory
+        add_meta_config['files'] = f'{self.output_path}/*.nc'
+
+        # Add the top level metadata.yaml and local env.yaml files to the datafiles list
+        # to give the addmeta tool access to the metadata and environment information for the experiment
+        add_meta_config['datafiles'].append(f'{self.expt.archive_path}/env.yaml')
+        add_meta_config['datafiles'].append(f'{self.expt.control_path}/metadata.yaml')
```
Qian-HuiChen:
```suggestion
        add_meta_config['datafiles'] = add_meta_config.get('datafiles', [])
        for file in [f'{self.expt.archive_path}/env.yaml', f'{self.expt.control_path}/metadata.yaml']:
            if os.path.exists(file) and file not in add_meta_config['datafiles']:
                add_meta_config['datafiles'].append(file)
```

payu/models/model.py
```diff
@@ -486,4 +490,31 @@ def get_cur_expt_time_using_submodel(self, model_types):
...
+
+        # Create an AddMeta instance from the configuration dictionary
+        addmeta_instance = AddMeta.from_config(add_meta_config)
+
+        if addmeta_instance.options.enable:
+
+            # Support model specific addmeta configuration in the model control directory
+            cmdfilepath = Path(self.control_path / 'addmeta.cmd')
```
Qian-HuiChen:
```suggestion
            cmdfilepath = Path(self.control_path) / 'addmeta.cmd'
```

payu/models/model.py
```diff
@@ -486,4 +490,31 @@ def get_cur_expt_time_using_submodel(self, model_types):
...
+            if cmdfilepath.exists():
+                model_options = addmeta_lib.cli.main_parse_args(['-c', str(cmdfilepath)])
```
Qian-HuiChen:
```suggestion
                model_options = addmeta_cli.main_parse_args(['-c', str(cmdfilepath)])
```

payu/subcommands/addmeta_cmd.py
```diff
@@ -0,0 +1,108 @@
...
+from payu.laboratory import Laboratory
+import payu.subcommands.args as args
+from payu.fsops import read_config
+from payu.telemetry import record_run
+import payu.errors as errors
+
+title = 'addmeta'
+parameters = {'description': 'Add metadata to model output files'}
+
+arguments = [args.model, args.config, args.initial, args.laboratory, args.dir_path]
```
Qian-HuiChen, line 20:
```suggestion
arguments = [args.model, args.config, args.initial, 
            args.laboratory, args.dir_path, args.dry_run]
```

payu/subcommands/addmeta_cmd.py
```diff
@@ -0,0 +1,108 @@
...
+            depends_on=depends_on,
+            )
+
+    except Exception as e:
+        raise errors.PayuRuntimeError(f"Failed to submit sync job: {e}")
+    return job_id
+
+
+def runcmd(model_type=None, config_path=None, init_run=None, lab_path=None, dir_path=None, 
+           depends_on=None):
```
Qian-HuiChen, line 37:
```suggestion
           depends_on=None, dry_run=False):
```

payu/subcommands/addmeta_cmd.py
```diff
@@ -0,0 +1,108 @@
...
+    addmeta_config = pbs_config.get('addmeta', {})
+
+    pbs_config = {
+        'ncpus': 1,
+        'queue': 'copyq',
+        'mem': '2GB',
+        'walltime': '0:30:00',
+        'qsub_flags': '',
+        'jobname': f'{base_dir[:13]}_a' if dir_path else "addmeta_job",
+    }
```
Qian-HuiChen, line 57:
```suggestion
    pbs_config['ncpus'] = 1
    pbs_config['queue'] = 'copyq'
    pbs_config['mem'] = '2GB'
    pbs_config['walltime'] = '0:30:00'
    pbs_config['qsub_flags'] = ''
    pbs_config['jobname'] = f'{base_dir[:13]}_a' if dir_path else "addmeta_job"
```

payu/subcommands/addmeta_cmd.py
```diff
@@ -0,0 +1,108 @@
...
+    pbs_config = pbs_config | addmeta_config.get('pbs', {})
+
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+
+    # Submit PBS job
+    job_id = cli.submit_job('payu-addmeta', pbs_config, pbs_vars, expt=None,
```
Qian-HuiChen:
```suggestion
    job_id = cli.submit_job('payu-addmeta', pbs_config, pbs_vars, expt=expt, 
```

payu/subcommands/addmeta_cmd.py
```diff
@@ -0,0 +1,108 @@
...
+    pbs_config = pbs_config | addmeta_config.get('pbs', {})
+
+    # Initialise experiment to determine archive path and run number (which is needed to write job file)
+    lab = Laboratory(model_type, config_path, lab_path)
+    expt = Experiment(lab)
+
+    # Submit PBS job
+    job_id = cli.submit_job('payu-addmeta', pbs_config, pbs_vars, expt=None, 
+                   current_run=int(init_run) if init_run is not None else None, type='addmeta',
+                   depends_on=depends_on)
```
Qian-HuiChen, line 71:
```suggestion
                   depends_on=depends_on, dry_run=dry_run)
```

Qian-HuiChen (commented):
I run some testing and think this might help

payu/models/model.py
```diff
@@ -486,4 +490,31 @@ def get_cur_expt_time_using_submodel(self, model_types):
...
+        )
+
+    def add_file_metadata(self):
+        """Add metadata to model output files using the addmeta tool"""
+
+        add_meta_config = self.expt.config.get('addmeta', {})
+
+        # Add default files glob. Not supported to define in config.yaml, but can be 
+        # overridden by model-specific addmeta configuration in the model control directory
+        add_meta_config['files'] = f'{self.output_path}/*.nc'
```
Qian-HuiChen:
```suggestion
        add_meta_config['files'] = list(Path(self.output_path).glob('*.nc'))
```

aidanheerdegen: This is passed as a pattern to `addmeta` which performs the `glob`, but thanks for the suggestion.

payu/models/model.py
aidanheerdegen: Thanks!

payu/subcommands/addmeta_cmd.py
aidanheerdegen: I got an error with `args.dry_run` included as it isn't a supported argument for this function. Maybe it should be?

payu/subcommands/addmeta_cmd.py
aidanheerdegen: Ah, I see you're recommending I add the argument, but it isn't being used ... but maybe it should be.

payu/subcommands/addmeta_cmd.py
aidanheerdegen: I'm happy to make this change but just wondering why the preference for this form?

payu/subcommands/addmeta_cmd.py
aidanheerdegen: See previous comments.

payu/subcommands/addmeta_cmd.py
Qian-HuiChen: Because this [line](https://github.com/payu-org/payu/pull/859/changes/BASE..3f90aca7c4175e97819791390c6df3193a23d033#diff-795737a5f28eb049c5b1d99ca55a508fe3ec6f47a2e7da18cfc3fc7c10bdbc38R39)  reads the pbs config, and also configure the control path when not defined.
The previous code overrides the existing pbs config

aidanheerdegen: Ah good point! Thanks. I think I'll change it to use `update` if you're happy with that? I like the readability of the dict format and I think `update` helps to reinforce that this is overriding some of an existing dict.

Qian-HuiChen: happy happy happy

payu/subcommands/addmeta_cmd.py
Qian-HuiChen: I am thinking that all other payu subcommands that submit a PBS job (run/collate/postscript/sync) accept a dry-run option. To keep it consistent, maybe we offer this dry-run option to addmeta as well (?)

payu/models/model.py
Qian-HuiChen: Ohhh I see. When I was testing it, addmeta errors out because it tried to read `/path/to/*.nc` as `/`, `p`, `a`, `t`, `h` etc.

payu/subcommands/addmeta_cmd.py
aidanheerdegen: Yes consistency is good, but an option that does nothing isn't good practice. `addmeta` doesn't have a `dry-run` option, but it is a good idea to have one, so I'll make an issue to add it. If it gets in before this is merged I'll add the option.

Sound ok?

payu/models/model.py
aidanheerdegen: Yes, so your solution was half right, it needs to be a list. I will fix.

payu/addmeta.py
```diff
@@ -0,0 +1,109 @@
...
+                value = self.options.datavar | value
+            self.options.__setattr__(key, value)
+
+    def run(self):
+        """Run the addmeta tool with the specified configuration"""
+
+        # addmeta_cli.main(self.options)
+
+        # Below is a copy of the main routine above to inject default metadata
+        # until the addmeta tool is updated to support default metadata.
```
aidanheerdegen, line 57: Made an issue for this

https://github.com/ACCESS-NRI/addmeta/issues/105

payu/subcommands/addmeta_cmd.py
aidanheerdegen: Made an issue for this https://github.com/ACCESS-NRI/addmeta/issues/106

aidanheerdegen: Made an issue for this https://github.com/ACCESS-NRI/addmeta/issues/106

Qian-HuiChen: Thanks for making the issue in addmeta repo. I am not sure if we are on the same page here. 

When `dry-run=True`, payu prints the `qsub` command without actually submitting the job. So for `payu addmeta --dry-run`, I am thinking of something like:
```
------ Dry run (submission skipped) ------
Generated command: qsub -P ... python3 /scratch/xx/bin/payu-addmeta
```

My understanding is that this dry-run behaviour can be implemented in payu, so it should work even if there is no --dry-run option in the addmeta repo itself.

aidanheerdegen: Ahh! You're right. I am in fact on the wrong page, so thanks for pointing that out. So the `--dry-run` option is independent of the command it is running, even if this is a good idea for `addmeta`.

I'm now wondering now about implementing this *after* (or as part of) #863. addmeta is fairly fast and not resource intensive, and it makes much more sense to run it as part of a post-processing step.

I say this as it would mean there wouldn't need to be a PBS submission for `addmeta`.

Qian-HuiChen: I think it is a good idea to run addmeta without PBS queuing. It can definitely be part of #863 :)

aidanheerdegen:
Tests are failing because this PR is using a development version of addmeta from this PR: https://github.com/ACCESS-NRI/addmeta/pull/111

PR #866, opened 2026-09-15
Improve and fix bugs in payu status

### LINKED ISSUE #864: Payu status to show `Run 5 / N` when doing multiple cycles
At the moment I'm doing a multi-cycle run with Payu, e.g. `Payu run -n X` the trouble was 24 hours later I couldn't remember what `N` was! (Some digging around in log files I was eventually able to find it.)

Would it be possible to show this in the `payu status` update, here's an example (where I've added the bold bit):
```bash
[cyb561.gadi-login-04: MC_8km_jra_ryf-test1]$ payu status
Run: 11 **of 12**
  Job ID:            178923139.gadi-pbs
  Run ID:            e2c266777dd5363bc24fdead260e014fb9835ef0
  Stage:             model-run
  Current Expt Time: 1900-12-21T00:00:00
  Job File:          /scratch/tm70/cyb561/access-om3/archive/MC_8km_jra_ryf-test1-780-dev-MC_8km_jra_ryf-2-28125d05/payu_jobs/11/run/178923139.gadi-pbs.json
```

As discussed with @aidanheerdegen @Qian-HuiChen.

Related: https://github.com/payu-org/payu/issues/640

### LINKED ISSUE #865: Payu should not retry query scheduler for server-killed jobs
If a job is killed by server (e.g., dependency failed), `qstat -xf -F THIS_JOB` will return an error as:
```
Return code: 153
stdout: {
    "timestamp":xxx,
    "pbs_version":"2024.1.2.20241017100211",
    "pbs_server":"gadi-pbs-01"
}
stderr: qstat: Unknown Job Id xxx.gadi-pbs
```
`get_job_info_json` (see [code](https://github.com/payu-org/payu/blob/7aa60620befd1bf6941c959d3fcae7b38cecffd3/payu/schedulers/pbs.py#L554-L564)) should only retry when `qstat` times out.
For other errors, such as PBS returning exit status 153 (Unknown Job Id), it should not retry.

### LINKED ISSUE #867: Payu status --update prints a module load message
`payu status --update` displays a message that is not necessary:
```
payu: Found modules in /opt/Modules/v4.3.0
```

Ref https://github.com/ACCESS-NRI/model-config-tests/pull/252#discussion_r4010976466

Qian-HuiChen:
This PR improves the `payu status` in the following three perspective.

1. Display the total run number in payu status, such as
```
Run: 65 of 65
  Job ID:            12222.gadi-pbs
  Run ID:            xxxx
  Stage:             archive
  Total Queue Time:  0h 2m 14s
  Model Finish Time: 1901-07-30T00:00:00
  Model Exit Code:   0 (Success)
  Output Log:        xxxx
  Error Log:         xxxx
  Job File:          xxxx.gadi-pbs.json
```
Closes #864 

2. Better handles the subprocess error and retry behaviour when querying the scheduler. For server-killed job, scheduler will return an error in qstat and this should not be retry. Only subprocess.timeExpired should be retried, up to 3 times.

Closes #865

3. Fix the unnecessary message printed by `payu status --update`. 
Closes #867.


payu/schedulers/pbs.py
```diff
@@ -582,12 +582,21 @@ def get_job_info_json(
             f"\n Error: {e}"
         )
         raise
```
Qian-HuiChen: I am not sure in what case this error will be raised. Should it be retried or return here?

payu/schedulers/pbs.py
```diff
@@ -573,7 +573,7 @@ def get_job_info_json(
     # Parse the JSON output
     try:
         qstat_output = subprocess.run(
-            cmd, capture_output=True, text=True, check=True,
+            cmd, capture_output=True, text=True, check=True, timeout=10
```
aidanheerdegen: This seems like a good idea to make the process a little more robust. Was this the motivation? 

Just checking as this isn't explicitly mentioned in the PR description or related issues.

I'm not sure how this will interact with the `@retry` decorator, which is setting a total delay of 10 seconds before giving up. 

Maybe we should be using a `stop_after_attempt` and a total timeout?

https://tenacity.readthedocs.io/en/latest/#stopping

Should the `subprocess.run` timeout be reduced to 5s? Is there a good reason to wait 10s?

payu/schedulers/pbs.py
aidanheerdegen: Good question. In general I'd `raise` things, but we don't want to stop a run if the PBS server is having temporary issues. So it might depend if this would block an actual run of the model.

You can customise the exceptions that `@retry` will retry from too FYI

https://tenacity.readthedocs.io/en/latest/#whether-to-retry

payu/schedulers/pbs.py
Qian-HuiChen: Yes, I am thinking only retry it for timeout error. In [2688df7](https://github.com/payu-org/payu/pull/866/commits/2688df7fad244f194971bde5de961656ad004108), I update it to retry for 3 times for timeout errors, and no retry for other errors.

payu/schedulers/pbs.py
Qian-HuiChen: Right. I updated it to raise an error in the latest two commits.

In `status.py`, when the postscript job info needs to be updated, `get_job_info` is wrapped in a `try/except` block so that it can continue running and parse the status from stdout ([code](https://github.com/payu-org/payu/pull/866/changes#diff-fca37d36316f2b9077aa237ee539c593de2ed038ada93dad5c27153114b727c8R509-R513)).

In `telemetry.record_run`, `get_job_info` is also wrapped in a `try/except` block ([code](https://github.com/payu-org/payu/pull/866/changes#diff-f9438cbda6d0d535a69ab0133f0f275028b50349a55985d279eb9e46e939cbf1R591)), so it won't prevent payu from resubmitting the job.

aidanheerdegen (approved):
Looks good, thanks!

Qian-HuiChen:
Manual tests passed with a newly-cloned OM2:
- `payu run -n 2` with collate, postscript and sync enabled. `payu status` and `payu status --update` show correct info without noise.
- Similar to first test but set postscript to exit with `1`. `payu status` and `payu status --update` show correct info without noise. `payu status --update` cleans up sync job file after postscript failed.

